PR #6, opened 2022-01-31
Allow providing fetch options for fetch transport

domasx2:
Adds ability to provide request options when initializing fetch transport. Useful to configure any auth credentials required by receiver or any gateway in front of a receiver.

### DIFF
--- a/packages/core/src/transports/fetch.ts
+++ b/packages/core/src/transports/fetch.ts
@@ -8,14 +8,19 @@ function sendAsBeacon(url: string, body: string): void {
   navigator.sendBeacon(url, blobBody);
 }
 
-function sendAsFetch(url: string, body: string, isDebugEnabled = false): void {
+type RequestOptions = Omit<RequestInit, 'url' | 'body'>;
+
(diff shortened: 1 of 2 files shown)


packages/core/src/transports/fetch.ts
```diff
@@ -24,7 +29,7 @@ function sendAsFetch(url: string, body: string, isDebugEnabled = false): void {
   }); // the empty callback is required as otherwise the catch will be ignored
 }
 
-export function getFetchTransport(url: string, isDebugEnabled = false): Transport {
+export function getFetchTransport(url: string, isDebugEnabled = false, options: RequestOptions = {}): Transport {
```
bfmatei, line 32: This transport already has 3 params. Should we create a single `options` parameter and add the current ones as part of that?

I see an issue with the current approach: you may want to pass headers but not the debugging flag

domasx2: done

kpelelis (approved):

bfmatei (approved):

PR #13, opened 2022-04-07
Meta improvements

domasx2:
Bunch of stuff:

1. Strongly typed metadata
2. Sugar config properties for setting commonly used metadata: `app`, `session`, `user`
3. API sugar for setting user: `agent.api.setUser()` 
4. Instrumentations decorated with `name` and `version`. These are collected into `sdk.integrations` meta property
5. `grafanaJavaScriptAgent` -> `grafanaAgent`

I "simplified" the metas API a bit, hopefully not controversial


packages/instrumentation-console/src/index.ts
```diff
@@ -10,19 +10,23 @@ export default function getConsoleInstrumentation({
...
+            try {
+              agent.api.pushLog(args, level);
+            } catch (err) {
+            } finally {
+              agent.api.callOriginalConsoleMethod(level, ...args);
+            }
+          };
+        });
+    },
+    version: '1.0',
```
bfmatei, line 29: Version set similar to how we do it in meta initialization.

domasx2: I added `prebuild` scripts that dump version from `package.json` to `src/version.ts` and then use that

domasx2:
I think we discussed some more changes, but failed to write them down :disappointed: @bfmatei maybe you remember?

packages/core/README.md
```diff
@@ -21,11 +21,14 @@ The agent requires a configuration parameter.
 
 | Property                | Description                                                             | Type                | Optional | Default Value            |
 | ----------------------- | ----------------------------------------------------------------------- | ------------------- | -------- | ------------------------ |
-| `globalObjectKey`       | String that should be used when defining the agent on the global object | `string`            | Y        | `grafanaJavaScriptAgent` |
-| `metas`                 | Array of metas that should be logged                                    | `Meta[]`            | Y        | `[]`                     |
+| `app`                   | Application metadata                                                    | `App`               | Y        | `undefined`              |
+| `globalObjectKey`       | String that should be used when defining the agent on the global object | `string`            | Y        | `grafanaAgent`           |
+| `metas`                 | Array of metas that should be logged                                    | `MetaItem[]`        | Y        | `[]`                     |
```
bfmatei, line 26: `metas` > `instrumentations` :D

packages/core/README.md
```diff
@@ -21,11 +21,14 @@ The agent requires a configuration parameter.
...
 | `instrumentations`      | Array of instrumentations that should be ran                            | `Instrumentation[]` | N        | `[]`                     |
 | `preventGlobalExposure` | Flag for toggling the definition on the global object                   | `boolean`           | Y        | `false`                  |
 | `transports`            | Array of transports that should be used                                 | `Transport[]`       | Y        | `[]`                     |
+| `session`               | Session metadata                                                        | `Session`           | Y        | `undefined`              |
```
bfmatei, line 30: `preventGlobalExposure` > `session` > `transports`

packages/core/package.json
```diff
@@ -18,6 +18,7 @@
   ],
   "scripts": {
     "start": "yarn build -w",
+    "prebuild": "node -p \"'export const VERSION = \\'' + String(require('./package.json').version) + '\\';'\" > src/version.ts",
```
bfmatei, line 21: Maybe move this into a separate script file?

packages/core/src/api/meta/initialize.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { Transports } from '../../transports';
+import type { MetaAPI } from './types';
+
+export function initializeMeta(_: Transports, metas: Metas): MetaAPI {
+  let meta: Meta = {};
+
+  metas.add(() => meta);
+
+  return {
+    setUser: (user: User | null) => {
```
bfmatei, line 11: Should we move `setUser` into a separate constant and just reference it here? It would be aligned w/ the rest of the agent code. A benefit would be easier maintenance on the long run.

packages/core/src/api/meta/initialize.ts
```diff
@@ -0,0 +1,15 @@
...
+      meta = { user: user ?? undefined };
```
bfmatei, line 12:
```suggestion
      meta = {
        ...meta,
        user: user ?? undefined,
      };
```

packages/instrumentation-console/package.json
```diff
@@ -18,6 +18,7 @@
   ],
   "scripts": {
     "start": "yarn build -w",
+    "prebuild": "node -p \"'export const VERSION = \\'' + String(require('./package.json').version) + '\\';'\" > src/version.ts",
```
bfmatei, line 21: Same about moving this into a script.

Will we have different versions for the subpackages vs core package? If not, we can export it once from the core package and then reuse it in subpackages. We can control the exact versions via `peerDependency`

packages/instrumentation-console/src/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.0.0';
```
bfmatei, line 1:
```suggestion
export const VERSION = '0.0.0-dev';
```

Suffix indicates a development version, according to semantic versioning.

packages/instrumentation-errors/package.json
```diff
@@ -18,6 +18,7 @@
   ],
   "scripts": {
     "start": "yarn build -w",
+    "prebuild": "node -p \"'export const VERSION = \\'' + String(require('./package.json').version) + '\\';'\" > src/version.ts",
```
bfmatei, line 21: Same about moving the script

packages/instrumentation-tracing/src/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.0.0';
```
bfmatei, line 1:
```suggestion
export const VERSION = '0.0.0-dev';
```

packages/instrumentation-web-vitals/package.json
```diff
@@ -18,6 +18,7 @@
   ],
   "scripts": {
     "start": "yarn build -w",
+    "prebuild": "node -p \"'export const VERSION = \\'' + String(require('./package.json').version) + '\\';'\" > src/version.ts",
```
bfmatei, line 21: Same about moving the script

packages/instrumentation-web-vitals/src/index.ts
```diff
@@ -10,17 +12,21 @@ const map = {
...
-const webVitalsInstrumentation: Instrumentation = async () => {
-  Object.entries(map).forEach(([indicator, executor]) => {
-    executor((metric) => {
-      agent.api.pushMeasurement({
-        type: 'web-vitals',
-        values: {
-          [indicator]: metric.value,
-        },
+const webVitalsInstrumentation: Instrumentation = {
+  initialize: async () => {
```
bfmatei, line 16:
```suggestion
  initialize: () => {
```

This does not need to be `async`

packages/instrumentation-web-vitals/src/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.0.0';
```
bfmatei, line 1:
```suggestion
export const VERSION = '0.0.0-dev';
```

packages/core/src/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.0.0';
```
bfmatei, line 1:
```suggestion
export const VERSION = '0.0.0-dev';
```

packages/core/package.json
bfmatei: We can also do something else here: export the version as a constant enum which is actually replaced in the compiled code like this:
```
export const enum VERSION = {
  CURRENT: '0.0.0-dev',
}

console.log(VERSION.CURRENT);
```

which would be compiled to: `console.log('0.0.0-dev');`

We can do it once in the core package, reference the constant enum in the subpackages which would replace everything with static strings.

bfmatei (commented):
some nitpicks, some questions, some suggestions

packages/core/package.json
domasx2: ohh right, let's do it

domasx2: - [ ] create version file in core only

packages/core/src/api/meta/initialize.ts
domasx2: - [ ] refac

packages/core/src/version.ts
domasx2: this is generated file

packages/instrumentation-console/package.json
domasx2: yup, will do

packages/core/README.md
```diff
@@ -160,29 +163,12 @@ initializeAgent({
...
       return {
-        initialViewport,
-
-        // You can also define instant metas here
-        viewPort: () => ({
-          height: window.innerHeight,
-          width: window.innerWidth,
-        }),
+        user: {
+          username: getUser().name
```
kpelelis, line 170: I am not sure about the legal mubmo jumbo, but should we run this through leagal?

packages/core/src/metas/initialize.ts
```diff
@@ -1,44 +1,48 @@
...
+    items = items.filter((i) => i !== item);
   };
 
-  add('sdk', () => ({
-    name: '@grafana/javascript-agent-core',
-    version: '0.0.1', // TODO: set correct version here
-  }));
+  const initial: Meta = {
+    sdk: {
+      name: '@grafana/agent-core',
```
kpelelis, line 19: We can probably pull that from the package json file

packages/core/src/version.ts
kpelelis: We can use package.json for this. It is a json file after all

kpelelis (changes requested):

packages/core/src/metas/initialize.ts
bfmatei: It's a little bit more difficult. `package.json` is outside of `src` mentioned in `tsconfig.json`. Also it may break once compiled as the path to `package.json` may changed.

I tried it in the early versions of the library.

packages/core/README.md
domasx2: I need a linter for this, my brain does not include a sort function :D

bfmatei: It's called OCD.  Some may consider it a bug, but in this case it is a feature.

packages/core/README.md
domasx2: I don't think so, this is just an example in readme

bfmatei (approved):

kpelelis (approved):

PR #15, opened 2022-05-02
Refac transports, instrumentations as classes

domasx2:
To unify signatures, refac transports, instrumentations as classes.

Will do metas separately, there's a PR open that refacs them

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -11,6 +11,7 @@
     "directory": "demo"
   },
   "source": "src/index.html",
+  "private": true,
   "scripts": {
     "start": "parcel serve",
     "build": "parcel build",


packages/web/src/instrumentations/console/index.ts
```diff
@@ -0,0 +1,28 @@
+import { agent, allLogLevels, LogLevel } from '@grafana/agent-core';
+import type { Instrumentation } from '@grafana/agent-core';
+
+export interface ConsoleInstrumentationOptions {
+  disabledLevels?: LogLevel[];
+}
+
+export class ConsoleInstrumentation implements Instrumentation {
```
bfmatei, line 8: As a suggestion: instead of implementing `Instrumentation`, we can extend a `BasicInstrumentation` which defines the agent as a property. Doing so will allow us to skip an import in every instrumentation file while keeping an OOP-ish approach.

packages/web/src/transports/console/index.ts
```diff
@@ -0,0 +1,20 @@
+import { agent, getTransportBody, LogLevel, prefixAgentMessage, TransportItem } from '@grafana/agent-core';
+import type { Transport } from '@grafana/agent-core';
+
+export interface ConsoleTransportOptions {
+  level?: LogLevel;
+}
+
+export class ConsoleTransport implements Transport {
```
bfmatei, line 8: Same as the integrations.

bfmatei (commented):
comments and stuff like that.

packages/web/src/instrumentations/console/index.ts
domasx2: -[ ] good idea

packages/core/src/instrumentations/base.ts
```diff
@@ -0,0 +1,14 @@
...
+import type { Instrumentation } from './types';
+
+export abstract class BaseInstrumentation implements Instrumentation {
+  abstract initialize(): void;
+
+  get agent(): Agent {
+    return init.agent;
+  }
+
+  abstract name: string;
```
bfmatei, line 12: should `name` and `version` be `static`?

packages/core/src/transports/base.ts
```diff
@@ -0,0 +1,11 @@
+import * as init from '../initialize';
+import type { Agent } from '../types';
+import type { Transport, TransportItem } from './types';
+
+export abstract class BaseTransport implements Transport {
```
bfmatei, line 5: Should `BaseTransport` also contain `name` and `version`?

packages/tracing-web/src/index.ts
```diff
@@ -1,10 +1,10 @@
...
-  initialize: () => {
+export class TracingInstrumentation implements Instrumentation {
+  initialize(): void {
     agent.api.pushLog(['A simple log from the tracing instrumentation package']);
-  },
-  name: '@grafana/agent-tracing-web-instrumentation',
-  version: VERSION,
-};
+  }
+  name = '@grafana/agent-tracing-web:instrumentation';
```
bfmatei, line 8: I know this may be annoying but order of members in a class should be: "variables" and then methods.

bfmatei (approved):
Some nitpicks.

kpelelis (approved):

packages/core/src/transports/base.ts
domasx2: we don't have a use for  it (yet) :thinking:

packages/core/src/instrumentations/base.ts
domasx2: I considered it, but it makes the `Instrumentation` type ugly and forces a specific way of implementing :disappointed:  imho not worthwhile

packages/tracing-web/src/index.ts
domasx2: - [ ] fix ordering

bfmatei (approved):

PR #17, opened 2022-05-18
Feature: filter events before sending

domasx2:
This PR adds ability to modify or filter events before they are sent over the wire.
Primary use case is to filter out "noise", expected errors that can't be easily prevented.

Config gets two new properties:
`ignoreErrors` - Array of strings or regexes. Errors with messages matching these will be dropped
`beforeSend` - Hook that accepts `TransportItem` and returns `TransportItem` or `null`. Can be used to write custom filter or transform the payload

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -21,18 +21,18 @@ initializeAgent({
 
 The agent requires a configuration parameter.
 
-| Property                | Description                                                             | Type                | Optional | Default Value  |
-| ----------------------- | ----------------------------------------------------------------------- | ------------------- | -------- | -------------- |
-| `app`                   | Application metadata                                                    | `App`               | Y        | `undefined`    |
-| `beforeSend`            | Hook to modify or discard an event before it is sent over the wire      | `BeforeSendHook`    | Y        | `undefined`    |
-| `globalObjectKey`       | String that should be used when defining the agent on the global object | `string`            | Y        | `grafanaAgent` |
-| `ignoreErrors`          | Error message patterns for errors that should be ignored                | `Patterns`          | Y        | `[]`           |
-| `instrumentations`      | Array of instrumentations that should be ran                            | `Instrumentation[]` | N        | `[]`           |
-| `metas`                 | Array of metas that should be logged                                    | `MetaItem[]`        | Y        | `[]`           |
-| `preventGlobalExposure` | Flag for toggling the definition on the global object                   | `boolean`           | Y        | `false`        |
-| `session`               | Session metadata                                                        | `Session`           | Y        | `undefined`    |
-| `transports`            | Array of transports that should be used                                 | `Transport[]`       | Y        | `[]`           |
-| `user`                  | User metadata                                                           | `User`              | Y        | `undefined`    |
+| Property                | Description                                                                            | Type                | Optional | Default Value  |
+| ----------------------- | -------------------------------------------------------------------------------------- | ------------------- | -------- | -------------- |
+| `app`                   | Application metadata                                                                   | `App`               | Y        | `undefined`    |
+| `beforeSend`            | Hook invoked before pushing event to transport. Can be used to modify or filter events | `BeforeSendHook`    | Y        | `undefined`    |
+| `globalObjectKey`       | String that should be used when defining the agent on the global object                | `string`            | Y        | `grafanaAgent` |
+| `ignoreErrors`          | Error message patterns for errors that should be ignored                               | `Patterns`          | Y        | `[]`           |
+| `instrumentations`      | Array of instrumentations that should be ran                                           | `Instrumentation[]` | N        | `[]`           |
+| `metas`                 | Array of metas that should be logged                                                   | `MetaItem[]`        | Y        | `[]`           |
+| `preventGlobalExposure` | Flag for toggling the definition on the global object                                  | `boolean`           | Y        | `false`        |
+| `session`               | Session metadata                                                                       | `Session`           | Y        | `undefined`    |
+| `transports`            | Array of transports that should be used                                                | `Transport[]`       | Y        | `[]`           |
+| `user`                  | User metadata                                                                          | `User`              | Y        | `undefined`    |
 
 ## Agent
 
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -21,18 +21,18 @@ initializeAgent({
 
 The agent requires a configuration parameter.
 
-| Property                | Description                                                             | Type                                       | Optional | Default Value  |
-| ----------------------- | ----------------------------------------------------------------------- | ------------------------------------------ | -------- | -------------- |
-| `app`                   | Application metadata                                                    | `App`                                      | Y        | `undefined`    |
-| `beforeSend`            | Hook to modify or discard an event before it is sent over the wire      | `(TransportItem) => TransportItem \| null` | Y        | `undefined`    |
-| `globalObjectKey`       | String that should be used when defining the agent on the global object | `string`                                   | Y        | `grafanaAgent` |
-| `ignoreErrors`          | Error message patterns for errors that should be ignored                | `Array<string \| RegExp>`                  | Y        | `[]`           |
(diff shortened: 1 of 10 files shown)


packages/core/jest.config.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
```
bfmatei, line 1: Is the comment needed? 🤔

packages/core/package.json
```diff
@@ -35,7 +35,10 @@
     "uuid": "^8.3.2"
   },
   "devDependencies": {
-    "@types/uuid": "^8.3.4"
+    "@types/jest": "^27.5.1",
```
bfmatei, line 38: These should go in the root `package.json` and not in the `core` package as it is a boilerplate dependency.

packages/core/README.md
```diff
@@ -21,16 +21,18 @@ initializeAgent({
...
+| Property                | Description                                                             | Type                                       | Optional | Default Value  |
+| ----------------------- | ----------------------------------------------------------------------- | ------------------------------------------ | -------- | -------------- |
+| `app`                   | Application metadata                                                    | `App`                                      | Y        | `undefined`    |
+| `beforeSend`            | Hook to modify or discard an event before it is sent over the wire      | `(TransportItem) => TransportItem \| null` | Y        | `undefined`    |
+| `globalObjectKey`       | String that should be used when defining the agent on the global object | `string`                                   | Y        | `grafanaAgent` |
+| `ignoreErrors`          | Error message patterns for errors that should be ignored                | `Array<string \| RegExp>`                  | Y        | `[]`           |
```
bfmatei, line 29:
```suggestion
| `ignoreErrors`          | Error message patterns for errors that should be ignored                | `Patterns`                  | Y        | `[]`           |
```

packages/core/README.md
```diff
@@ -21,16 +21,18 @@ initializeAgent({
...
-| `session`               | Session metadata                                                        | `Session`           | Y        | `undefined`    |
-| `transports`            | Array of transports that should be used                                 | `Transport[]`       | Y        | `[]`           |
-| `user`                  | User metadata                                                           | `User`              | Y        | `undefined`    |
+| Property                | Description                                                             | Type                                       | Optional | Default Value  |
+| ----------------------- | ----------------------------------------------------------------------- | ------------------------------------------ | -------- | -------------- |
+| `app`                   | Application metadata                                                    | `App`                                      | Y        | `undefined`    |
+| `beforeSend`            | Hook to modify or discard an event before it is sent over the wire      | `(TransportItem) => TransportItem \| null` | Y        | `undefined`    |
```
bfmatei, line 27:
```suggestion
| `beforeSend`            | Hook to modify or discard an event before it is sent over the wire      | `BeforeSendHook` | Y        | `undefined`    |
```

packages/core/src/transports/initialize.ts
```diff
@@ -1,20 +1,52 @@
...
 
   const add: Transports['add'] = (...transports) => {
-    value.push(...transports);
+    transports.push(...transports);
   };
 
-  const execute: Transports['execute'] = (payload) => {
-    value.forEach((transport) => transport.send(payload));
+  const execute: Transports['execute'] = (item) => {
+    let modified = beforeSendHooks.reduce<TransportItem | null>((prev, hook) => (prev ? hook(prev) : null), item);
```
bfmatei, line 22: Should we switch from `reduce` to `for` iteration just to be able to exit faster from the loop?

packages/core/src/transports/initialize.ts
```diff
@@ -1,20 +1,52 @@
...
+}
+
+function createBeforeSendHookFromIgnorePatterns(patterns: Patterns): BeforeSendHook {
+  return (item) => {
+    if (item.type === TransportItemType.EXCEPTION && item.payload) {
+      const event = item.payload as ExceptionEvent;
+      const msg = `${event.type}: ${event.value}`;
+      if (
+        patterns.find((pattern) => {
+          return typeof pattern === 'string' ? msg.includes(pattern) : !!msg.match(pattern);
```
bfmatei, line 44: We have a `isString` helper.

bfmatei (commented):
A couple of comments here and there.

packages/core/src/transports/initialize.ts
domasx2: Definitely doens't matter for performance, unlikely to have more than 1 or 2 items. Is current version hard to read though?

packages/core/README.md
domasx2: Hm, not sure about aliasing such simple types.. Where do we define them then? below the table?

packages/core/package.json
domasx2: - [ ] move to core

packages/core/jest.config.js
domasx2: autogenerated by jest, not sure what it does :shrug:

packages/core/src/transports/initialize.ts
domasx2: - [ ] replace with `isString`

packages/core/README.md
bfmatei: We can point to the definition in the code using a link 🙂

domasx2: ok, makes sense
- [ ] point to type in code

domasx2: I changed the type, but we'll have to link in a separate PR. This code is not in main yet

packages/core/src/transports/initialize.ts
domasx2: refactored. Feels like golang, verbose yet simple :thinking:  not bad

packages/core/README.md
bfmatei: 👍

packages/core/src/transports/initialize.ts
```diff
@@ -1,20 +1,58 @@
...
+}
+
+function createBeforeSendHookFromIgnorePatterns(patterns: Patterns): BeforeSendHook {
+  return (item) => {
+    if (item.type === TransportItemType.EXCEPTION && item.payload) {
+      const event = item.payload as ExceptionEvent;
+      const msg = `${event.type}: ${event.value}`;
+      if (
+        patterns.find((pattern) => {
+          return isString(pattern) ? msg.includes(pattern) : !!msg.match(pattern);
```
kpelelis, line 50: nit: You can save this ternary operator by creating a new `RegEx` object. It accepts both the RegExp literal and a string

packages/core/src/transports/transports.test.ts
```diff
@@ -0,0 +1,106 @@
+import type { APIEvent, ExceptionEvent } from '../api';
```
kpelelis, line 1: As always thanks for tests

kpelelis (approved):
LGTM! nits

packages/core/src/transports/initialize.ts
domasx2: ack, though I think this is fine as well

packages/core/README.md
```diff
@@ -24,7 +24,9 @@ The agent requires a configuration parameter.
 | Property                | Description                                                             | Type                | Optional | Default Value  |
 | ----------------------- | ----------------------------------------------------------------------- | ------------------- | -------- | -------------- |
 | `app`                   | Application metadata                                                    | `App`               | Y        | `undefined`    |
+| `beforeSend`            | Hook to modify or discard an event before it is sent over the wire      | `BeforeSendHook`    | Y        | `undefined`    |
```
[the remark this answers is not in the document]
domasx2: I'll update description
- [ ] update description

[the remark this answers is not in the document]
domasx2: reworded description. I'd like to keep event name. It's modeled after other comparable libs, and imho makes sense to name it after event trigger rather than use cases

kpelelis (approved):

bfmatei (approved):

PR #18, opened 2022-05-26
Add web-specific initializeAgent function to @grafana/agent-web

domasx2:
Adds a web-specific initializer that has simple shortcuts for "basic" setup.

Under the hood it generates a core config with sensible defaults and calls core initializer.

Here's what a simple init can look like, short and sweet:

```javascript
import {  initializeAgent } from '@grafana/agent-web';

const agent = initializeAgent({
  url: 'http://localhost:12345/collect',
  apiKey: 'secret',
  app: {
    name: 'frontend',
    version: '1.0.0',
  },
});
```


packages/web/src/config.ts
```diff
@@ -0,0 +1,66 @@
...
+  globalObjectKey?: string;
+  preventGlobalExposure?: boolean;
+  metas?: MetaItem[];
+  instrumentations?: Instrumentation[];
+  transports?: Transport[];
+}
+
+export const defaultMetas: MetaItem[] = [browserMeta, pageMeta];
+
+export const getDefaultInstrumentations = (): Instrumentation[] => [
```
bfmatei, line 31: Should we add a configuration parameter here to include tracing and console instrumentation as well? This will prevent some additional manual package installations.

packages/web/src/config.ts
```diff
@@ -0,0 +1,66 @@
...
+  new ErrorsInstrumentation(),
+  new WebVitalsInstrumentation(),
+];
+
+export function makeCoreConfig(browserConfig: BrowserConfig): Config {
+  const transports: Transport[] = [];
+  if (browserConfig.transports) {
```
bfmatei, line 38: Why not both here? Use case: as a developer i might have the fetch transport set via the `browserConfig.url` property but I might temporarily need to enable another transport. Because we limit the availability of `browserConfig.url` with ``browserConfig.transports`, I will manually need to install the fetch transport package and add it to the transports list.

packages/web/src/instrumentations/console/index.ts
```diff
@@ -5,7 +5,7 @@ export interface ConsoleInstrumentationOptions {
 }
 
 export class ConsoleInstrumentation extends BaseInstrumentation {
-  static defaultDisabledLevels: LogLevel[] = [LogLevel.DEBUG, LogLevel.LOG];
+  static defaultDisabledLevels: LogLevel[] = [LogLevel.DEBUG, LogLevel.TRACE];
```
bfmatei, line 8: I strongly suggest disabling the `LOG` as well :).

bfmatei, line 8: Also, don't forget to update the README file for this.

bfmatei (commented):
Comments here, comments there.

packages/web/src/config.ts
domasx2: it's all in the `agent-web` package, no separate package. My goal here was to avoid confusion and be explicit about what config will be actual

packages/web/src/instrumentations/console/index.ts
domasx2: - [ ] update readme

packages/web/src/config.ts
bfmatei: `getDefaultInstrumentations` -> `getWebInstrumentations`?

packages/web/src/instrumentations/console/index.ts
bfmatei: The problem with `log` is that there is too much logging happening at this level. Someone may enable it ending up spamming the receiver.

packages/web/src/config.ts
domasx2: sure, but still can't add tracing without balooning the bundle

bfmatei: Yup, that's true. Let's keep tracing out of this.

domasx2: done

kpelelis (approved):
LGTM! One small comment regarding required params

packages/web/src/config.ts
```diff
@@ -0,0 +1,66 @@
...
+  Transport,
+} from '@grafana/agent-core';
+
+import { ErrorsInstrumentation, WebVitalsInstrumentation } from './instrumentations';
+import { browserMeta, pageMeta } from './metas';
+import { FetchTransport } from './transports';
+
+export interface BrowserConfig {
+  url?: string;
+  apiKey?: string;
```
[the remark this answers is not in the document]
domasx2: if you wanna use something else than fetch transport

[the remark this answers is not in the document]
domasx2: eg, only console transport in dev env ? :thinking:  
anyhow, there's a rumtime check later that requires either `url` or `transports` to be present

bfmatei (approved):

PR #24, opened 2022-06-03
Feature: agent.api.pushError method

### LINKED ISSUE #20: Support capturing Error objects
Add support for capturing Error objects like so:

```javascript
const err = new Error('oh nooo');
agent.api.pushException(err);
```

domasx2:
This pr adds a method to capture exceptions from `Error` objects:

```javascript
const err = new Error('bad thing happend');
grafanaAgent.api.pushError(err);
```

Notes:
Added `parseStacktrace` parameter to core config, so that web / node agents can supply own stacktrace parser

fixes #20

- [x] add tests

### DIFF
--- a/packages/core/src/api/exceptions/initialize.ts
+++ b/packages/core/src/api/exceptions/initialize.ts
@@ -1,12 +1,18 @@
+import type { Config } from '../../config';
 import type { Metas } from '../../metas';
 import { TransportItemType } from '../../transports';
 import type { TransportItem, Transports } from '../../transports';
 import { getCurrentTimestamp } from '../../utils';
 import type { TracesAPI } from '../traces';
 import { defaultExceptionType } from './const';
-import type { ExceptionEvent, ExceptionsAPI } from './types';
+import type { ExceptionEvent, ExceptionsAPI, PushExceptionOptions } from './types';
 
-export function initializeExceptions(transports: Transports, metas: Metas, tracesApi: TracesAPI): ExceptionsAPI {
+export function initializeExceptions(
+  config: Config,
+  transports: Transports,
+  metas: Metas,
+  tracesApi: TracesAPI
+): ExceptionsAPI {
   const pushException: ExceptionsAPI['pushException'] = (value, { stackFrames, type } = {}) => {
     try {
       const item: TransportItem<ExceptionEvent> = {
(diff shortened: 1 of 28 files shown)


packages/core/src/api/exceptions/initialize.ts
```diff
@@ -32,7 +38,21 @@ export function initializeExceptions(transports: Transports, metas: Metas, trace
...
+    if (error.stack && config.parseStacktrace) {
+      opts.stackFrames = config.parseStacktrace(error).frames;
+    }
+
+    return pushException(message, opts);
+  };
+
   return {
     pushException,
+    pushError,
```
bfmatei, line 56: imho, the naming is a little bit confusing. An exception is an error technically and we also have a `pushLog` which can be an error.

bfmatei: Error objects are pretty hard to work with as they are non-standard across apps, browsers and platforms.

We can provide a helper method to the users that allow them to manually parse an error and then leverage on `pushExceptions` as more or less this is what we are doing under the hood.

bfmatei (changes requested):
Like I mentioned in the comment, users will/might get confused by this. I think we need to discuss this more before implementing it.

Also docs are missing.

packages/core/src/api/exceptions/initialize.ts
domasx2: I feel strongly about this: 
It's idiomatic to throw `Error` objects in js, and the most common use case will be to report such error afte receiving it through `catch(e) {...}`. Thus we need an easy way to do it. All comparable libs have it as well.

Conversly, the current `pushException` feels like an internal thing. If the user has a non-Error object, probaly a string message, most likely he should use the `api.pushLog` with `error` level. There's not really a situation where you have stack trace, type and message but not an `Error`, other than the `onerror` event, which we already cover.

So to make things less confusing, maybe let's make `pushException` internal?

bfmatei: So basically:
- allow other packages and/or the users to pass a stacktrace parser and make `pushException` receive an error object instead of string. The stacktrace would be extracted using the mentioned parser and the type would be extracted from the `name` property on the object.
- for string-based errors just use the logging API from the lib

In this case I don’t see the need for another method at this level tho’. Just enhance the existing one? Maybe add a config property that forces a specific type and stacktrace if somebody actually really wants to override them.

domasx2: It seems like we're on the same page more or less then? Leave one method which consumes `Error`, parser ir configurable, provided automatically if you init via `@grafana/agent-web`, or `@grafana/agent-node` in the future

domasx2: updated

bfmatei (approved):

PR #27, opened 2022-06-13
Add esm module build

domasx2:
There was an issue with our build set up: `package.json` would specify `module: true` (meaning es modules), but we build common-js modules. Under some circumstances this would make webpack produce an error. 

Refactored our build setup to publish two distinct builds:
* common-js modules with `es5` target, located at `build/cjs` and refered to by `index` field in `package.json`
* es modules with `es6` target located at `build/esm` and refered to by `module` field in `package.json`

AFAIK this should support well enough both modern and "legacy" builds.

Other minor drive-by changes:
* `@grafana/agent-web` reexports `Agent` type from `@grafana/agent-core`. This way It's not required to add `@grafana/agent-core` as top level dependency just for the type.
* Deleted some random things that I didn't understand the purpose of, like `packages/tsconfig.json`. Doesnt' seem like anything broke :)


demo/src/actions.ts
```diff
@@ -1,4 +1,4 @@
-import '@grafana/agent-web/dist/globals';
+import '@grafana/agent-web/build/cjs/globals';
```
bfmatei, line 1: This looks ugly :-/

.gitignore
```diff
@@ -8,6 +8,7 @@
 .parcel-cache
 coverage
 dist
+build
```
bfmatei, line 11: a `build` folder besides a `dist` folder? Let's stick to one i think

packages/core/bin/gen-version.js
```diff
@@ -1,5 +1,5 @@
 #!node
-import { writeFileSync, readFileSync } from 'fs';
+const { writeFileSync, readFileSync } = require('fs');
```
bfmatei, line 2: I think we can skip using `gen-version` and just rely on `npm_package_version` env variable?

packages/core/package.json
```diff
@@ -11,18 +11,20 @@
...
+    "README.md",
+    "LICENSE"
   ],
   "scripts": {
     "start": "yarn build -w",
     "postversion": "yarn prebuild",
     "prebuild": "node bin/gen-version.js",
-    "build": "tsc -p tsconfig.json",
-    "clean": "rimraf dist/ yarn-error.log",
+    "build": "tsc --build tsconfig.all.json",
```
bfmatei, line 26: ain't the `-p` flag necessary?

packages/core/tsconfig.all.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "files": [],
+  "references": [{ "path": "./tsconfig.json" }, { "path": "./tsconfig.esm.json" }]
```
bfmatei, line 4: Should we rename the `tsconfig.json` to `tsconfig.cjs.json`?

.gitignore
bfmatei: AFAIK `dist` is the convention for buildable libraries.

packages/tsconfig.json
```diff
@@ -1,17 +0,0 @@
-{
```
bfmatei, line 1: This file is (was?! i didn't have the chance to test it) necessary in various IDEs for autocomplete and so on. Monorepos confuse IDEs and text editors :-/

packages/web/tsconfig.json
```diff
@@ -1,15 +1,9 @@
...
-  "extends": "../tsconfig.json",
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "./dist",
+    "outDir": "./build/cjs",
     "rootDir": "./src",
     "tsBuildInfoFile": "../../.build-cache/web.tsbuildinfo"
   },
-  "include": ["./src"],
-  "references": [
```
bfmatei, line 9: Wait, no reference?

packages/web/src/index.ts
```diff
@@ -4,3 +4,4 @@ export * from './transports';
 export * from './config';
 export * from './initialize';
 export { LogLevel } from '@grafana/agent-core';
+export type { Agent } from '@grafana/agent-core';
```
bfmatei, line 7: I think we also need to export the actual `agent` object.

tsconfig.base.esm.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "module": "ES6",
+    "moduleResolution": "node",
```
bfmatei, line 5: `moduleResolution: "node"` is the default value IIRC, hence we can skip it

tsconfig.base.esm.json
```diff
@@ -0,0 +1,9 @@
...
+    "target": "es6",
+    "downlevelIteration": true
```
bfmatei, line 7: `downlevelIteration` in ES6 code does not look right. `Symbol` should be supported for ES6. Also `downlevelIteration` produces more noise and less performant code.

bfmatei (commented):
Comments, nitpicks, suggestions etc.

packages/integration-react/package.json
```diff
@@ -11,21 +11,18 @@
...
+  "main": "build/cjs/index.js",
+  "module": "build/esm/index.js",
+  "types": "build/cjs/index.d.ts",
   "files": [
-    "dist"
+    "build",
+    "README.md",
+    "LICENSE"
   ],
-  "imports": {
```
bfmatei, line 20: IIRC the `imports` property was there for allowing imports from the build folder instead of the source folder. I had issues in the beginning with this.

packages/web/src/index.ts
bfmatei: The rationale behind this is: if you manually install `@grafana/agent-core` and import the agent from there, that would be another instance of the agent instead of the one used by `@grafana/agent-web`.

packages/core/bin/gen-version.js
kpelelis: If a package is being bundle though, do we get the env variable of the agent or the package?

kpelelis (changes requested):
I agree with most of the points Bogdan made so I am requesting changes just to know the state of this pull request

packages/core/bin/gen-version.js
bfmatei: LE: just tested. `npm_package_version` is unreliable as it is not actually replaced (tried tsc/ts-node) but left there in the compiled code. Guess what happens when you load a library that relies on `npm_package_version` internally... it is actually getting the version from the `package.json` from where you run the `npm start cmd` or `yarn cmd` :-/

.gitignore
domasx2: I don't think there's a convention - eg, opentelemetry as well as sentry use `build`. I kinda like it more as it's more self explanatory, but no strong feelings

domasx2: `dist` is only left in `.gitignore` atm

packages/core/package.json
domasx2: Well, this works :thinking:  I copied from otel. Do you prefer `-p`? as far as I can tell, it's equivalent

packages/core/tsconfig.all.json
domasx2: we have to keep a `tsconfig.json` at root level I think, for tooling to find? Can add a `cjs` tho maybe that derives from `base`

packages/tsconfig.json
domasx2: vs code seems fine without, didn't notice any effect. I'm not sure why would an ide read it either, it's not topomost to any ts code :thinking:  can you test? Or would you rather I just add it back?

packages/web/tsconfig.json
domasx2: Oops. Why did nothing break tho?

-[  ] restore

tsconfig.base.esm.json
domasx2: - [ ] skip module resolution

tsconfig.base.esm.json
domasx2: - [ ] disable `downlevelIteration`

demo/src/actions.ts
domasx2: fixed, `import '@grafana/agent-web/globals';` :)

.gitignore
domasx2: reverted back to `dist`

packages/core/tsconfig.all.json
domasx2: Actually I think current situation is ok. Nothing wrong with default being `cjs`, adding extra file does not really change anything

packages/tsconfig.json
domasx2: restored

bfmatei (dismissed):

kpelelis (approved):

PR #30, opened 2022-06-17
Add agent.pause() and agent.unpause()

domasx2:
Adds ability to pause the agent, preventing events from being sent to transports.
Can be used when you do not want to track certain pages in your SPA.
Alternative to https://github.com/grafana/grafana-javascript-agent/pull/29

### DIFF
--- a/packages/core/src/initialize.ts
+++ b/packages/core/src/initialize.ts
@@ -12,11 +12,16 @@ export function initializeAgent(config: Config): Agent {
   const transports = initializeTransports(config);
   const api = initializeAPI(config, transports, metas);
 
+  const pause = () => transports.pause();
+  const unpause = () => transports.unpause();
+
   const agent = initializeGlobalAgent({
     config,
     metas,
     transports,
     api,
+    pause,
+    unpause
   });
 
   if (!agent.config.preventGlobalExposure) {
--- a/packages/core/src/initialize.ts
+++ b/packages/core/src/initialize.ts
@@ -21,7 +21,7 @@ export function initializeAgent(config: Config): Agent {
     transports,
     api,
     pause,
-    unpause
+    unpause,
   });
 
   if (!agent.config.preventGlobalExposure) {
--- a/packages/core/src/transports/initialize.ts
+++ b/packages/core/src/transports/initialize.ts
@@ -7,7 +7,7 @@ import { BeforeSendHook, Transport, TransportItemType, Transports } from './type
 export function initializeTransports(config: Config): Transports {
   const transports: Transport[] = [...config.transports];
 
-  let paused = config.paused ?? false;
+  let paused = config.paused;
 
   const beforeSendHooks: BeforeSendHook[] = [];
   if (config.beforeSend) {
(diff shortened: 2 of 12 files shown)


packages/core/src/initialize.ts
```diff
@@ -12,11 +12,16 @@ export function initializeAgent(config: Config): Agent {
...
+  const pause = () => transports.pause();
+  const unpause = () => transports.unpause();
+
   const agent = initializeGlobalAgent({
     config,
     metas,
     transports,
     api,
+    pause,
+    unpause
```
bfmatei, line 24: Missing comma

packages/core/src/initialize.ts
```diff
@@ -12,11 +12,16 @@ export function initializeAgent(config: Config): Agent {
   const transports = initializeTransports(config);
   const api = initializeAPI(config, transports, metas);
 
+  const pause = () => transports.pause();
+  const unpause = () => transports.unpause();
+
   const agent = initializeGlobalAgent({
     config,
     metas,
     transports,
```
bfmatei, line 21: Should we converts `transports` to an object and move `pause`/`unpause` methods there? It would make sense to me since that's what we are pausing.

domasx2: It is an object with `pause` / `unpause` :thinking:  I just wanted to have `pause()` method at top level `agent` as well for discoverability

bfmatei (approved):

domasx2:
also added ability to initialize paused, updated docs

packages/core/src/transports/initialize.ts
```diff
@@ -7,7 +7,7 @@ import { BeforeSendHook, Transport, TransportItemType, Transports } from './type
 export function initializeTransports(config: Config): Transports {
   const transports: Transport[] = [...config.transports];
 
-  let paused = false;
+  let paused = config.paused ?? false;
```
bfmatei, line 10: At this point, the config should already contain the default values to avoid nullish coalescing.

packages/web/src/config.ts
```diff
@@ -20,9 +22,10 @@ import { browserMeta, pageMeta } from './metas';
...
 
 export interface BrowserConfig {
+  app: App;
+
   url?: string;
   apiKey?: string;
-  app: App;
   session?: Session;
   user?: User;
   globalObjectKey?: string;
```
bfmatei, line 31: I think this object becomes a little harder to follow. We have metas, agent state flags (paused), configs etc.

domasx2: well, "advanced" stuff is optional :thinking:  what do you propose? 
If it's about the interface definition, I can maybe get clever and pull most of these from the core config interface

bfmatei: I was thinking about child objects, similarly to what i proposed for the API (`pause` and `unpause` moved on transports etc.).

packages/core/src/transports/initialize.ts
domasx2: updated

packages/web/src/config.ts
domasx2: I'm not a fan of nested config objects tbh, I think keeping it as flat as possible is most user friendly - much easier to discover what you need eg just by type suggestions

PR #32, opened 2022-06-20
Fix: require semicolon

domasx2:
~~Config eslint to require semicolons~~

Adds `prettier --check` lint stage to everything, so CI caches mishaps.
Also prettier-formats e2e tests


bfmatei (commented):
TBH i would rather make prettier run in check mode during CI and a git hook. This, being a stylistic issue, would be caught by it and also other checks would be performed.

demo/package.json
```diff
@@ -20,7 +20,8 @@
     "quality:test": "exit 0",
     "quality:format": "prettier --write \"./**/*.{js,jsx,ts,tsx,css,scss,md,yaml,yml,json}\"",
     "quality:lint": "run-s quality:lint:*",
-    "quality:lint:ts": "eslint \"src/**/*.ts\"",
+    "quality:lint:eslint": "eslint \"src/**/*.ts\"",
+    "quality:lint:prettier": "prettier --check src/**/*.ts src/*.ts",
```
bfmatei, line 24:
```suggestion
    "quality:lint:prettier": "prettier --check **/*.{js,jsx,ts,tsx,css,scss,md,yaml,yml,json}",
```
AFAIK `**/*.ts` is including the root folder as well, hence the removal of `src/*.ts`.

Also, it's good to lint other files as well, not only the `ts` files.

package.json
```diff
@@ -32,7 +32,8 @@
     "quality:format": "lerna run --parallel quality:format",
     "quality:lint": "run-p quality:lint:*",
     "quality:lint:packages": "lerna run --parallel quality:lint",
-    "quality:lint:root": "markdownlint README.md CHANGELOG.md docs/**/*.md",
+    "quality:lint:markdown": "markdownlint README.md CHANGELOG.md docs/**/*.md",
```
bfmatei, line 35: The root `package.json` should trigger the actions of sub-packages and at the same time run actions for the root files (outside of subp-packages) if needed.

I would suggest the following:
```
    "quality:lint:packages:markdown": "lerna run --parallel quality:lint:markdown",
    "quality:lint:root": "run-p quality:lint:root:*",
    "quality:lint:root:markdown": "markdownlint README.md CHANGELOG.md docs/**/*.md",
    "quality:lint:root:prettier": "prettier --check **/*.{js,jsx,ts,tsx,css,scss,md,yaml,yml,json} !packages/**/* !demo/**/*",
```

Otherwise we will end up with files being linted and other not being linted. I didn't check the implementation myself, i just wrote it from memory :D.

Also I can handle this on the branch directly if you want.

bfmatei (commented):
A couple of discussions.

demo/package.json
domasx2: `src/**/*.ts` does not files in src folder, tested.
Wont just leaving `**/*` also include `node_modules`, `build` and etc?

bfmatei: IIRC, if they are in `.prettierignore` they won't be checked (needs testing).

domasx2: done

demo/package.json
```diff
@@ -18,9 +18,10 @@
     "clean": "rimraf dist/ yarn-error.log",
     "quality": "run-s quality:*",
     "quality:test": "exit 0",
-    "quality:format": "prettier --write \"./**/*.{js,jsx,ts,tsx,css,scss,md,yaml,yml,json}\"",
+    "quality:format": "prettier --ignore-path ../.prettierignore --write \"./**/*.{js,jsx,ts,tsx,css,scss,md,yaml,yml,json}\"",
```
bfmatei, line 21: Just out of curiosity... `ignore-path` is required but `--config` ain't?

.prettierignore
```diff
@@ -13,3 +13,6 @@
 coverage
 dist
 node_modules
+
+# drone sign step formats it in a way that prettier doesn't like
+.drone/drone.yml
```
bfmatei, line 18: Is the signature applied to the whitespace as well?

Another idea is to run `drone sign` w/o the `--save` flag and just manually update the drone config file. It wouldn't update that often to actually matter a copy/paste process after signature generation.

bfmatei (approved):
Nitpicks here, nitpicks there.

demo/package.json
domasx2: Each package has it's own prettier config that inherits from root config, so doesn't have to be specified explicitly.
But each package uses root prettierignore, because it can't be extended (afaik), and don't wanna copypaste

.prettierignore
domasx2: I don't think that prettier's opinions on yaml formatting are worth adidng manual steps to it :) 
This config is auto formatted by signing anyhow, maybe it's fine

bfmatei: The problem is that some people do have prettier set to auto-format a file when it is opened by the IDE, hence causing corrupted files. But the question still stands... is whitespace considered when doing the signature?

demo/package.json
bfmatei: I run some tests locally now. Prettier does a lookup for a the config file as it is not in every package file but it ain't doing it for the ignore file. So it is ok like it is now.

bfmatei (approved):

.prettierignore
domasx2: yeah, formatting matters, the whole text block is signed.. how bout I add a .prettierignore with drone config at package level?

PR #41, opened 2022-07-11
Implement debugging

bfmatei:
This PR brings an internal debugging mechanism to the JavaScript agent.


packages/web/src/initialize.ts
```diff
@@ -1,8 +1,12 @@
...
 
-export function initializeAgent(config: BrowserConfig): Agent {
+export function initializeGrafanaAgent(config: BrowserConfig): Agent {
   const coreConfig = makeCoreConfig(config);
+
   return coreInit(coreConfig);
 }
+
+// TODO: Remove this alias after the updating the projects where we dogfood
+export const initializeAgent = initializeGrafanaAgent;
```
domasx2, line 12: maybe wrap with a function that logs a deprecation notice?

domasx2 (approved):
some small nits, but overall lgtm!

packages/core/src/originalConsole/types.ts
```diff
@@ -0,0 +1 @@
+export type OriginalConsole = Console;
```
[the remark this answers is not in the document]
bfmatei: Went w/ this approach for future safety (i.e. `Omit<Console, 'group' | 'groupEnd'>`). This may be needed in the future.

domasx2 (approved):

cypress/e2e/demo/errors.cy.ts
```diff
@@ -1,10 +1,12 @@
 context('Errors', () => {
   function checkErrorReported(type: string, value: string, expectStacktrace = true) {
-    cy.waitExceptions((events) => {
-      expect(events).to.have.lengthOf(1);
-      const exception = events[0];
+    cy.waitExceptions((evts) => {
```
domasx2, line 3: did the word `event` piss off a linter? :D

demo/src/index.ts
```diff
@@ -3,7 +3,7 @@ import { TracingInstrumentation } from '@grafana/agent-tracing-web';
 import { initializeGrafanaAgent, getWebInstrumentations } from '@grafana/agent-web';
 
 initializeGrafanaAgent({
-  url: 'http://localhost:12345/collect',
+  url: '/collect',
```
domasx2, line 6: wut, same host? :thinking:

cypress/e2e/demo/errors.cy.ts
bfmatei: Neah. Mass-replaced "event" with "evt", including for the plurals. It's a good practice to avoid using `event` as it is also a global variable hence it can cause errors (i.e. removing the parameter from a function and the build would just work).

demo/src/index.ts
bfmatei: For the time being, the demo also requires the receiver to be installed locally but that's not necessary unless you want to really check the data somewhere (and most of the times I don't as I just check the payloads) that are sent via `fetch`.

I added a proxy configuration (see `.proxyrc.js`) that responds with `200` to all the requests.

bfmatei: I think we can improve this a lot when we refactor the demo 🤔 .

PR #44, opened 2022-07-29
Limit fetch transport concurrency

### LINKED ISSUE #37: Fetch transport: limit concurrent requests
In case requests are hanging or taking a long time we don't want to exhaust page's connection pool. Limit outgoing requests to a sensinble max concurrency

domasx2:
This adds a PromiseBuffer implementation that can buffer "tasks" (promise factories), and execute them while limiting concurrency. It's possible to configure max uncompleted tasks to accept, as well as the concurrency factor. `FetchTransport` then makes use of it.

fixes #37


packages/core/src/utils/promiseBuffer.ts
```diff
@@ -0,0 +1,71 @@
...
+
+          reject(reason);
+        }
+      );
+    }
+  };
+
+  const add: PromiseBuffer<T>['add'] = (promiseProducer) => {
+    if (buffer.length + inProgress >= size) {
+      throw new Error('Task buffer full');
```
bfmatei, line 55: Mhm, I would suggest to use the `internalLogger` instead of throwing an Error which can have unwanted side-effects.

bfmatei (approved):
Minor suggestion

packages/core/src/utils/promiseBuffer.ts
domasx2: the intention here it should be possible to determine that task was not accepted and do something about it. So it is captured and logged by caller instead

PR #50, opened 2022-08-30
Add support for React

bfmatei:
The initial PR for supporting React.

The following features are added:
- Error Boundary - it's a wrapper component that catches internal React issues. It can also be used with `withErrorBoundary` HOC
- Router - it depends on the version of the Router (v4/v5 vs v6)
- Tracer - It catches the mount time, all the re-renders and the entire render time

For the time being the PR is just draft (no docs etc.). WDYT?


packages/integration-react/src/router/initialize.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import type { ReactIntegrationConfig } from '../types';
+import { ReactRouterVersion } from './types';
+import { initializeReactRouterV4V5Instrumentation } from './v4v5';
+import { initializeReactRouterV6Instrumentation } from './v6';
+
+export function initializeReactRouterInstrumentation(config: ReactIntegrationConfig, agent: Agent): void {
+  if (!agent.api.isOTELInitialized()) {
+    agent.internalLogger.error(
+      'The Grafana React Router instrumentation requires tracing instrumentation. Please enable it in the "instrumentations" section of your config.'
```
domasx2, line 11: it's also possible to config otel separatetly and then call `agent.api.initOTEL`

domasx2: even if no tracing, we can (and should?) use `agent.api.pushEvent` to create a `navigation` event?

domasx2 (commented):
looks good! some small "how to" doc would be great to go with it

packages/integration-react/src/router/initialize.ts
bfmatei: Yeah. Right now it uses the events API :D. Gonna remove this check.

domasx2 (approved):

PR #52, opened 2022-09-08
Session instrumentation

domasx2:
* Web SDK initializes with autogenerated session id
* `Metas` supports adding listener callbacks that are invoked when meta changes. This is to support sending new session start event if session changes. Prior art: Sentry does this in the same way for the same purpose :) 
* `SessionInstrumentation` sends `session_start` event on init and every time new session is added to meta
* `web-tracing` package gains `SessionProcessor` pass-through processor that will add `session_id` attribute to every span, if session is available

Only generating default session id for web, not core. Do not want to assume how sessions might look for other platforms

### DIFF
--- a/docs/sources/tutorials/quick-start-browser.md
+++ b/docs/sources/tutorials/quick-start-browser.md
@@ -285,6 +285,15 @@ agent.api.setUser({
   }
 });
 
+// unset user
+agent.api.resetUser();
+
+// set session metadata, to be included with every event
+agent.api.setSession(createSession({ plan: 'paid' }));
+
+// unset session
+agent.api.resetSession();
+
 // push measurement
 agent.api.pushMeasurement({
   type: 'cart-transaction',
--- a/docs/sources/tutorials/quick-start-browser.md
+++ b/docs/sources/tutorials/quick-start-browser.md
@@ -130,6 +130,7 @@ import {
   initializeGrafanaAgent,
   LogLevel,
   WebVitalsInstrumentation,
+  SessionInstrumentation,
 } from '@grafana/agent-web';
 
 const agent = initializeGrafanaAgent({
@@ -139,6 +140,7 @@ const agent = initializeGrafanaAgent({
     new ConsoleInstrumentation({
       disabledLevels: [LogLevel.TRACE, LogLevel.ERROR], // console.log will be captured
     }),
+    new SessionInstrumentation(),
   ],
   transports: [
     new FetchTransport({
@@ -206,7 +208,7 @@ import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base';
 import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
 import { initializeGrafanaAgent } from '@grafana/agent-web';
-import { GrafanaAgentTraceExporter } from '@grafana/agent-trace-web';
+import { GrafanaAgentTraceExporter, SessionProcessor } from '@grafana/agent-trace-web';
 
 const VERSION = '1.0.0';
 const NAME = 'frontend';
@@ -231,7 +233,7 @@ const resource = Resource.default().merge(
 );
 
 const provider = new WebTracerProvider({ resource });
-provider.addSpanProcessor(new BatchSpanProcessor(new GrafanaAgentTraceExporter({ agent })));
+provider.addSpanProcessor(new SessionSpanProcessor(new BatchSpanProcessor(new GrafanaAgentTraceExporter({ agent }))));
 provider.register({
   propagator: new W3CTraceContextPropagator(),
   contextManager: new ZoneContextManager(),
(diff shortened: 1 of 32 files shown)


docs/sources/tutorials/quick-start-browser.md
```diff
@@ -206,7 +208,7 @@ import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base';
 import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
 import { initializeGrafanaAgent } from '@grafana/agent-web';
-import { GrafanaAgentTraceExporter } from '@grafana/agent-trace-web';
+import { GrafanaAgentTraceExporter, SessionProcessor } from '@grafana/agent-trace-web';
```
bfmatei, line 211: Should we name the `SessionProcessor` similarly to `GrafanaAgentTraceExporter` to keep naming consistent?

packages/core/src/api/meta/initialize.ts
```diff
@@ -6,15 +6,34 @@ import type { MetaAPI } from './types';
...
 
-  let meta: Meta = {};
+  let metaSession: Partial<Meta> | undefined = undefined;
+  let metaUser: Partial<Meta> | undefined = undefined;
 
-  metas.add(() => meta);
+  const setUser: MetaAPI['setUser'] = (user?: MetaUser) => {
+    if (metaUser) {
+      metas.remove(metaUser);
+    }
```
bfmatei, line 15:
```suggestion
    }

```

packages/core/src/api/meta/initialize.ts
```diff
@@ -6,15 +6,34 @@ import type { MetaAPI } from './types';
...
+    metaUser = {
+      user,
+    };
```
bfmatei, line 18:
```suggestion
    };

```

packages/core/src/api/meta/initialize.ts
```diff
@@ -6,15 +6,34 @@ import type { MetaAPI } from './types';
...
 
-  const setUser: MetaAPI['setUser'] = (user: MetaUser | null) => {
-    meta = { ...meta, user: user ?? undefined };
+  const setSession: MetaAPI['setSession'] = (session) => {
+    if (metaSession) {
+      metas.remove(metaSession);
+    }
+    metaSession = {
+      session,
+    };
```
bfmatei, line 28:
```suggestion
    };

```

packages/core/src/api/meta/types.ts
```diff
@@ -1,5 +1,7 @@
-import type { MetaUser } from '../../metas';
+import type { MetaSession, MetaUser } from '../../metas';
 
 export interface MetaAPI {
-  setUser: (user: MetaUser | null) => void;
+  setUser: (user?: MetaUser) => void;
```
bfmatei, line 4: Just a wild suggestion: separate between `setUser` and `resetUser`? As a user of the library it's easier to refactor later (i.e. searching for `resetUser` rather than `setUser(null)` or `setUser(undefined)` etc.

packages/core/src/initialize.ts
```diff
@@ -26,6 +26,8 @@ export function initializeGrafanaAgent(config: Config): Agent {
   const transports = initializeTransports(internalLogger, config);
   const api = initializeAPI(internalLogger, config, transports, metas);
 
+  api.setSession(config.session);
```
bfmatei, line 29: shouldn't this be inside `initializeAPI`? Either like this or actually setting it in the initialization function of the metas API.

packages/core/src/metas/initialize.ts
```diff
@@ -2,19 +2,36 @@ import type { Config } from '../config';
...
 import { isFunction } from '../utils';
 import { VERSION } from '../version';
-import type { Meta, Metas, MetaItem } from './types';
+import type { Meta, Metas, MetaItem, MetasListener } from './types';
 
 export function initializeMetas(internalLogger: InternalLogger, config: Config): Metas {
   let items: MetaItem[] = [];
+  let listeners: MetasListener[] = [];
+
+  const _getValue = () => {
```
bfmatei, line 11: Underscore variables are usually the ones marked as "not used" rather than internal. Let's keep it camelcase.

bfmatei (changes requested):
Some nitpicks, some necessary changes.

packages/core/src/initialize.ts
domasx2: My thinking is that setting session is not the concern of _initializing the api itself_ , thus does not belong there. Not setting it in metas init, as I thought it  best to always set session using this setter

docs/sources/tutorials/quick-start-browser.md
domasx2: `GrafanaAgentTraceExporter` is named so because it exports *to* the grafana agent, not because it's `GrafanaAgent` related. Buut maybe it wouldn't hurt anyway to rename :thinking:

packages/core/src/api/meta/types.ts
domasx2: But then it will result in code like `if (user) { setUser(user); } else { removeUser() }` which isn't great either :thinking:  I think unless something is going terribly wrong, there should be a single `setUser` or `setSession` in the entire app anyway, so not a big deal to find

packages/core/src/initialize.ts
bfmatei: Ah, I was seeing at as "initialize with those (initial) values".

packages/core/src/api/meta/types.ts
domasx2: alright, done, both `resetSession` and `resetUser`

PR #54, opened 2022-09-19
bump lerna

domasx2:
bump lerna to get rid of the high severity vulernability alert for `parse-url`


bfmatei:
In my React PR i also updated all the dependencies, including `web-vitals` and others. I think we should close this one and focus on merging that one given the amount of changes there.

domasx2:
ok!

PR #55, opened 2022-09-21
Add backoff for fetch transport upon rate limiting

### LINKED ISSUE #45: Fetch transport: implement backoff when rate limiting response is received
If collector returns 429, back off for a while. Needs fleshing out details. Can we have the receiver provide backoff duration?

domasx2:
This updates `FetchTransport` to drop events for a period after receiving 429 too many requests. The period is taken from `Retry-After` header if present, or 5 seconds default. This should help with spam. 

fixes #45

### DIFF
--- a/packages/web/src/transports/fetch/transport.ts
+++ b/packages/web/src/transports/fetch/transport.ts
@@ -5,15 +5,22 @@ import type { FetchTransportOptions } from './types';
 
 const DEFAULT_BUFFER_SIZE = 30;
 const DEFAULT_CONCURRENCY = 5; // chrome supports 10 total, firefox 17
+const DEFAULT_RATE_LIMIT_BACKOFF_MS = 5000;
 
 export class FetchTransport extends BaseTransport {
   readonly name = '@grafana/agent-web:transport-fetch';
   readonly version = VERSION;
 
   promiseBuffer: PromiseBuffer<Response | void>;
+  private reateLimitBackoffMs: number;
+  private getNow: () => number;
+  private disabledUntil: Date = new Date();
 
   constructor(private options: FetchTransportOptions) {
     super();
+
+    this.reateLimitBackoffMs = options.defaultRateLimitBackoffMs ?? DEFAULT_RATE_LIMIT_BACKOFF_MS;
+    this.getNow = options.getNow ?? (() => Date.now());
 
     this.promiseBuffer = createPromiseBuffer({
       size: options.bufferSize ?? DEFAULT_BUFFER_SIZE,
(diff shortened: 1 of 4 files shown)


packages/web/src/transports/fetch/transport.ts
```diff
@@ -5,16 +5,23 @@ import type { FetchTransportOptions } from './types';
...
 const DEFAULT_BUFFER_SIZE = 30;
 const DEFAULT_CONCURRENCY = 5; // chrome supports 10 total, firefox 17
+const DEFAULT_RATE_LIMIT_BACKOFF_MS = 5000;
 
 export class FetchTransport extends BaseTransport {
   readonly name = '@grafana/agent-web:transport-fetch';
   readonly version = VERSION;
 
   promiseBuffer: PromiseBuffer<Response | void>;
+  private reateLimitBackoffMs: number;
```
connorlindsey, line 15: Type: reate -> rate

connorlindsey (approved):
lgtm

bfmatei (commented):
I added a couple of suggestions for vertical spacing. Mainly the following:
- `return` statements are usually separated by a blank line from other statements for an easier spotting when looking over the code.
- `if` and other blocks are usually separated by blank lines (one before and one after) for more readability
- in tests, statements that are preparing for the next `expect` statement should be grouped with that statement for improved readability ('those things are happening => expect this, than other things are happening => expect this new statement' etc.)
- private variables in classes should be grouped together as much as possible for easier code navigation
- in most cases, multiline statements are easier to read when they are surrounded by blank lines. It's easier to see where that statement begins and where it ends.

packages/web/src/transports/fetch/types.ts
```diff
@@ -6,11 +6,18 @@ export interface FetchTransportOptions {
...
   // how many requests to execute concurrently
   concurrency?: number;
-  // will be added as `x-api-key` header
-  apiKey?: string;
+  // if rate limit response does not include a Retry-After header,
+  // how many milliseconds to back off before attempting a request.
+  // intermediate events will be dropped, not buffered
+  defaultRateLimitBackoffMs?: number;
+  // get current date
+  getNow?: () => number;
```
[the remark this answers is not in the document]
domasx2: This is a provision for tests, so we can control the clock. Can extract, but would rather not export

packages/web/src/transports/fetch/transport.ts
domasx2: fixed!

bfmatei (approved):
lgtm

PR #56, opened 2022-09-23
Implement deduplication filter

bfmatei:
This PR brings deduplication capabilities in the agent. The functionality can be toggled via the `config` and the API has flags for force pushing the event (useful in some cases).


packages/core/README.md
```diff
@@ -24,6 +24,7 @@ The agent requires a configuration parameter with the following properties:
 | Property                | Description                                                             | Type                  | Default Value Variable                                   |
 | ----------------------- | ----------------------------------------------------------------------- | --------------------- | -------------------------------------------------------- |
 | `app`                   | Application metadata                                                    | `App`                 |                                                          |
+| `dedupe`                | A flag for toggling deduplication filter                                | `boolean`             |                                                          |
```
connorlindsey, line 27: Should we call out the default value here?

connorlindsey (commented):
lgtm but will let someone else with more agent experience approve

domasx2:
LGTM, 
But - consider alternative approach, creating `beforeSendHook`, like we do with `ignorePatterns`: https://github.com/grafana/grafana-javascript-agent/blob/7f3c0d16d2485cf91310ad25decfd839ad351e9e/packages/core/src/transports/initialize.ts#L66

Then dedupe impl is in a single place in a composable manner. Maybe we could have also instead of `beforeSendHook` had `middlewares: Middleware[]` :thinking:

packages/core/README.md
bfmatei: In the core package there's no default :-/. Although `undefined == false`, one can't initialize the core package without manually providing this flag.

domasx2 (approved):

PR #60, opened 2022-10-17
Remove empty log file to avoid tracking changes of it

cedricziel:
The empty log file still allowed git to track changes. This removes it while keeping the directory.


bfmatei (approved):
LGTM!

PR #68, opened 2022-11-10
Allow multiple instances of same transport type

domasx2:
I wanted to add two instances of `FetchTransport` to my side project, one to sent to cloud endpoint, one to an agent instance.
It didn't work, as `transports.add` deduplicates transports by name, which is the same for all instances of `FetchTransport`.

I'm certain that adding multiple instances of `FetchTransport` is a legit use case, updated to dedupe by reference only. 

Also updated `transports.remove`. Previously it would remove the last instance with the same name, and would not remove it if it was first in the transport list -  `0` is `false` :)

### DIFF
--- a/packages/core/src/transports/initialize.ts
+++ b/packages/core/src/transports/initialize.ts
@@ -41,7 +41,7 @@ export function initializeTransports(internalLogger: InternalLogger, config: Con
     newTransports.forEach((newTransport) => {
       internalLogger.debug(`Adding "${newTransport.name}" transport`);
 
-      const exists = transports.some((existingTransport) => existingTransport.name === newTransport.name);
+      const exists = transports.some((existingTransport) => existingTransport === newTransport);
 
       if (exists) {
         internalLogger.warn(`Transport ${newTransport.name} is already added`);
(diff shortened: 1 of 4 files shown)


packages/core/src/transports/initialize.ts
```diff
@@ -111,18 +111,9 @@ export function initializeTransports(internalLogger: InternalLogger, config: Con
     transportsToRemove.forEach((transportToRemove) => {
       internalLogger.debug(`Removing "${transportToRemove.name}" transport`);
 
-      const existingTransportIndex = transports.reduce<number | null>(
-        (acc, existingTransport, existingTransportIndex) => {
-          if (acc === null && existingTransport.name === transportToRemove.name) {
-            return existingTransportIndex;
-          }
+      const existingTransportIndex = transports.indexOf(transportToRemove);
```
bfmatei, line 114: Wouldn't it be better with `.find` instead of `indexOf` given that we don't use the index anywhere?

domasx2: It's used on 122: `transports.splice(existingTransportIndex, 1);`

bfmatei (approved):
lgtm

PR #72, opened 2022-11-21
remove OTLP traces downgrade

domasx2:
We were downgrading traces OTLP JSON payload to older version of OTLP to support Grafana Agent, because it was using OTEL collector <0.55.
Grafana Agent 0.29.0 upgrades to OTEL collector 0.61.0 which switches to new version of OTLP payload for traces.

**Breaking change** This PR removes the traces payload downgrade, so now it supports Grafana Agent >= 0.29.0, but no longer supports < 0.29.0.


bfmatei (approved):
lgtm overall

PR #74, opened 2022-11-30
Fix broken link in readme 

sunker:
**What is this feature and why do we need it?**

Seems like the link to @grafana/faro-react was broken. This PR should fix that.


domasx2 (approved):

domasx2:
Thank you!

PR #76, opened 2022-12-07
Add CDN support

bfmatei:
**What is this feature and why do we need it?**

This PR aims to bring support for IIFE bundles so they can be loaded from CDNs.

The idea was to get the TS code, pass it through Rollup (which compiles the code and uses Terser to minify it) and publish the bundle along with the other packages. One can than use unpkg to serve the bundles. See an example usage at the bottom of this PR.

Besides this, there are a couple of other improvements:
- `fast-deep-equal` was replaced with an in-house approach (a lil' faster given the use case but also because it had issues with bundling).
- refactored a little bit the tsconfig files so the project
- added cache for eslint and prettier
- changed the build output so the types are bundled only once instead of twice (once for `esm` and once for `cjs`).
- fixed types for both Jest and Cypress
- prevent builds from including test files
- updated dependencies
- Both Cypress and Jest no longer need empty exports

**Checklist**:

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

```html
<!doctype html>
<html>
  <head>
    <script>
      // Initialize Faro when the main bundle loads
      window.init = () => {
        window.GrafanaFaroWebSdk.initializeFaro({
          app: {
            name: 'test',
          },
          transports: [new window.GrafanaFaroWebSdk.ConsoleTransport()],
        });
      }

      // Dynamically add the tracing instrumentation when the bundle loads
      window.addTracing = () => {
        window.GrafanaFaroWebSdk.faro.instrumentations.add(new window.GrafanaFaroWebTracing.TracingInstrumentation());
      }
    </script>
  </head>
  <body>
    <!-- Load the bundles from "somewhere" -->
    <script src="bundle/faro-web-sdk.iife.js" onload="window.init()"></script>
    <script src="bundle/faro-web-tracing.iife.js" onload="window.addTracing()"></script>
  </body>
</html>
```


packages/core/package.json
```diff
@@ -19,26 +19,32 @@
     "url": "https://github.com/grafana/faro-web-sdk.git",
     "directory": "packages/core"
   },
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
-  "types": "dist/cjs/index.d.ts",
+  "browser": "./dist/bundle/faro-core.iife.js",
```
domasx2, line 22: are you sure this is correct? AFAIK this field is for bundlers use, so propably should still be esm/cjs/amd, not iife.

packages/web-tracing/src/instrumentation.ts
```diff
@@ -29,7 +29,7 @@ export class TracingInstrumentation extends BaseInstrumentation {
   }
 
   initialize(): void {
-    const config = this.faro.config;
+    const config = faro.config;
```
domasx2, line 32: This change introdcues a problem if you have multiple instances of Faro, something we haven't fully solved yet.. 
Perhaps we should simply not allow multiple instances for simplicity sake?

domasx2 (commented):
Great work, only few minor concerns. Also I think needs changelog & docs (readme) update

domasx2 (approved):
:+1:

PR #84, opened 2022-12-20
Add view meta

bfmatei:
## Description

This PR brings the idea of view meta as some apps do not rely on URLs but rather on "views".

There is a "default" view that it set by the web-sdk and it can be changed similarly to how we do it with the session meta.

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated


domasx2 (commented):
LGTM!
but:
It adds new meta prop to payload, so needs backend support in both endpoint and Grafana agent integration

bfmatei:
@domasx2 - yup! Should we hold merging this until collector implements support for view? AFAIK the collector will just drop the `view` meta.

cc: @rlankfo @kpelelis

PR #92, opened 2023-01-11
Fix circular dependency in core package

### LINKED ISSUE #91: Circular dependency detected in core package
## Description

When building the IIFE bundle in `core` package, rollup complains about a circular dependency. Given that the IIFE bundle is loaded in the browser, it's best to fix this as other issues may result from this.

## Steps to reproduce

1. Run `yarn build`

## Expected behavior

No circular dependency errors should occur.

## Actual behavior

Circular dependency error occurs.

## Environment

- **SDK version:** -
- **SDK instrumentations:** -
- **Device type:** -
- **Device name:** -
- **OS:** -
- **Browser:** -

## Demo

-

## Context

-

bfmatei:
## Description

This PR fixes a circular dependency in the `core` package.

Besides this, it also updates the dependencies and adds some missing exports (check the changelog).

## Fixes

Fixes #91

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated


leveyjam (approved):
Looks good - I build and the warning is gone.

Q: do we have a convention around separate commits (or separate PRs) for the version upgrades? 
(I find it can be helpful for things like bisecting changes to find the commit which introduced an error/build failure, etc. if this was 2 PRs/commits instead of one!)

PR #101, opened 2023-01-23
Faro to Otel HTTP Payload transform - Part 1

codecapitano:
## Description

We want to be as compliant as possible to the OTel Payload schema. 
There for we want to introduce a new transport which converts Faro Transport Items to OTel payload. 

This PR is the first in a row which introduces the spec as well as the respective transforms to convert Faro Events, Errors, Measurements and Logs to ResourceLogs which are as compliant to OTel as we currently can.

* [OTel Semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification). Note: We use OTel Semantic Conventions where possible. [In cases there is no equivalent for a Faro attribute we prefixed the attribute with `grafana.` [discussion]](https://raintank-corp.slack.com/archives/C030T51HM2L/p1675351697590449).
* [OTel schema definitions](https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto)

Note:
I divided the implementation of the OtelHttpTransport in 3 parts.
* Part 1: is this PR which implements the Faro model to Otel http payload transforms
* Part 2: implements the transport #114 
* Part 3: implements the traces to Otel http payload transformations

## Open/Missing:
* [Add OTel compatible Traces transform](https://github.com/grafana/faro-web-sdk/issues/108)
* [Otel compatible Logs and Traces transport](https://github.com/grafana/faro-web-sdk/issues/109)
* [Align measurement payload to only be able to contain exactly one measurement](https://github.com/grafana/faro-web-sdk/issues/110)

## Checklist
- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


codecapitano:
# Logs Payload definition [draft]
This definition contains the spec to transport logs, events and errors via log resource.

Note: We also want to transport metrics with the logs resource, so this will be added next.

```json
{
  "resourceLogs": [
    {
      "resource": {
        "attributes": [
          // Taken from BROWSER CONVENTIONS. Some may differ from Faro, some only exist in Faro.
          // https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/browser/
          {
            "key": "browser.mobile", // OTEL SEMANTIC CONVENTIONS
            "value": { "boolValue": false }
          },
          {
            "key": "browser.name", // FARO
            "value": { "stringValue": "Firefox" }
          },
          {
            "key": "browser.version", // FARO
            "value": { "stringValue": "109.0" }
          },
          {
            "key": "browser.platform", // FARO
            "value": { "stringValue": "Mac OS 13.2" }
          },

          // Taken from TELEMENTRY CONVENTIONS. Some may differ from Faro, some only exist in Faro.
          // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#telemetry-sdk
          {
            "key": "telemetry.sdk.name", // OTEL SEMANTIC CONVENTIONS
            "value": { "stringValue": "@grafana/faro-core" }
          },
          {
            "key": "telemetry.sdk.language", // OTEL SEMANTIC CONVENTIONS
            "value": { "stringValue": "javascript" } // or nodejs
          },
          {
            "key": "telemetry.sdk.version", // OTEL SEMANTIC CONVENTIONS
            "value": { "stringValue": "1.0.0-beta2" }
          },

          {
            "key": "session.id", // FARO
            "value": { "stringValue": "abcdefg" }
          },
          {
            "key": "session.attributes", // FARO
            "value": { "kvlistValue": [] }
          },

          // Taken from GENERAL IDENDITY ATTRIBUTES. Some may differ from Faro, some only exist in Faro.
          // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-identity-attributes
          {
            "key": "enduser.id", // OTEL SEMANTIC CONVENTIONS
            "value": { "stringValue": "119" }
          },
          {
            "key": "enduser.name", // FARO
            "value": { "stringValue": "119" }
          },
          {
            "key": "enduser.email", // FARO
            "value": { "stringValue": "119" }
          },
          {
            "key": "enduser.attributes", // FARO
            "value": { "kvlistValue": [] }
          },

          {
            "key": "app.name", // FARO
            "value": { "stringValue": "foo" }
          },
          {
            "key": "app.release", // FARO
            "value": { "stringValue": "6.6.7." }
          },
          {
            "key": "app.version", // FARO
            "value": { "stringValue": "1.23" }
          },
          {
            "key": "app.environment", // FARO
            "value": { "stringValue": "production" }
          }
        ],

        // We'll hardcode this to 0
        "droppedAttributesCount": 0
      },

      "scopeLogs": [
        {
          // The scope should originate from the instrumentation that captured this
          // For the time being we don't have this information
          // We will use the sdk as the source of information
          "scope": {
            "name": "@grafana/faro-core",
            "version": "1.0.0-beta4"
          },
          "logRecords": [
            // +++ EVENT +++
            {
              // We convert the timestamp to milliseconds and then multiply them by 1 million
              // The timestamp comes from the timestamp property on the payload
              "timeUnixNano": 1674464646296000000,
              // We'll use the same value from timestamp property on the payload
              "observedTimeUnixNano": 1674464646296000000,
              // We can just use 1 and TRACE as values for event signals
              "severityNumber": 1,
              "severityText": "TRACE",
              "body": {
                "stringValue": "click_button_with_attributes"
              },
              "attributes": [
                // This should be a convention to identify events from the other logs
                {
                  "key": "faro.event", // FARO
                  "value": {
                    "boolValue": true
                  }
                },
                // Taken from SEMANTIC CONVENTION FOR EVENT ATTRIBUTES. Some may differ from Faro, some only exist in Faro.
                // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/semantic_conventions/events.md
                {
                  "key": "event.name", // OTEL SEMANTIC CONVENTIONS
                  "value": {
                    "stringValue": "click_button_with_attributes"
                  }
                },
                {
                  "key": "event.domain", // OTEL SEMANTIC CONVENTIONS
                  "value": {
                    "stringValue": "browser"
                  }
                },
                {
                  "key": "event.attributes", // FARO
                  "value": {
                    "kvlistValue": [
                      {
                        "key": "foo",
                        "value": {
                          "stringValue": "bar"
                        }
                      },
                      {
                        "key": "baz",
                        "value": {
                          "stringValue": "bad"
                        }
                      }
                    ]
                  }
                },
                {
                  "key": "view.name", // FARO
                  "value": {
                    "stringValue": "default"
                  }
                },
                {
                  "key": "page.url", // FARO
                  "value": {
                    "stringValue": "http://localhost:5173"
                  }
                }
              ],
              // We hardcode it to 0 for now
              "droppedAttributesCount": 0,
              // This is optional, so we can omit it completely
              // "flags": 0,
              // The following are taken from the trace property
              // We can just omit them if they don't exist in the original payload
              "traceId": "582d698467d6d4edd62011c11335a13d",
              "spanId": "e0668e8ad9e6badc"
            },

            // +++ ERROR +++
            {
              // We convert the timestamp to milliseconds and then multiply them by 1 million
              // The timestamp comes from the timestamp property on the payload
              "timeUnixNano": 1674468151431000000,
              // We'll use the same value from timestamp property on the payload
              "observedTimeUnixNano": 1674468151431000000,
              // We can just use 21 and FATAL as values for errors signals
              "severityNumber": 21,
              "severityText": "FATAL",
              "body": {
                "stringValue": "This is a thrown error"
              },
              "attributes": [
                // This should be a convention to identify errors from the other logs
                {
                  "key": "faro.error",
                  "value": {
                    "boolValue": true
                  }
                },
                {
                  "key": "error.stacktrace", // FARO
                  "value": {
                    "kvlistValue": [
                      {
                        "key": "filename",
                        "value": {
                          "arrayValue": {
                            "values": [
                              {
                                "stringValue": "http://localhost:5173/src/client/pages/Features/ErrorInstrumentation.tsx"
                              },
                              {
                                "stringValue": "http://localhost:5173/node_modules/.vite/deps/chunk-63FLCSTF.js?v=502c6d22"
                              },
                              {
                                "stringValue": "http://localhost:5173/node_modules/.vite/deps/@opentelemetry_context-zone.js?v=02a1e0ae"
                              }
                            ]
                          }
                        }
                      },
                      {
                        "key": "function",
                        "value": {
                          "arrayValue": {
                            "values": [
                              {
                                "stringValue": "throwError"
                              },
                              {
                                "stringValue": "HTMLUnknownElement.callCallback2"
                              },
                              {
                                "stringValue": "_ZoneDelegate.invokeTask"
                              }
                            ]
                          }
                        }
                      },
                      {
                        "key": "lineno",
                        "value": {
                          "arrayValue": {
                            "values": [
                              {
                                "intValue": 21
                              },
                              {
                                "intValue": 3674
                              },
                              {
                                "intValue": 460
                              }
                            ]
                          }
                        }
                      },
                      {
                        "key": "colno",
                        "value": {
                          "arraylue": {
                            "values": [
                              {
                                "intValue": 11
                              },
                              {
                                "intValue": 22
                              },
                              {
                                "intValue": 171
                              }
                            ]
                          }
                        }
                      }
                    ]
                  }
                },
                // Taken from Semantic CONVENTIONS FOR EXCEPTIONS. Some may differ from Faro, some only exist in Faro.
                // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/semantic_conventions/exceptions.md
                {
                  "key": "exception.message", // OTEL SEMANTIC CONVENTIONS
                  "value": {
                    "stringValue": "The exception message."
                  }
                },
                {
                  "key": "exception.stacktrace", // OTEL SEMANTIC CONVENTIONS
                  "value": {
                    "stringValue": "A stacktrace as a string in the natural representation for the language runtime. The representation is to be determined and documented by each language SIG."
                  }
                },
                {
                  "key": "exception.type", // OTEL SEMANTIC CONVENTIONS
                  "value": {
                    "stringValue": "The type of the exception (its fully-qualified class name, if applicable). The dynamic type of the exception should be preferred over the static type in languages that support it."
                  }
                },

                {
                  "key": "view.name", // FARO
                  "value": {
                    "stringValue": "default"
                  }
                },
                {
                  "key": "page.url", // FARO
                  "value": {
                    "stringValue": "http://localhost:5173"
                  }
                }
              ],
              // We hardcode it to 0 for now
              "droppedAttributesCount": 0,
              // This is optional, so we can omit it completely
              // "flags": 0,
              // The following are taken from the trace property
              // We can just omit them if they don't exist in the original payload
              "traceId": "33cd71f73e44814794b2472917c71f81",
              "spanId": "00fdd358b506931a"
            },

            // +++ LOG +++
            {
              // We convert the timestamp to milliseconds and then multiply them by 1 million
              // The timestamp comes from the timestamp property on the payload
              "timeUnixNano": 1674472561805000000,
              // We'll use the same value from timestamp property on the payload
              "observedTimeUnixNano": 1674472561805000000,
              // For severity we can have the following convention:
              // console.log - 10 / INFO2 -> this is a special case
              // console.trace - 1 / TRACE
              // console.debug - 5 / DEBUG
              // console.info - 9 / INFO
              // console.warn - 13 / WARN
              // console.error - 17 / ERROR
              "severityNumber": 10,
              "severityText": "INFO2",
              "body": {
                "stringValue": "Faro was initialized"
              },
              "attributes": [
                {
                  "key": "foo",
                  "value": {
                    "stringValue": "bar"
                  }
                },
                // This should be a convention to Faro logs from other logs
                {
                  "key": "faro.log",
                  "value": {
                    "boolValue": true
                  }
                },

                {
                  "key": "view.name",
                  "value": {
                    "stringValue": "default"
                  }
                },
                {
                  "key": "page.url",
                  "value": {
                    "stringValue": "http://localhost:5173"
                  }
                }
              ],
              // Just hardcode it to 0 for now
              "droppedAttributesCount": 0
              // This is optional, so we can omit it completely
              // "flags": 0,
              // The following are taken from the trace property
              // We can just omit them if they don't exist in the original payload
              // "traceId": "",
              // "spanId": ""
            }
          ]
        }
      ]
    }
  ]
}
```

codecapitano:
@bfmatei @domasx2 

**Proposal:** 
let's prefix all attributes which are defined by faro / which are not defined in OTELs Semantic Attributes with `faro`.

**Pros:**
With this we make it explicit that theses attributes are not part of the OTEL standard.

**Cons:** 
If an attribute would with the same name (without the `faro` prefix) we need extra work to introduce the Otel compliant attribute and deprecate the respective faro specific attribute.   

Like: `faro.foo.bar` 

Wdyt?

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,174 @@
...
+      if (!isNaN(date)) {
+        return new Date(date);
+      }
+    }
+
+    return new Date(now + this.rateLimitBackoffMs);
+  }
+}
+
+interface OtlpTransportBody {
```
domasx2, line 170: note that OTLP specifies that traces, logs and metrics have each to be sent to separate endpoints, see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,174 @@
...
+  resourceMetrics: MeasurementEvent[];
```
domasx2, line 171: I think this is the correct metric type: https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/sdk-metrics/src/export/MetricData.ts#L69

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,174 @@
...
+  resourceLogs: (LogEvent | ExceptionEvent | EventEvent)[];
+  resourceSpans: TraceEvent[];
```
domasx2, line 173: I think this is the correct type https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/otlp-transformer/src/trace/types.ts#L27

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,174 @@
...
+      }
+    }
+
+    return new Date(now + this.rateLimitBackoffMs);
+  }
+}
+
+interface OtlpTransportBody {
+  resourceMetrics: MeasurementEvent[];
+  resourceLogs: (LogEvent | ExceptionEvent | EventEvent)[];
```
domasx2, line 172: Not sure that there are typescript types for otlp logs yet :/ might need to be generated from protobuf

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,174 @@
...
+  private readonly getNow: () => number;
+
+  private readonly sendBatchSize: number;
+  private readonly sendTimeoutMs: number;
+
+  private readonly signalsBuffer: TransportItem<APIEvent>[] = [];
+
+  promiseBuffer: PromiseBuffer<Response | void>;
+
+  constructor(private options: OtlpTransportOptions) {
```
domasx2, line 40: otel js implementation for traces: https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/exporter-trace-otlp-http
otel js implementation for metrics: https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-exporter-metrics-otlp-http/src/platform/browser/OTLPMetricExporter.ts

Probably hard to incorporate these, but if insipiration is needed :)

domasx2 (commented):
couple of notes re otlp

domasx2:
Re `faro.*` - not sure, haven't seen it done this way. Might be ok to just add non standard attrs. I suggest to bring this up with opentelemetry folks though, they will have more perspective

packages/web-sdk/src/transports/otlp/transform/semanticResourceAttributes.ts
```diff
@@ -0,0 +1,39 @@
...
+  BROWSER_LANGUAGE: 'browser.language', // TODO: Q: shall we add this to meta.ts => window.navigator.language
+} as const;
+
+export const faroResourceAttributes = {
+  BROWSER_MOBILE: SemanticBrowserAttributes.BROWSER_MOBILE, // browser.mobile
+  BROWSER_NAME: 'browser.name', // FARO
+  BROWSER_VERSION: 'browser.version', // FARO
+  BROWSER_PLATFORM: SemanticBrowserAttributes.BROWSER_PLATFORM, // 'browser.platform'
+
+  TELEMETRY_SDK_NAME: SemanticResourceAttributes.TELEMETRY_SDK_NAME, // 'telemetry.sdk.name',
```
codecapitano, line 23: Note:
Unfortunately `SemanticResourceAttributes` and `SemanticAttributes` are all of type string. 
The object which contains the real values is auto generated. 

This means using widens the derive type to string instead of a union of the actual values. 
So maybe we should use the respective string values instead of retrieving the form the npm module.

packages/web-sdk/src/transports/otlp/transform/Resource.ts
```diff
@@ -0,0 +1,58 @@
...
+import { faroResourceAttributes } from './semanticResourceAttributes';
+import type { Attribute, FaroResourceAttributes, PayloadResourceChild } from './types';
+
+type PayloadResource = {
+  resource: {
+    attributes: Attribute<FaroResourceAttributes>[];
+    droppedAttributesCount: number;
+  };
+};
+export class Resource implements PayloadResourceChild<PayloadResource> {
```
codecapitano, line 14: Note:
Not decided right now if this stays a class or if I convert it to
a simple function which returns the JSON directly.

packages/web-sdk/src/transports/otlp/proposal-logs-payload-1.json
```diff
@@ -0,0 +1,465 @@
...
+              // We can just use 1 and TRACE as values for event signals
+              "severityNumber": 1,
+              "severityText": "TRACE",
+              "body": {
+                "stringValue": "click_button_with_attributes"
+              },
+              "attributes": [
+                // This should be a convention to identify events from the other logs
+                {
+                  "key": "faro.event", // FARO
```
domasx2, line 116: IMHO unnecessary to specify that event is an event :-) exceptions/measurements can be extracted by `event.name`

packages/web-sdk/src/transports/otlp/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  defaultRateLimitBackoffMs?: number;
+
+  // get current date. for mocking purposes in tests
+  getNow?: ClockFn;
+
+  // addition options for global.Fetch
+  requestOptions?: OtlpTransportRequestOptions;
+}
+
+export type ClockFn = () => number;
```
bfmatei, line 36: This should be moved to a shared type and be used in both transports

packages/web-sdk/src/transports/otlp/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  apiKey?: string;
+
+  // how many requests to buffer in total
+  bufferSize?: number;
+
+  // how many signals to send in one request
+  sendBatchSize?: number;
+
+  // how many seconds to wait till we send the the signals
+  timeout?: number;
```
bfmatei, line 19: Should we name this `debounceTimeout` or something similar?

packages/web-sdk/src/transports/otlp/types.ts
```diff
@@ -0,0 +1,36 @@
+export interface OtlpTransportRequestOptions extends Omit<RequestInit, 'body' | 'headers'> {
```
bfmatei, line 1: This can be in a shared type

packages/web-sdk/src/transports/otlp/transport.ts
bfmatei: We also don't send `resourceMetrics` for the time being.

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,199 @@
...
+  //     if (!isNaN(date)) {
+  //       return new Date(date);
+  //     }
+  //   }
+
+  //   return new Date(now + this.rateLimitBackoffMs);
+  // }
+}
+
+interface OtlpTransportBody {
```
bfmatei, line 195: Each of those should be optional (`resourceLogs?:`).

packages/web-sdk/src/transports/otlp/transport.test.ts
```diff
@@ -0,0 +1,7 @@
+describe.skip('FetchTransport', () => {
+    // TODO:
+
+    // beforeEach(() => {
+    //   fetch.mockClear();
+    // });
+})
```
bfmatei, line 7: Don't forget an empty line at the end.

packages/web-sdk/src/transports/otlp/transform/ErrorLogRecord.ts
```diff
@@ -0,0 +1,22 @@
+import type { LogEvent, TransportItem } from 'packages/core/dist/types';
```
bfmatei, line 1: Those should be `from '@grafana/faro-core'`.

packages/web-sdk/src/transports/otlp/transform/ErrorLogRecord.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { ExceptionEvent } from 'packages/core/src/api';
+import type { Attribute, PayloadMember } from './types';
+
+export type ErrorLogRecordPayload = {
```
bfmatei, line 5: We prefer interfaces in favor of types.

packages/web-sdk/src/transports/otlp/transform/LogLogRecord.test.ts
```diff
@@ -0,0 +1,64 @@
+import { LogEvent, LogLevel, TransportItem, TransportItemType } from '@grafana/faro-core';
+import { LogLogRecord } from './LogLogRecord';
+
+const item: TransportItem<LogEvent> = {
```
bfmatei, line 4: If we define them outside of a `beforeEach`, it's best to declare them as constant objects:
```ts
const example = {
  // ... props
} as const;
```

packages/web-sdk/src/transports/otlp/transform/LogRecordFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+import { LogLogRecord } from './LogLogRecord';
+import { MeasurementLogRecord } from './MeasurementLogRecord';
+
+import type { LogTransportItem } from './types';
+
+export type LogRecord = LogLogRecord | ErrorLogRecord | EventLogRecord | MeasurementLogRecord;
+
+export class LogRecordFactory {
+  static getNewLogRecord(item: LogTransportItem): LogRecord {
+    if (isLogEventTransport(item)) {
```
bfmatei, line 14: Switch this to a `switch`

packages/web-sdk/src/transports/otlp/transform/LogRecordFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+      return new ErrorLogRecord(item);
+    }
+    if (isEventEventTransport(item)) {
+      return new EventLogRecord(item);
+    }
+    if (isMeasurementEventTransport(item)) {
+      return new MeasurementLogRecord(item);
+    }
+
+    throw new Error(`Unknown TransportItemType: ${item.type}`);
```
bfmatei, line 27: Use the `internalLogger` instead of throwing errors the JS way.

packages/web-sdk/src/transports/otlp/transform/Resource.test.ts
```diff
@@ -0,0 +1,247 @@
...
+        key: 'app.release',
+        value: { stringValue: 'app-v1.23' },
+      },
+    ],
+    droppedAttributesCount: 0,
+  },
+} as const;
+
+describe('Resource', () => {
+  test('Builds resource payload object for given transport item.', () => {
```
bfmatei, line 170: We use `it` instead of `test`

packages/web-sdk/src/transports/otlp/transform/attributeUtils.ts
```diff
@@ -0,0 +1,39 @@
+import type { MetaAttributes } from 'packages/core/dist/types';
+import type { AttributeTypes, FaroResourceAttributes } from './types';
+
+export const attributeValueType = {
```
bfmatei, line 4: We prefer `enums` wherever possible

packages/web-sdk/src/transports/otlp/transform/attributeUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+  kvList: 'kvListValue',
+} as const;
+
+export function toAttribute<T>(
+  attributeName: T,
+  attributeValue: any,
+  attributeType: AttributeTypes = attributeValueType.string
+) {
+  if (attributeValue == null || attributeValue === '') {
+    return;
```
bfmatei, line 16: It's best to add something to the `return`. I.E. `return undefined`

packages/web-sdk/src/transports/otlp/transform/attributeUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+  bool: 'boolValue',
+  string: 'stringValue',
+  kvList: 'kvListValue',
+} as const;
+
+export function toAttribute<T>(
+  attributeName: T,
+  attributeValue: any,
+  attributeType: AttributeTypes = attributeValueType.string
+) {
```
bfmatei, line 14: All functions should have a return type

packages/web-sdk/src/transports/otlp/otlpPayloadLogger.ts
```diff
@@ -0,0 +1,3 @@
+import { createInternalLogger } from '@grafana/faro-core';
+
+export const internalLogger = createInternalLogger();
```
codecapitano, line 3: Wrong usage of logger.

Internal logger is part of the `transport` class and provided via `BaseExtension`.

Will add this in follow and remove this logger.

packages/web-sdk/src/transports/otlp/payload/transform/types.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface ResourcePayload {
+  attributes: Attribute[];
+}
+
+export interface ScopeLog {
+  scope: { name: string; version: string };
+  logRecords: LogRecordPayload[];
+}
+
+export type LogRecordPayload =
```
domasx2, line 14: I don't think it's worthile to disambiguate to payload types like this; all the properties are generic and potentially relevant

packages/web-sdk/src/transports/otlp/payload/transform/types.ts
```diff
@@ -0,0 +1,57 @@
+import type { APIEvent, TransportItem } from '@grafana/faro-core';
+
+import type { Attribute } from '../attribute';
+
+export interface ResourcePayload {
```
domasx2, line 5: I was looking for typescript types in the reference sdk. There's https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/api-logs/src/types/LogRecord.ts, but I'm not sure these work as export request types, and there's no log exporter yet in there

packages/web-sdk/src/transports/otlp/payload/OtelPayload.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  addResourceItem(transportItem: TransportItem): void {
+    const { type, meta } = transportItem;
+
+    switch (type) {
+      case TransportItemType.LOG:
+      case TransportItemType.EXCEPTION:
+      case TransportItemType.EVENT:
+      case TransportItemType.MEASUREMENT:
+        const resourceLogWithMeta = this.resourceLogsWithMetas.find(({ resourceLog }) => resourceLog);
```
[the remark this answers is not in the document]
codecapitano: Oh damn good catch. Yes it should find the item by the exact same meta.

packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
```diff
@@ -0,0 +1,199 @@
...
+export function getResourceLogPayload(transportItem: LogTransportItem) {
+  const resource = getResource(transportItem);
+
+  return {
+    resource,
+    scopeLogs: [getScopeLog(transportItem)],
+  };
+}
+
+function getResource(transportItem: LogTransportItem): Readonly<ResourcePayload> {
```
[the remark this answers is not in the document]
codecapitano: Much better!

packages/web-sdk/src/transports/otlp/payload/transform/types.ts
codecapitano: Hmmm thought about thsi to but added them because e. g. `LogLogRecordPayload` has a bunch of properties which `MeasurementLogRecord` doesn't have.
Also for types have different non-optional properties than others. 

Would you still put everything in one type and make all properties optional then?

codecapitano: You're right it does not provide much value to disambiguate here.
Changed it.

packages/web-sdk/src/transports/otlp/payload/OtelPayload.ts
```diff
@@ -0,0 +1,59 @@
...
+          resourceLog.scopeLogs.push(toScopeLog(transportItem));
+        } else {
+          this.resourceLogsWithMetas.push({
+            resourceLog: toResourceLog(transportItem),
+            meta,
+          });
+        }
+
+        break;
+      case TransportItemType.TRACE:
```
domasx2, line 52: what about traces?

packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
```diff
@@ -0,0 +1,192 @@
...
+  const body = toAttributeValue(payload.name) as { stringValue: string; key: string };
+
+  return {
+    timeUnixNano,
+    body,
+    attributes: [
+      ...getCommonLogAttributes(meta),
+      toAttribute('event.name', payload.name), // No prefix because this is a semantic attribute. But event.name constant is currently missing in sematic-conventions npm package
+      toAttribute('event.domain', payload.domain), // No prefix because this is a semantic attribute. But event.domain constant is currently missing in sematic-conventions npm package
+      toAttribute('grafana.event.attributes', payload.attributes),
```
domasx2, line 130: if we're gonna scope it, might make more sense to use `faro.event.attributes`, this is "faro" sdk after all.
BUT: why not just inline them as `event.[attr key]` each? I think this is what I'd expect as a user

domasx2 (approved):
looks good, just a few more comments

packages/web-sdk/src/transports/otlp/payload/OtelPayload.ts
codecapitano: Traces will be added with PR No 3.

packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
domasx2: So this is intended to be an unopinionated API for pushing events, not semantically Faro specific. I think it should allow providing arbitrary event attributes, and it should be up to the caller to decide on the attribute name. Eg, an instrumentation may want to provide a valid semantic OTEL attribute, but would be unable to because it is automatically prefixed. If we have Faro instrumentations creating Faro specific attributes, the instrumentation should be responsible for naming, not the core SDK. Does that make sense?

codecapitano: After a call with @domasx2 we decided to:
- Remove the prefixes from most attributes
- For some very Faro specific ones like `faro.error.stacktrace`, which has it's own data structure, we keep the prefix but change it to Faro.

Issue #90, opened 2023-01-11
OTLP compliant transport


bfmatei:
## Description

The current `FetchTransport` uses a proprietary schema for the payload which is not compliant with the OTLP schema.

## Proposed solution

The best approach to this issue is to create another HTTP transport, which we will refer to as `OtlpTransport`.

The `OtlpTransport` should be pretty similar to the current `FetchTransport`, but with an important difference: it should batch multiple events into a single HTTP request rather than sending an HTTP request for each signal. The batching should be configurable by number of signals send in a single request but also by the delay of waiting for a new signal before sending the current one.

For example, let's assume that whenever a signal is sent to a transport, it will wait for another `250ms` (this should be configurable) in order to wait for another signal. If no other signal came, the signal is sent. But if 30 (this should also be configurable) other signals are sent, then we should batch those 30 in a single request and any other subsequent signal will be part of the next batch.

The `OtlpTransport` should also implement the retrying mechanism and concurrency handling mechanism that are currently present in the `FetchTransport`.

As a stretch goal for the `OtlpTransport`, simple sampling should also be implemented. Simple sampling refers to dropping a percentage of the signals unless they are are from certain categories (i.e. errors, error logs or events). This should be disabled by default and it would offer a way for implementing proper sampling in the future.

The `OtlpTransport` will replace the `FetchTransport` as the default transport in `@grafana/faro-web-sdk` but `FetchTransport` should continue to exist for backwards compatibility and Grafana Agent compatibility purposes.

## Context

- [OTLP schema reference](https://github.com/open-telemetry/opentelemetry-proto/tree/main/opentelemetry/proto)

codecapitano:
1st PR: [Otel ResourceLogs Payload Transform](https://github.com/grafana/faro-web-sdk/pull/101)

codecapitano:
Open tasks/follow ups:

* [Add OTel compatible Traces transform](https://github.com/grafana/faro-web-sdk/issues/108)
* [Otel compatible Logs and Traces transport](https://github.com/grafana/faro-web-sdk/issues/109)
* [Align measurement payload to only be able to contain exactly one measurement](https://github.com/grafana/faro-web-sdk/issues/110)

PR #114, opened 2023-02-20
109-Otel Http Transport

codecapitano:
## Description

This is the second Part of #101. It implements the OtlpHttpTransport as specified in #90 
It transforms the Faro model to Otel compliant model and sends it to the configured endpoints.

### Note/Discussion
The `BatchTransport` acts as  a wrapper for the other transports. It is responsible for the batching (count and time based) as well as for flushing it's buffer on `visibilitychange`.
Transport which will be wrapped by the `BatchTransport` must implement or extend the `BatchBaseTransport` interface.
This ensures that the respective transports provide the `sendBatch` method.

**Usage**
```js
new BatchTransport(
     new OtlpHttpTransport({
          tracesURL: 'www.example.com/v1/traces',
          logsURL: 'www.example.com/v1/logs',
      }),
      { batchSendCount: 100 }
)
```

### Note:
Traces handling will be implemented int the third part as defined in #101 

## Checklist
- [x] Tests added
~- [ ] Changelog updated~
~- [ ] Documentation updated~


packages/web-sdk/src/transports/batch/transport.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (this.signalCount >= this.batchSendCount) {
+      this.flush();
+      return;
+    }
+
+    this.timeoutId = window.setTimeout(() => this.flush(), this.batchSendTimeout);
+  }
+
+  private flush() {
+    this.transport.send(this.signalBuffer);
```
domasx2, line 50: what if `this.signalBuffer` is empty?

packages/web-sdk/src/transports/batch/transport.ts
```diff
@@ -0,0 +1,54 @@
...
+    // If on mobile, it also sends data if user switches from the browser to a different app.
+    document.addEventListener('visibilitychange', () => {
+      if (document.visibilityState === 'hidden') {
+        this.flush();
+      }
+    });
+  }
+
+  send(item: TransportItem): void {
+    clearTimeout(this.timeoutId);
```
domasx2, line 36: so if items keep being added at regular interval < 250s, it might not flush until buffer is full. Maybe it should force flush after max X milliseconds, even if items are being added?

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,120 @@
...
+    const otelPayload = new OtelPayload(undefined, this.internalLogger);
+    const items = isArray(item) ? item : [item];
+
+    items.forEach((item) => otelPayload.addResourceItem(item));
+    this.sendPayload(otelPayload.getPayload());
+  }
+
+  private sendPayload(payload: OtelTransportPayload): void {
+    try {
+      if (this.disabledUntil > new Date(Date.now())) {
```
domasx2, line 44: does current backoff make sense now that there are 3 receiving URLs per signal? maybe it should be separate for each? :thinking:  could be separate backend entirely, eg sending logs to Loki and traces to Tempo

packages/web-sdk/src/transports/batch/transport.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    jest.restoreAllMocks();
+    jest.useRealTimers();
+  });
+
+  it('Sends payload after no new signal has arrived for "batchSendCount" milliseconds after the last one and resets count and payload.', () => {
+    jest.useFakeTimers();
+
+    const mockSendFunction = jest.fn();
+
+    class MockTransport extends BaseTransport {
```
kpelelis, line 33: nit: Since you are reusing this throughout the tests, you can abstract it into a generator function

packages/web-sdk/src/transports/batch/transport.ts
```diff
@@ -0,0 +1,54 @@
...
+    // Send batched/buffered data when user navigates to new page, switches or closes the tab, minimizes or closes the browser.
+    // If on mobile, it also sends data if user switches from the browser to a different app.
+    document.addEventListener('visibilitychange', () => {
+      if (document.visibilityState === 'hidden') {
+        this.flush();
+      }
+    });
+  }
+
+  send(item: TransportItem): void {
```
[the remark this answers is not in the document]
kpelelis: TIL: Typescript will allow this

packages/web-sdk/src/transports/batch/transport.ts
kpelelis: +1. I am used to batching being time framed and not reset with each extra send request

kpelelis (changes requested):
First pass! Very nice work. I like the idea of batching things together to reduce roundtrips. Thank you for all the testing you added for this feature, it helps a lot

packages/web-sdk/src/transports/batch/transport.ts
codecapitano: I this one see it the same as Kostas.

packages/web-sdk/src/transports/batch/transport.ts
codecapitano: Very good catch!

I'll the same config as Otel does. `send_batch_max_size`

From teh docs:
>send_batch_max_size (default = 0): The upper limit of the batch size. 0 means no upper limit of the batch size. This property ensures that larger batches are split into smaller units. It must be greater than or equal to send_batch_size.

packages/web-sdk/src/transports/otlp/transport.ts
```diff
@@ -0,0 +1,120 @@
...
+            body,
+            keepalive: true,
+            ...(restOfRequestOptions ?? {}),
+          })
+            .then((response) => {
+              if (response.status === 429) {
+                this.disabledUntil = this.getRetryAfterDate(response);
+                this.logWarn(`Too many requests, backing off until ${this.disabledUntil}`);
+              }
+
```
domasx2, line 87:
```suggestion
response.text().then(() => {}).catch(() => {});
```
consume the body :D

packages/web-sdk/src/transports/otlp/payload/OtelPayload.ts
```diff
@@ -14,46 +12,56 @@ interface ResourceLogsMetaMap {
...
   }
 
   getPayload(): OtelTransportPayload {
     return {
-      resourceLogs: this.resourceLogsWithMetas.map(({ resourceLog }) => resourceLog),
-      resourceSpans: [],
-      resourceMetrics: [],
+      resourceLogs: this.resourceLogsWithMetas.length
+        ? this.resourceLogsWithMetas.map(({ resourceLog }) => resourceLog)
+        : undefined,
```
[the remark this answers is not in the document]
codecapitano: Nice,let's make it explicit.
Its is also saver to handle by consumers of the paload.

packages/web-sdk/src/transports/otlp/transport.test.ts
```diff
@@ -0,0 +1,238 @@
+import { LogEvent, LogLevel, TransportItem, TransportItemType, VERSION } from '@grafana/faro-core';
```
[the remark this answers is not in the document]
codecapitano: Thank you Kostas very happy to hear that ❤️

packages/web-sdk/src/transports/otlp/transport.ts
codecapitano: You're right it would be better to backoff per endpoint

packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
```diff
@@ -94,7 +93,6 @@ function toLogRecord(transportItem: LogTransportItem): LogRecord {
       return toMeasurementLogRecord(transportItem as TransportItem<MeasurementEvent>);
     default:
       const error = `Unknown TransportItemType: ${type}`;
-      internalLogger.error(error);
```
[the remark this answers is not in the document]
codecapitano: Hi @kpelelis 
I used the internalLogger in wrong way in this module.

**Problem was:**
I instantiated the logger myself in it's own file.
But here I can't access the Faro config which may include a custom logger defined by the user.

So I removed it and started investigating if I:
* Can receive the already configured logger form some kind of context 
* Or if I have to pass it down from the transport because each transport has the configured logger attached by default.

**Solution**
We have no context where I can get the logger from so I will pass it further down the tree starting from the transport.

packages/web-sdk/src/transports/batch/transport.ts
kpelelis: Is this resolved? I see that the timeout is reset in each send request. Or maybe Github is not good with changes :(

packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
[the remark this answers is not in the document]
kpelelis: I am unresolving the thread just to keep track of the changes

kpelelis (changes requested):
This is almost good to go from my side. I unresolved 2 threads with some comments so that we don't lose track of what needs to change

packages/web-sdk/src/transports/batch/transport.ts
codecapitano: Oooh thanks Kostas. 

Yes it is resolved with this change:
```
if (this.batchSendTimeout > 0) {
      window.setTimeout(() => this.flush(), this.batchSendTimeout);
    }
```

But I forgot to remove the left over `clearTimeout` at the top of  `send()` 🙈 
Will clean this up

kpelelis (approved):

Issue #118, opened 2023-02-28
Capturing performance.marks / performance.measure 


NasarTarique:
## Description

Having faro capture the performance.marks and performance.measure and sending it as a trace for certain custom events/ actions would make it easier to measure and track  the user activity . 

## Proposed solution

Add an instrumentation for capturing  performance.mark/measure info 

## Context
[performance.measure](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure)

codecapitano:
Hi @NasarTarique thanks a lot for your participation and ideas.

We already started to work on instrumenting auto capture of Performance, Navigation and Resource timings (#112 ).
We will very likely build the feature in a way that users can choose  to add more `entryTypes` to be observed and captured (by default we will not add many of the, because this may negatively impact client performance).

If you need that kind of functionality immediately you can either implement a custom instrumentation or [add functionality to your app which captures performance entries](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver#examples), such as marks and measurements, and [use Faros push* APIs to send the data  to your endpoint](https://grafana.com/docs/grafana-cloud/faro-web-sdk/custom-signals/).


See also: [https://grafana.com/docs/grafana-cloud/faro-web-sdk/faro-web-sdk-components/provided-instrumentations/](https://grafana.com/docs/grafana-cloud/faro-web-sdk/faro-web-sdk-components/provided-instrumentations/)

codecapitano:
@NasarTarique We just released an **experimental** [instrumentation to capture performance timeline events](https://www.npmjs.com/package/@grafana/faro-instrumentation-performance-timeline).
It offers an option to define which events it should capture. 

The README shows an example how to capture `mark` and `measure`.

PR #121, opened 2023-03-03
Experimental PerformanceTimeline Instrumentation

codecapitano:
## Description

Experiment to instrument [PerformanceTimeline](https://www.w3.org/TR/performance-timeline/)
This enables tracking [resource and navigation timings](https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings#navigation_timing) as well other [Performance Entries](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry).

#### You can add the following to the instrumentation list in the [initialize file of our demo app](https://github.com/grafana/faro-web-sdk/blob/c1e67546a7de6a6165f269cf34adab2bbebb8799/demo/src/client/faro/initialize.ts#L15) and view the event in the dashboard or inspect them in the browser.

```
 new PerformanceTimelineInstrumentation(),
```

#### Default the instrument observes navigation and resource timings, but you can choose what entry types to observe and receive performance timings for. This would btw enable feature request #118.

```
new PerformanceTimelineInstrumentation({
  observeEntryTypes: [
    ...DEFAULT_PERFORMANCE_TIMELINE_ENTRY_TYPES,
    { type: 'event', durationThreshold: 16, buffered: true },

    { type: 'mark',  buffered: true }, // #118
    { type: 'measure', , buffered: true }, // #118
  ],
}),
```

#### Skip Entries
1. It is possible skip entries by URL,
2. by a specific property contained in an performance entry defined by key/value pair (applies to all entries), 
3. by a specific property contained in specific entry types only defined by scope plus key/value pair,

Skipping entries by key/value pairs is useful when you want to remove noise. For example only [tracking human visible navigations](https://github.com/grafana/faro-web-sdk/pull/121). 

```
 new PerformanceTimelineInstrumentation({
        ignoredURLs: [...]
      }),
  ```
  
  ```
 new PerformanceTimelineInstrumentation({
    skipEntries: [
        { key: 'initiatorType', value:  'link'},
        { key: 'serverTiming value: [] },
     ]
   }),
  ```

```
new PerformanceTimelineInstrumentation({
        skipEntries: [
          {
            applyToEntryTypes: ['navigation'],
            skipEntries: [
              { key: 'type', value: 'reload' },
              { key: 'type', value: 'back_forward' },
            ],
          },
        ],
      }),
```

Screenshot 2023-03-03 at 17 24 29

## Current Issues
This instrumentation is in a very early state and not meant for use in production.

Performance Observer may create a lot of events in very short time. So it is advised to use use a BatchTransport.
BatchTransport is currently only supported by the OtelHttpTransport which will be released soonish.
If we use it with the FetchTransport the transport request buffer will be filled quickly and it starts dropping events. 

## Next steps
* Test the hell out of it. Maybe by dogfooding Grafana?
* Make the instrument it's own npm experimental package so we don't deliver it with the default sdk. 
Will also reduce payload for CDN (iife bundle) users to only add it when needed.

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### CODE REVIEW

**Comment by domasx2** [comment]
Super cool :) an option to skip entries by name might make sense

File: packages/web-sdk/src/instrumentations/performance/instrumentation.ts [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+            const isExcluded = skipEntries.some(({ key, value }) => {
+              return convertedEntry[key] === value;
+            });
+
+            if (isExcluded) {
+              continue;
+            }
+          }
+
+          this.api.pushEvent('performance-entry', convertedEntry);
```
codecapitano, line 86: use underline as separator `this.api.pushEvent('performance_entry, ...)`

codecapitano: Solved: it's camel case and Loki changes it to snake case

packages/instrumentation-performance-timeline/README.md
```diff
@@ -0,0 +1,149 @@
...
+#### By URL
+
+It is possible to skip capturing entries by URL. This is can be archived by specifying respective
+URLs in the `ignoredURLs` array.
+By default Faro skips urls defined by the transports. Usually these are the receiver URLs.
+
+##### Example of how to specify skip URLs
+
+```ts
+elineInstrumentation({
```
cedricziel, line 96:
```suggestion
new PerformanceTimelineInstrumentation({
```

cedricziel (approved):

PR #124, opened 2023-03-07
have fetch transport read the response body

domasx2:
## Description

Have fetch transport consume response body. If it's not consumed, request is not considered finished.

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/transports/fetch/transport.ts
+++ b/packages/web-sdk/src/transports/fetch/transport.ts
@@ -1,4 +1,4 @@
-import { BaseTransport, createPromiseBuffer, getTransportBody, PromiseBuffer, VERSION } from '@grafana/faro-core';
+import { BaseTransport, createPromiseBuffer, getTransportBody, noop, PromiseBuffer, VERSION } from '@grafana/faro-core';
 import type { TransportItem } from '@grafana/faro-core';
 
 import type { FetchTransportOptions } from './types';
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -60,9 +60,14 @@ export class FetchTransport extends BaseTransport {
               this.disabledUntil = this.getRetryAfterDate(response);
               this.logWarn(`Too many requests, backing off until ${this.disabledUntil}`);
             }
-
+            // read the body so the connection can be closed
+            response
+              .text()
+              .then(() => {})
```
bfmatei, line 66: Just a nitpick: I think we can use `noop` here

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -60,9 +60,14 @@ export class FetchTransport extends BaseTransport {
...
+              .catch(() => {});
             return response;
           })
+
```
bfmatei, line 70:
```suggestion
```

bfmatei (approved):
Nitpicks

packages/web-sdk/src/transports/fetch/transport.ts
domasx2: removed the `.then()`, I think it's redundant, and added `noop` to `.catch()`

PR #130, opened 2023-03-09
Fix: Only use Resource relevant Metas to decide to create a new ResourceLog or add item to the respective existing one. 

codecapitano:
## Description

Only use Resource relevant Metas to decide to create a new ResourceLog or add item to the respective existing one. 

## Fixes
We accidentally used the whole Meta Object to find out (via diff) if a ScopeLog already exists for the given Meta.
But the default Meta contains much more Metas as used to define the Otel Resource. 
This lead to a case where a new ScopeLog is created because non Otel Resource relevant Metas have changed.

So even if the Resource relevant metas i. e.  MetaBrowser, MetaSDK and MetaApp did not change, a new ResourceLog was created because some of the other Metas changed.

## Checklist

- [x] Tests aligned to cover that use case 
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/transports/otlp/payload/OtelPayload.test.ts
+++ b/packages/web-sdk/src/transports/otlp/payload/OtelPayload.test.ts
@@ -82,6 +82,14 @@ describe('OtelPayload', () => {
 
     otelPayload.addResourceItem({
       ...transportItem,
+      meta: {
+        ...transportItem.meta,
+        // Page meta is NOT used to create the Resource object.
+        // This is to ensure hat we do only diff Resource related metas when defining if item belongs to a Resource we've already created.
+        page: {
+          id: '123',
+        },
+      },
       payload: {
         ...transportItem.payload,
         name: 'event-name-2',
--- a/packages/web-sdk/src/transports/otlp/payload/OtelPayload.ts
+++ b/packages/web-sdk/src/transports/otlp/payload/OtelPayload.ts
@@ -1,12 +1,13 @@
-import { deepEqual, InternalLogger, Meta, TransportItem, TransportItemType } from '@grafana/faro-core';
+import { deepEqual, InternalLogger, TransportItem, TransportItemType } from '@grafana/faro-core';
 
 import { initLogsTransform, LogsTransform } from './transform';
+import type { ResourceMetas } from './transform/types';
 import type { ResourceLog } from './transform';
 import type { OtelTransportPayload } from './types';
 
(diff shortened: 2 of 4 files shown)


packages/web-sdk/src/transports/otlp/payload/OtelPayload.test.ts
```diff
@@ -82,6 +82,14 @@ describe('OtelPayload', () => {
 
     otelPayload.addResourceItem({
       ...transportItem,
+      meta: {
+        ...transportItem.meta,
+        // Page meta is NOT used to create the Resource object.
+        // This is to ensure hat we do only diff Resource related metas when defining if item belongs to a Resource we've already created.
```
leveyjam, line 88: nit: "hat->that", "we only diff" (~do~)
```
This is to ensure that we only diff Resource related metas when defining if item belongs to a Resource we've already created.
```

packages/web-sdk/src/transports/otlp/payload/OtelPayload.ts
```diff
@@ -1,12 +1,13 @@
...
 
 import { initLogsTransform, LogsTransform } from './transform';
 import type { ResourceLog } from './transform';
+import type { ResourceMetas } from './transform/types';
 import type { OtelTransportPayload } from './types';
 
 interface ResourceLogsMetaMap {
   resourceLog: ResourceLog;
-  meta: Meta;
+  resourceMetas: ResourceMetas;
```
leveyjam, line 10: nit-picking / suggestion - remove the plural from the type and property? (resourceMeta: ResourceMeta) (just mentioning because it was singular before!)

packages/web-sdk/src/transports/otlp/payload/transform/types.ts
```diff
@@ -36,3 +36,5 @@ export type LogsTransform = {
   toScopeLog: (transportItem: LogTransportItem) => ScopeLog;
   toLogRecord: (transportItem: LogTransportItem) => LogRecord;
 };
+
+export type ResourceMetas = Pick<Meta, 'app' | 'browser' | 'sdk'>;
```
leveyjam, line 40: Consider renaming "ResourceMeta"?

leveyjam (approved):
Not a TS/JS/ECMAScript expert, but it's understandable / making the comparison match the Resource equality rules 👍 
Minor suggestions about naming / one doc typo/fix.

PR #139, opened 2023-03-21
feat: add traces transform

codecapitano:
## Description

This is the thirds part Part of https://github.com/grafana/faro-web-sdk/pull/101.
It transforms the Faro Trace model to Otlpl compliant model.

Note:
The PR only shows that many files because I renamed a function which is used by them. 

## Checklist
- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
+++ b/packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
@@ -4,6 +4,7 @@ import {
   SemanticResourceAttributes,
   TelemetrySdkLanguageValues,
 } from '@opentelemetry/semantic-conventions';
+import type { TraceEvent } from 'packages/core/src/api';
 
 import {
   EventEvent,
@@ -19,7 +20,17 @@ import type { InternalLogger } from '@grafana/faro-core';
 
 import { isAttribute, toAttribute, toAttributeValue } from '../attribute';
 
-import type { LogRecord, LogsTransform, LogTransportItem, Resource, ResourceMeta, ScopeLog } from './types';
+import type {
+  LogRecord,
+  LogsTransform,
+  LogTransportItem,
+  Resource,
+  ResourceLogs,
+  ResourceMeta,
+  ResourceSpans,
+  ScopeLog,
+  TraceTransform,
+} from './types';
 
 /**
  * Seems currently to be missing in the semantic-conventions npm package.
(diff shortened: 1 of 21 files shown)


packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
```diff
@@ -197,3 +182,45 @@ export function initLogsTransform(internalLogger: InternalLogger): LogsTransform
...
+function toResource(transportItem: TransportItem): Readonly<Resource> {
+  const { browser, sdk, app }: ResourceMeta = transportItem.meta;
+
+  return {
+    attributes: [
+      toAttribute(SemanticBrowserAttributes.BROWSER_MOBILE, browser?.mobile),
+      toAttribute(SemanticBrowserAttributes.BROWSER_USER_AGENT, browser?.userAgent),
+      toAttribute(SemanticBrowserAttributes.BROWSER_LANGUAGE, browser?.language),
+      toAttribute('browser.os', browser?.os),
+      // toAttribute(SemanticBrowserAttributes.BROWSER_BRANDS, browser?.brands),
```
eskirk, line 211: Any reason to leave this in? Future purposes?

packages/web-sdk/src/transports/otlp/payload/transform/types.ts
```diff
@@ -20,21 +15,36 @@ export interface LogRecord {
   traceId?: string;
   spanId?: string;
 }
+export interface ScopeLog {
+  scope: { name: string; version: string };
```
eskirk, line 19: Any interest in making an interface for this? I could go either way - but just wondering if you personally have a minimum number of fields required before making an explicit type.

packages/web-sdk/src/transports/otlp/payload/transform/transform.ts
codecapitano: THnaks Elliot, yes this is left on purpose because we need to decide how to add that data. 
Why?:
Otel defines browser name and version in to be in the same string.
Thosename+versioncombinations are added to the `browser.brands` array.
Loosely such a list represents a group of browsers believed to be compatible with each other.

This information is received from [`navigator.userAgentData.brands` which currently is only suppported in Chromium based browsers](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/brands#browser_compatibility).
For us it would be hard to decide what is compatible engine etc. 

So this will be done in an extra PR to get some more opinions in how to solve it.

Mr current idea is to discuss two things:

1. Detect if browser supports `NavigatorUAData`, then query it directly from that interface.
If not, then add browserName+version string to the brands list. 
This would of course leave out the additional items (compatible browsers) that Browsers which have support for the API do provide (see example below from Chrome). 

2. Simply wait till all browsers have support an keep our current Faro approach (it's how it is implemented atm)

3. A mix of the above. If a browser has support for the API also add browser.brands, if not leave it empty and keep the current model which is `browser.name: string` / `browser.version: string`

packages/web-sdk/src/transports/otlp/payload/transform/types.ts
codecapitano: MUch better. Let me make it interface.

PR #150, opened 2023-04-05
chore: add brands array for browsers supporting `userAgentData.

codecapitano:
## Description

Add the `browser.brands` property as defined by the[ Otel Semantic Attributes spec for browsers](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/browser/).

**Note:**
This information is received from [navigator.userAgentData.brands which currently is only suppported in Chromium based browsers](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/brands#browser_compatibility).

### Current proposal: 
The current proposal adds browser.brands only for those browsers which support the `userAgentData` API.
For others browsers the property will be set to `unknown` as it is for the other properties in `BrowserMeta`

### Alternatives
**1. Always add brands.**
 For browser which do not support `userAgentData` we add `[{brand: name, version: version}]`. Pro: the property always as value added and no unknown states. Con: browsers supporting the flag add a few more items to the list which may lead to inconsistencies in the data we sent.

**2. Never add brands**
Simply wait till all browsers implement `userAgentData`
Pro: no additional work for use. Con: not aligned with Otel spec when using OtlpHttpTransport
 
## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/transports/otlp/payload/transform/toResourceLog.test.ts
+++ b/packages/web-sdk/src/transports/otlp/payload/transform/toResourceLog.test.ts
@@ -19,6 +19,11 @@ const item: Readonly<TransportItem<LogEvent>> = {
       mobile: false,
       userAgent: 'browser-ua-string',
       language: 'browser-language',
+      brands: [
+        { brand: 'Google Chrome', version: '111' },
+        { brand: 'Not(A:Brand', version: '8' },
+        { brand: 'Chromium', version: '111' },
+      ],
     },
     sdk: {
       name: 'integration-web-sdk-name',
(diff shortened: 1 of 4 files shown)


packages/web-sdk/src/transports/otlp/payload/transform/toResourceLog.test.ts
```diff
@@ -19,6 +19,11 @@ const item: Readonly<TransportItem<LogEvent>> = {
       mobile: false,
       userAgent: 'browser-ua-string',
       language: 'browser-language',
+      brands: [
+        { brand: 'Google Chrome', version: '111' },
+        { brand: 'Not(A:Brand', version: '8' },
```
kpelelis, line 24: Is this testcase, just to illustrate the input is not sanitised? By the way, thanks for testing even minor changes like this

kpelelis (approved):
:rocket:

PR #190, opened 2023-05-15
add additional error context to excceptions API

### LINKED ISSUE #141: Allow extra context for errors
## Description

Currently `faro.api.pushError()` only accepts and `Error` argument, there is no way to add extra context data.
One significant use case would be capturing React error boundary errors, where in addition to the regular stack trace
there is the component stack trace. It would be great to be able to add is as extra information. 

## Proposed solution

Add optional context parameter, same as for `faro.api.pushLog()`. Will require changes for recevier too

codecapitano:
## Description

Add option to put extra context to exception logs.

I kept the API similar to Logs API where we add additional context via the options object.

[Cloud docs PR to show how to add additional error context](https://github.com/grafana/cloud-docs/pull/922).

## Fixes

Fixes #
Fixes #141 

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Next
+
+- Add: add option to provide additional error context in `pushError` api.
+
 ## 1.0.3 – 1.0.5
 
 - Add: Config option to include URLs for which requests shall have tracing headers added.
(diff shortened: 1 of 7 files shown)


domasx2 (approved):
LGTM!

Does receiver support it this property?
Error boundary in react package currently replaces `error.stack` with component stack. Could now be updated to keep the original js stack and add component stack to context

codecapitano:
> react

Thanks for the review Domas ❤️, 

* I totally forgot about the receiver. Will double check and implement the missing functionality.
* I'll update the React instrumentation ones I ensured that the receiver properly handles the context prop.

PR #193, opened 2023-05-16
fix: narrow LogContext and ExceptionContext types.

codecapitano:
## Description

**Why**
The Faro Receiver defines the type for `LogContext` and `ErrorContext` as a string to map.
Whereas the Web-SDK defines it as type `BaseObject` which is nearly `any` type.

**What**
Narrow `LogContext` and `ErrorContext` down to string to string map.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/core/src/api/exceptions/types.ts
+++ b/packages/core/src/api/exceptions/types.ts
@@ -1,4 +1,3 @@
-import type { BaseObject } from '../../utils';
 import type { TraceContext } from '../traces';
 
 export type StacktraceParser = (err: ExtendedError) => Stacktrace;
@@ -20,7 +19,9 @@ export interface Stacktrace {
   frames: ExceptionStackFrame[];
 }
 
-export type ExceptionContext = BaseObject;
+export type ExceptionContext = {
+  [key: string]: string;
+};
 
 export interface ExceptionEvent {
   timestamp: string;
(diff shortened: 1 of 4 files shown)


packages/core/src/api/exceptions/types.ts
```diff
@@ -20,7 +19,9 @@ export interface Stacktrace {
   frames: ExceptionStackFrame[];
 }
 
-export type ExceptionContext = BaseObject;
+export type ExceptionContext = {
```
domasx2, line 22: most idiomatic TS way would be `export type ExceptionContext = Record<string, string>`

domasx2 (approved):

packages/core/src/api/exceptions/types.ts
codecapitano: Oh yes of course thank

packages/core/src/api/exceptions/types.ts
```diff
@@ -20,7 +19,7 @@ export interface Stacktrace {
   frames: ExceptionStackFrame[];
 }
 
-export type ExceptionContext = BaseObject;
+export type ExceptionContext = Record<string, string>;
```
JoniJnm, line 22: why the value type is a string if `pushError` and `pushLog` apply a `stringifyObjectValues(context)`?

https://github.com/grafana/faro-web-sdk/issues/1998

PR #195, opened 2023-05-17
Core: Use `globalThis` rather than `window`

svennergr:
## Description

This PR changes the `GlobalObject` to be computed from `globalThis`, `global` and `self`, rather than sticking to `global` if `window` is not defined. 

Background: I am currently building a browser extension, which uses webworkers to run background tasks. Those webworkers don't have a `global` nor a `window` object. The current implementation in core fails and leads to errors. By using `globalThis` the issue can be mitigated, as `globalThis` is present in Node, where it points to `global`, in a usual browser window, where it points to `window` and in webworkers, where it points to `self` (I think).

I just went with this check I implemented because that is also what was used in Typescript:
https://github.com/microsoft/TypeScript/blob/v4.6.4/src/compiler/corePublic.ts#L119-L122

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next
 
 - Add: add option to provide additional error context in `pushError` api.
+- Fix: Use `globalThis` instead of `global` or `window` in case the SDK is used in webworkers.
 
 ## 1.0.3 – 1.0.5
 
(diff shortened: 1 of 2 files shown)


codecapitano (approved):
Awesome, thanks so much for your contribution, LGTM 🎉

PR #200, opened 2023-05-23
Fetch instrumentation experimental package

eskirk:
## Description

This PR will add the experimental package responsible for instrumenting the `fetch` API. 

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


docker-compose.yaml
```diff
@@ -59,7 +59,7 @@ services:
       - '${AGENT_CONFIG_PATH_LOCAL}:${AGENT_CONFIG_PATH}'
       - 'demo_demo_logs:${AGENT_LOGS_PATH}'
     entrypoint:
-      - '/bin/grafana-agent'
+      - '/bin/agent'
```
codecapitano, line 62: Q: A few weeks ago we merged a PR because in upstream symbolic links are now prefixed with `grafana-*`.
Has this chnaged again? 

* [PR: docker-compose: fix grafana-agent path #173 ](https://github.com/grafana/faro-web-sdk/pull/173)
* [Docs/Upgrade guide v0.33](https://grafana.com/docs/agent/latest/static/upgrade-guide/#v033)

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  constructor() {
+    super();
+  }
+
+  /**
+   * Instrument fetch with Faro
+   */
+  private _instrumentFetch(): typeof fetch {
```
codecapitano, line 21: Q: `typeof fetch` will return `function`. 
Not sure if this is what you want or if you want the return type of `fetch()` as constructed by using teh TS utility type   `ReturnType<Type>`.

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
+import { BaseInstrumentation, faro, globalObject } from '@grafana/faro-core';
+
+import { fetchGlobalObjectKey, originalFetch, originalFetchGlobalObjectKey } from './constants';
+
+export class FetchInstrumentation extends BaseInstrumentation {
+  readonly name = '@grafana/faro-web-sdk:instrumentation-fetch';
+  readonly version = '1.0.0'; // TODO - pull from package.json
+
+  initialize(): void {
+    console.log('Initializing FetchInstrumentation');
```
codecapitano, line 10: Hint: 
I think the console.log is there for dev purposes.
if you want to log data later (debug, infor, warn, error) you can do this with Faros' internalLogger utility which is provided via the BaseInstrumentation.

```ts
this.internalLogger.info('foo')
```

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  constructor() {
+    super();
+  }
+
+  /**
+   * Instrument fetch with Faro
+   */
+  private _instrumentFetch(): typeof fetch {
+    return function instrumentedFetch(input, init): Promise<Response> {
```
codecapitano, line 22: Add [types as defined by the fetch spec](https://fetch.spec.whatwg.org/#fetch-method): 

```ts
instrumentedFetch(input: RequestInfo, init: RequestInit = {})
```

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+      const request = new Request(input, copyInit);
+      if (body) {
+        copyInit.body = body;
+      }
+
+      return originalFetch(input instanceof Request ? request : input, copyInit).then(function (response: Response) {
+        const simplifiedHeaders = {} as Record<string, string>;
+        new Headers(response.headers).forEach((v, k) => (simplifiedHeaders[`response_header_${k}`] = v));
+
+        // TODO - add instrumented functions to capture onError and onSuccess
```
codecapitano, line 40: Q: Certain data like url etc. is not tracked. I think you'll add it later but want to give a hint anyways :)

codecapitano: Suggestion/idea: Consider sending object with respective props like e. g.

```ts
{
    timesTamp: ...
    urls: ...
    headers: ...
    ...
}

File: experimental/instrumentation-fetch/src/instrumentation.ts [hunk 6]
```diff
@@ -0,0 +1,84 @@
...
+      enumerable: false,
+      writable: false,
+      value: originalFetch,
+    });
+  }
+
+  /**
+   * implements disable function
+   */
+  disable(): void {
```
codecapitano, line 69: Q: why do we have a enable/didable function? In what cases do we call `disable()` and who's calling it?
Asking because the other instruments don't have that functionality.
Hint: `pause()` is handled at the `transports` level, no additional wotk for instrumentation developers :)

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,84 @@
...
+  private _instrumentFetch(): typeof fetch {
+    return function instrumentedFetch(input, init): Promise<Response> {
+      let copyInit = init ? Object.assign({}, init) : {};
+
+      let body;
+      if (copyInit && copyInit.body) {
+        body = copyInit.body;
+        copyInit.body = undefined;
+      }
+
```
codecapitano, line 30: Suggestion: don't forget to add the `skipUrls` functionality. 
It's also important to prevent Faro from reporting on itself.

docker-compose.yaml
eskirk: Yes, this is new. I think I had something misconfigured on my end and this fixed things. Will revert this change.

experimental/instrumentation-fetch/src/instrumentation.ts
eskirk: Yes this console log was here for debugging purposes, will switch it out in favor of the internal logger.

experimental/instrumentation-fetch/src/instrumentation.ts
eskirk: I love the hints thank you!

experimental/instrumentation-fetch/src/instrumentation.ts
eskirk: Good points. I have been following the OTel fetch instrumentation and they have the enable/disable functionality. I thought it was potentially useful but hadn't really considered the use cases. Knowing that we have pause should be sufficient :+1:

experimental/instrumentation-fetch/src/instrumentation.ts
eskirk: Good point - this would require changes to the receiver? Or does this just mean generating some kind of unique request ID that we can then use in the backend? Depending on the scope I could take care of that with this PR.

experimental/instrumentation-fetch/src/instrumentation.ts
eskirk: I opted to return an entirely new function and overwrite the `fetch` property on the global object with this new function. Do you see any issues that might cause? I felt like it was a clean way to do it, but open to other options.

experimental/instrumentation-fetch/src/instrumentation.ts
eskirk: Also, just to clarify - `timestamp` seems to be one of the fields handled in the `core` package. So providing timestamp twice doesn't seem necessary?

codecapitano: > 

Sorry my bad.Yes you're absolutely right. The core package takes care of that.

experimental/instrumentation-fetch/src/instrumentation.ts
codecapitano: > Good point - this would require changes to the receiver? Or does this just mean generating some kind of unique request ID that we can then use in the backend? Depending on the scope I could take care of that with this PR.

I don't know yet but I expect we need changes on the receivers side. So imho it would make sense to tackle this as an own task because it may be a bigger thing. Also not sure if we need this for GA or not.

demo/src/client/faro/initialize.ts
```diff
@@ -19,6 +20,9 @@ export function initializeFaro(): Faro {
       ...getWebInstrumentations({
         captureConsole: true,
       }),
+      new FetchInstrumentation({
+        ignoredUrls: [`http://localhost:${env.faro.portAppReceiver}/collect`],
```
codecapitano, line 24: The default transport (Fetch) which is added by Faro [automatically add the collector URL defined above to the transport](https://github.com/grafana/faro-web-sdk/blob/f799029fad306b2a16c1a29cef068576f079ab23/packages/web-sdk/src/config/makeCoreConfig.ts#L31-L32).

So you can get the default collector url in the instrument like this:

```ts
  private getIgnoreUrls(): Array<string | RegExp> {
    return this.transports.transports?.flatMap((transport) => transport.getIgnoreUrls());
  }
  ```

Comment (eskirk) [reply demo/src/client/faro/initialize.ts:24]: Very very clean, thank you @codecapitano

File: experimental/instrumentation-fetch/src/instrumentation.ts [hunk 9]
```diff
@@ -0,0 +1,162 @@
...
+  getIgnoredUrls(): Array<string | RegExp> {
+    return this.ignoredUrls ?? [];
+  }
+
+  /**
+   * Instrument fetch with Faro
+   */
+  private instrumentFetch(): typeof fetch {
+    const instrumentation = this;
+    return function fetch(this: typeof globalObject, input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
```
codecapitano, line 74: Suggestion: 
Instead of passing `this` as a parameter you can either use `bind` or an arrow function when you create the respective fetch object  (`value:...`).

for example somethig like:
```ts
{
  value: () => this.instrumentFetch(),
}
```

or 

```ts
{
  value: () => this.instrumentFetch(),
}
```

codecapitano (commented):
Cool stuff Ellliot! Mostly some question or suggestions left.

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.ignoredUrls = this.options?.ignoredUrls ?? this.getTransportIgnoreUrls();
+
+    Object.defineProperty(globalObject, fetchGlobalObjectKey, {
+      configurable: true,
+      enumerable: false,
+      writable: false,
+      value: this.instrumentFetch(),
+    });
+
+    Object.defineProperty(globalObject, originalFetchGlobalObjectKey, {
```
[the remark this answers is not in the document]
eskirk: Just for more flexibility - perhaps the user wanted to use the un-instrumented fetch function for some reason?

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+      enumerable: false,
+      writable: false,
+      value: originalFetch,
+    });
+  }
+
+  /**
+   * Generate a unique request id object for each fetch request
+   */
+  private requestId(idOnly?: boolean): Record<string, string> | string {
```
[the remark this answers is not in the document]
eskirk: I only did this for style reason... I don't love having to index objects using properties with underscores - but simplicity is better so I will implement this change 👍🏻

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+          faro.api.pushEvent(
+            resolvedFetchEventName,
+            {
+              ...trimmedResponse,
+              ...parsedHeaders,
+              ...(instrumentation.requestId() as Record<string, string>),
+            },
+            eventDomain,
+            {
+              skipDedupe: true,
```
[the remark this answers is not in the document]
eskirk: Not sure why i initially set that... removing

experimental/instrumentation-fetch/src/instrumentation.ts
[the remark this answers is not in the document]
codecapitano: You mean like using the original fetch for some request the instrumented one for others?

Do have a specific use case in mind? 

Maybe we can provide an option to disable/pause the fetch instrumentation. 
Then we would not add more things to he global object?

PR #204, opened 2023-06-01
Make single TransportItem transports backwards compatible

kpelelis:
## Description

This pull request ensures that the implementations of the Transport interface that were accepting a single item in their `send` function, will work as expected. This type of Transports is getting deprecated instead and will be removed with v2

## Checklist

- [X] Tests added
- [ ] Changelog updated
- [X] Documentation updated

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -296,6 +296,34 @@ The transports also support batch processing wiht the `batchEnabled` flag set to
 and a positive, non zero `batchSendTimeout`. The batch executor will group items that share the same `metas` together
 and call the `send` function for each group of signals.
 
+### Updating transports to support multiple items
+
+Currently the Transport interface supports implementations of the `send` function that accept either a `TransportItem`
+or an array of them. The former is deprecated and will be removed as of version 2. To update your transport:
+
+```ts
+export class MyTransport extends BaseTransport {
+  async send(item: TransportItem): Promise<void> {
+    // do something with the item
+  }
+}
+
+/* should be */
+
+export class MyTransport extends BaseTransport {
+  async send(items: TransportItem[]): Promise<void> {
+    // do something with the array of items
+    // all of them share the same metas
+  }
+
+  // This is required for the transports layer to differentiate
+  // between the two types of Transport implementations
+  override isBatched(): boolean {
+    return true;
+  }
+}
+```
+
 ## Unpatched console
 
 Some instrumentations might override the default console methods but Faro instance provides a way to access the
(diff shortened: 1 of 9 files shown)


packages/core/README.md
```diff
@@ -296,6 +296,34 @@ The transports also support batch processing wiht the `batchEnabled` flag set to
 and a positive, non zero `batchSendTimeout`. The batch executor will group items that share the same `metas` together
 and call the `send` function for each group of signals.
 
+### Updating transports to support multiple items
+
+Currently the Transport interface supports implementations of the `send` function that accept eiter a `TransportItem`
```
codecapitano, line 301: Littel typo

```suggestion
Currently the Transport interface supports implementations of the `send` function that accept either a `TransportItem`
```

packages/core/README.md
```diff
@@ -296,6 +296,34 @@ The transports also support batch processing wiht the `batchEnabled` flag set to
...
+or an array of them. The former is deprecated and will be removed as of version 2. To update your transport:
+
+```lang=typescript
```
codecapitano, line 304: No need for a change: you can also say: `ts` instead of `lang=typescript`

packages/core/src/transports/initialize.ts
```diff
@@ -128,6 +128,7 @@ export function initializeTransports(
       } else {
         send([item]);
       }
+      transports.filter((t) => !t.isBatched()).forEach((t) => t.send(item));
```
codecapitano, line 131: Q: I don't really get the purpose of this `send()` chain.

Can you help to clear the dust 🙈

So:
1. If bacthing is enabled we add an item to the executor: `batchExecutor.addItem(item);`
2. Else we send a single item array
3. Finally, indipendent of 1 and 2 we get all non-batch transport and call `send()`

3.1 we loop over all batched transports, which means nothing get send because in 3. we filter all non-batched, then call send and filter for batched items.

codecapitano (commented):
Good work, just a few typos and a question about the send() logic.

packages/core/src/transports/base.ts
```diff
@@ -5,6 +5,13 @@ import type { Transport, TransportItem } from './types';
 export abstract class BaseTransport extends BaseExtension implements Transport {
   abstract send(items: TransportItem[]): void | Promise<void>;
 
+  isBatched(): boolean {
+    this.internalLogger.warn(
+      `Custom transports supporting a single TransportItem are deprecated. You should change your transport to accomodate an array of TransportItem. You can read how to upgrade your transports in the README.`
```
codecapitano, line 10: here was anotehr typo in my suggestion I think 🙈

```suggestion
      `Custom transports supporting a single TransportItem are deprecated. You should change your transport to accommodate an array of TransportItem. You can read how to upgrade your transports in the README.`
```

codecapitano (approved):
LGTM 🦖

PR #208, opened 2023-06-07
Clean up batch processing

kpelelis:
## Description

After some suggestions by @bfmatei, there might be cases of transports that require instantaneous execution of the signal (such as a WebSocket, a Console or any stream). To that end, this commit removes the deprecation notice of the un-batched transports and instead changes the core logic which is responsible for sending the item.
## Checklist

- [ ] Tests added
- [X] Changelog updated
- [X] Documentation updated

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -296,10 +296,11 @@ The transports also support batch processing wiht the `batchEnabled` flag set to
 and a positive, non zero `batchSendTimeout`. The batch executor will group items that share the same `metas` together
 and call the `send` function for each group of signals.
 
-### Updating transports to support multiple items
+### Batched transports
 
 Currently the Transport interface supports implementations of the `send` function that accept either a `TransportItem`
-or an array of them. The former is deprecated and will be removed as of version 2. To update your transport:
+or an array of them. The latter is useful when you want to batch process signals with your custom transport. In order
+to achieve this, you should override the `isBatched` method of the BaseTransport.
 
 ```ts
 export class MyTransport extends BaseTransport {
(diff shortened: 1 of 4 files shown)


bfmatei (commented):
Looks good from a tech perspective. @codecapitano should be the one approving it.

packages/core/README.md
```diff
@@ -296,10 +296,11 @@ The transports also support batch processing wiht the `batchEnabled` flag set to
 and a positive, non zero `batchSendTimeout`. The batch executor will group items that share the same `metas` together
 and call the `send` function for each group of signals.
 
-### Updating transports to support multiple items
+### Batched transports
 
 Currently the Transport interface supports implementations of the `send` function that accept either a `TransportItem`
-or an array of them. The former is deprecated and will be removed as of version 2. To update your transport:
+or an array of them. The later is useful when you want to batch process signals with your custom transport. In order
```
codecapitano, line 302: Little typo

```suggestion
or an array of them. The latter is useful when you want to batch process signals with your custom transport. In order
```

codecapitano (approved):
Thnaks Kostas, LGTM 🦖

packages/core/src/transports/initialize.ts
```diff
@@ -102,37 +102,67 @@ export function initializeTransports(
     return filteredItems;
   };
 
-  const send = (items: TransportItem[]) => {
+  const batchedSend = (items: TransportItem[]) => {
```
codecapitano, line 105: Suggestion:
Maybe we can rename this to better reflect that thsi send method if only for the BatchExecutor?
E. g. `batchExecuterSend`

PR #214, opened 2023-06-16
Add browserslist to faro packages

cedricziel:
## Description

In order to indicate to downstream consumers the compile targets to allow polyfilling, this change adds the browserslist entries to each package.json of the provided packages in the repo.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-tracing/package.json
+++ b/packages/web-tracing/package.json
@@ -25,6 +25,7 @@
   "module": "./dist/esm/index.js",
   "types": "./dist/types/index.d.ts",
   "files": [
+    ".browserslistrc",
     "dist",
     "README.md",
     "LICENSE"
@@ -72,7 +73,5 @@
   },
   "publishConfig": {
     "access": "public"
-  },
-  "gitHead": "418b7f9ae1d5e6397404d4d6a1938ee557b8575c",
-  "browserslist": [ "supports es6-module" ]
+  }
 }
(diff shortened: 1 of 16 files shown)


packages/web-tracing/package.json
```diff
@@ -73,5 +73,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "gitHead": "418b7f9ae1d5e6397404d4d6a1938ee557b8575c"
+  "gitHead": "418b7f9ae1d5e6397404d4d6a1938ee557b8575c",
```
cedricziel, line 76: Do we need this?

cedricziel:
> I am hesitating a bit adding a browserlistrc to each package separately.
My understanding is that we need it in every directory so it gets distributed with the packages. no?

> Browserlist list will traverse the directory tree till it finds a browserlistrc.
Up or down?

codecapitano:
> > I am hesitating a bit adding a browserlistrc to each package separately.
> 
> My understanding is that we need it in every directory so it gets distributed with the packages. no?
> 
> > Browserlist list will traverse the directory tree till it finds a browserlistrc.
> 
> Up or down?

It looks in each dir 🙈 We can speed this up by setting an env variable pointing to the root config.
Asked in Slack about best practices here. Let's wait for some input :)

codecapitano:
> 

Clarified. We're good to go :)

codecapitano (approved):
LGTM 🦖 One little thing. Seems like browserlist is not included to the fetch-instrumentation package.

PR #215, opened 2023-06-20
Remove OTel user-interaction instrumentation from the default set

cedricziel:
## Description

This instrumentation is very chatty and should only be enabled when needed. It is not needed for the default set of instrumentations.

## Fixes

Fixes #

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next
 
+- Remove: Remove OTel user-interaction instrumentation from the default set (#215)
+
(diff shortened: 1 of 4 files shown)


codecapitano:
@cedricziel please update the changelog so that users are aware that we no more track `user-interaction` out of the box with the tracing package.

domasx2 (approved):

codecapitano (approved):

PR #217, opened 2023-06-20
Fix yarn version

domasx2:
:wave:  Lerna doesn't work well with yarn 2+. Fix version to yarn classic in the meanwhile


eskirk (approved):
Just making sure I understand - all this does is add a pre-bundled yarn module to our repo that will act as the executor for all our yarn commands? 

If so, perfect 👍🏻 remove any inconsistencies with publishing/building/etc. Thanks @domasx2

domasx2:
@eskirk correct :)

PR #226, opened 2023-07-03
fix: removing UserInteractionInstrumentation from quickstart

AMontagu:
UserInteractionInstrumentation is not part of the faro default instrumentation anymore.


codecapitano (approved):
Thanks a lot @AMontagu. LGTM 🦖

PR #253, opened 2023-08-21
252 oltphttptransport resulting in 400 response status

codecapitano:
## Why
The `faro-transport-otlp-http` used the wrong OTLP structure when sending the respective signals
The problem happened in the `OtlpHttpTransport.class` when iterating over the `OtelTransportPayload` object.

The `OtelTransportPayload` object has the following structure:
```ts
export interface OtelTransportPayload {
  readonly resourceLogs: Readonly<ResourceLogs>;
  readonly resourceSpans: Readonly<ResourceSpans>;
}
```

In the `sendPayload` method in `OtlpHttpTransport` we iterate over the object and do some signal specific configuration before sending the signal.

The respective loop looks like this:
```ts
   for (const [key, value] of Object.entries(payload)) { ... }
```

When stringifying the signal payload to attach it to the requests body we did:
```ts  
const body = JSON.stringify(value); // this removed the outer part 
```
Which removed the outer part so we only sent an array instead of an object.

The solution is to (re)-build the correct otel payload like:
```ts
const body = JSON.stringify({ [key]: value }); // ---> const body = JSON.stringify({ resourceLogs: ResourceLogs })
```

## What
* (Re-) Build the correct otel payload before sending it (`const body = JSON.stringify({ [key]: value });`)
* Add new types for `ResourceLogs` and `ResourceSpans` for better readability
* Remove the left over `resolve` object from `vite.config` in demo 

### Relevant OTLP Specs
* [logs.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/81a296f9dba23e32d77f46d58c8ea4244a2157a6/opentelemetry/proto/logs/v1/logs.proto#L44)
* [logs.example](https://github.com/open-telemetry/opentelemetry-proto/blob/main/examples/logs.json)
* [trace.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/81a296f9dba23e32d77f46d58c8ea4244a2157a6/opentelemetry/proto/trace/v1/trace.proto#L44)
* [trace.example](https://github.com/open-telemetry/opentelemetry-proto/blob/main/examples/trace.json)

### Now
Screenshot 2023-08-21 at 15 11 04
Screenshot 2023-08-21 at 15 10 24

### Before
Screenshot 2023-08-21 at 15 09 16
Screenshot 2023-08-21 at 15 09 03

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/demo/vite.config.ts
+++ b/demo/vite.config.ts
@@ -6,9 +6,6 @@ import { getEnvConfig, getPublicEnvConfig } from './src/common';
 export default defineConfig(({ mode }) => {
   return {
     plugins: [react()],
-    resolve: {
-      preserveSymlinks: true, // TODO - remove once fetch instrumentation is published to NPM
-    },
     server: {
       watch: {
         awaitWriteFinish: true,
--- a/experimental/CHANGELOG.md
+++ b/experimental/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+### FaroOtlpHttpTransport
+
+1 [fix]: The outer shape of the payload was not correct which lead to 400 response status codes ([#252](https://github.com/grafana/faro-web-sdk/issues/252)).
+
 ## 1.1.3
 
 ### FetchInstrumentation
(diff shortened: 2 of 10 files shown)


LittleLadySunshine (approved):
@codecapitano which envs is this PR going out to? I ask because of lines 34-36. If we're testing let's make sure this just goes to preprod envs only.

codecapitano:
> @codecapitano which envs is this PR going out to? I ask because of lines 34-36. If we're testing let's make sure this just goes to preprod envs only.

Thank you Denise. Which file are you referring to with lines 34–36?

The way we want users to install and update the Web-SDK is via npm / unpk (the CDN version of npm).
So merging has no effect on environments. Auto-updates will usually trigger once we released a new version to npm.

PR #263, opened 2023-09-01
Add specific meta data if Faro is running in K6 browser

codecapitano:
## Why

To enable new capabilities to include K6 data into Frontend O11Y we must be able to distinguish between lab and field data.

ToDo:
- [ ] Validate solution by using the K6 browser

## What
* Add a meta object (`K6Meta`) in case Faro is running in the K6 browser. This is determined by the existence of a k6 object on the global object.
* We also ensure that the Faro default metas are always added. 

## Links
* [k6 browser PR](https://github.com/grafana/xk6-browser/pull/1007)

## Screenshots
Note: 
The following screenshot is not taken from within the K6 browser.
I added the k6 object via code in our demo env just for the sake of taking the screenshot and showing the data

Screenshot 2023-09-01 at 17 56 09

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


packages/web-sdk/src/metas/k6/meta.ts
```diff
@@ -0,0 +1,11 @@
+import { isObject } from '@grafana/faro-core';
+import type { Meta, MetaItem } from '@grafana/faro-core';
+
+export const k6Meta: MetaItem<Pick<Meta, 'k6'>> = () => {
+  const isK6Browser = isObject((window as any).k6);
```
cedricziel, line 5: Does this need a condition? I.e: only add this meta if true?

codecapitano: Not really because we only the meta if Faro is running in k5 browser.
It's redundant, will remove it.

eskirk:
Reviewed once, will review again tomorrow but looks good

PR #265, opened 2023-09-07
New user  session handling

codecapitano:
## Why
We want to provide an option for users to decide if they want to use persistent session tracking which can persist a session between page navigations (static websites) and synchronize a session across browsing contexts (tabs and windows). The new session manager invalidates or better creates new session if a session is invalid. 

Users can also pick in-memory session management which aligns to the concepts of the above as far as possible.
But it has limitations, especially om static websites because all state is hold in-memory.
We will enhance this situation in the follow-up PRs.

This is the first PR to present, test and refine the concept. We will improve things in the next weeks.

## What
* Add a throttle function — used an own implementation to keep the additional payload small
* Add a service to wrap the local storage APIs
* Add the session manager module which contains utility functions for the session managers as well as both session managers. I didn't use a class to save on bundle size because classes are not minimized well by nearly all common minifiers. 

## Note
This PR is a (bigger) PoC and will likely be heavily refactored.

Already known refactorings 
* we'll change the in-memory session manger to use Session Storage. This is to make it easier to make session data outlive page navigations  

See: https://github.com/grafana/faro-web-sdk/issues/355

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/session/sessionHandler.ts
+++ b/packages/web-sdk/src/instrumentations/session/sessionHandler.ts
@@ -9,6 +9,10 @@ export interface FaroUserSession {
   started: number;
   sessionMeta?: MetaSession;
 }
+export interface UserSessionUpdater {
+  handleUpdate: () => void;
+  init: () => void;
+}
 
 // TODO: make this configurable from the outside
 export const SESSION_EXPIRATION_TIME = 4 * 60 * 60 * 1000; // n hrs
@@ -61,12 +65,7 @@ export function isUserSessionValid(session: FaroUserSession | null): boolean {
   return inactivityPeriodValid;
 }
 
-interface PersistentUserSessionUpdater {
-  handleUpdate: () => void;
-  init: () => void;
-}
-
-export function persistentUserSessionsUpdater(initialSessionId?: string): PersistentUserSessionUpdater {
+export function persistentUserSessionsUpdater(initialSessionId?: string): UserSessionUpdater {
   const throttledSessionUpdate = throttle(() => {
     const sessionFromLocalStorage = fetchUserSession();
 
@@ -147,12 +146,7 @@ export function persistentUserSessionsUpdater(initialSessionId?: string): Persis
   };
 }
 
-interface InMemoryUserSessionsUpdater {
-  handleUpdate: () => void;
-  init: () => void;
-}
-
-export function inMemoryUserSessionsUpdater(initialSessionId?: string): InMemoryUserSessionsUpdater {
+export function inMemoryUserSessionsUpdater(initialSessionId?: string): UserSessionUpdater {
   let inMemoryUserSession: FaroUserSession = createUserSessionObject(initialSessionId);
 
   function handleUpdate() {
@@ -208,8 +202,7 @@ export function inMemoryUserSessionsUpdater(initialSessionId?: string): InMemory
   };
 }
 
-export type SessionUpdater = PersistentUserSessionUpdater | InMemoryUserSessionsUpdater;
-export function getSessionUpdater(initialSessionId?: string): SessionUpdater {
+export function getSessionUpdater(initialSessionId?: string): UserSessionUpdater {
   if (faro.config.experimentalSessions?.persistent && isLocalStorageAvailable) {
     return persistentUserSessionsUpdater(initialSessionId);
   }
--- a/packages/web-sdk/src/instrumentations/session/instrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/session/instrumentation.ts
@@ -1,6 +1,6 @@
 import { BaseInstrumentation, Conventions, Meta, MetaSession, VERSION } from '@grafana/faro-core';
 
-import { getSessionUpdater, SessionUpdater } from './sessionHandler';
+import { getSessionUpdater, UserSessionUpdater } from './sessionHandler';
 
 export class SessionInstrumentation extends BaseInstrumentation {
   readonly name = '@grafana/faro-web-sdk:instrumentation-session';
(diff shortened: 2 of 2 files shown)


pawan-regoti:
in which version is this planned to be released?

docs/sources/developer/architecture/components/internal-logger.md
```diff
@@ -4,7 +4,7 @@ The internal logger is a helper utility that is used to log messages from the li
 relies on the unpatched console, the messages printed by the internal logger are not captured during
```
kpelelis, line 4: Memo: Add an entry to the CHANGELOG for the packages affected

packages/web-sdk/src/instrumentations/session/sessionManager.ts
```diff
@@ -0,0 +1,210 @@
...
+
+export function removeUserSession() {
+  removeItem(STORAGE_KEY);
+}
+
+export function storeUserSession(session: FaroUserSession): void {
+  setItem(STORAGE_KEY, JSON.stringify(session));
+}
+
+export function receiveUserSession(): FaroUserSession | null {
```
kpelelis, line 38: nit: A better naming would be `fetchUserSession`

packages/web-sdk/src/instrumentations/session/sessionManager.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+export function isUserSessionValid(session: FaroUserSession | null): boolean {
+  if (session == null) {
+    return false;
+  }
+
+  const now = dateNow();
+  const lifetimeValid = now - session.started < SESSION_EXPIRATION_TIME;
+  const inactivityPeriodValid = now - session.lastActivity < SESSION_INACTIVITY_TIME;
```
kpelelis, line 55: nit: You can short circuit this by first checking the lifetimeValid result, quick returning if the session is expired. And then checking the inactivity

packages/core/src/config/types.ts
```diff
@@ -22,7 +22,17 @@ export interface Config<P = APIEvent> {
 
   beforeSend?: BeforeSendHook<P>;
   ignoreErrors?: Patterns;
+  // TODO: Deprecate?
   session?: MetaSession;
+  // TODO: rename once feature is ready. Do not use in any prod envs!
+  experimental_sessions_enabled?: boolean;
+  // TODO: rename to "sessions" once feature is ready. Do not use in any prod envs!
+  experimental_sessions?: {
```
kpelelis, line 30: rename to `experimentalSessions`

packages/web-sdk/src/config/makeCoreConfig.ts
```diff
@@ -58,8 +59,29 @@ export function makeCoreConfig(browserConfig: BrowserConfig): Config | undefined
...
+
     user: browserConfig.user,
     view: browserConfig.view ?? defaultViewMeta,
   };
 }
+
+function createSessionMeta(): MetaSession {
+  const userSession = receiveUserSession();
+  const isSessionValid = isUserSessionValid(userSession);
+  const sessionId = isSessionValid ? userSession?.sessionId : createSession().id;
```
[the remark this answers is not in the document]
kpelelis: The `isSessionValid` is only being used once

kpelelis (changes requested):
Hey thanks for pulling that off. I added some comments here and there. Happy to discuss these further

packages/web-sdk/src/instrumentations/session/sessionManager.ts
```diff
@@ -0,0 +1,210 @@
...
+
+      faro.api?.setSession(newSession.sessionMeta);
+      faro.config.experimental_sessions?.onSessionSwitch?.(
+        sessionFromLocalStorage?.sessionMeta ?? null,
+        newSession.sessionMeta!
+      );
+    }
+  }, STORAGE_UPDATE_DELAY);
+
+  (function init() {
```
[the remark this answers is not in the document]
codecapitano: > Maybe the whole persistent session manager can be moved into a class and this can be the init function of it, the code above can be the store method and you could be also providing a fetch method

I don't use classes on purpose, because minifiers don't handle them well so we add unnecessary big payload.
But we can move into its own js module.

packages/web-sdk/src/instrumentations/session/sessionHandler.ts
```diff
@@ -0,0 +1,218 @@
...
+    });
+  }
+
+  return {
+    handleUpdate: throttledSessionUpdate,
+    init,
+  };
+}
+
+interface InMemoryUserSessionsUpdater {
```
kpelelis, line 150: This interface is the same as the PeristentSessionUpdater right? Maybe merge them into one?

kpelelis (approved):
Very nice work! One minor comment regarding the Manager interface

packages/web-sdk/src/utils/throttle.test.ts
```diff
@@ -0,0 +1,64 @@
+import { throttle } from './throttle';
```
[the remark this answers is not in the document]
codecapitano: Thanks for commenting on the missing test!

packages/web-sdk/src/instrumentations/session/sessionHandler.ts
codecapitano: Oh good catch, yes this is correct. Will merge them.

codecapitano:
> in which version is this planned to be released?

@pawan-regoti 
A first experimental version of it will be available with the next release of the web-sdk but it is in experimental state and not meant for production use!

So we expect some refactorings till a v1 of this feature is ready. Also features like sampling based on sessions are not implemented yet (we will start working on it soonish).

Cheers,
Marco

cc @cedricziel

PR #353, opened 2023-10-24
fix(web-sdk): disable keepalive when large payload

### LINKED ISSUE #192: Error when sending data to the server when body size > 64kb and keepalive fetch option enable.
## Description

The issue come from the keepalive options in the FetchTransport: https://github.com/grafana/faro-web-sdk/blob/88011e78770c1821a81203c46966b6032addc1ec/packages/web-sdk/src/transports/fetch/transport.ts#L55

When removing it the bug disapear.

This bug **only happen** when:
- Using DocumentLoadInstrumentation from @opentelemetry/instrumentation-document-load
- Spliting chunk (https://webpack.js.org/plugins/split-chunks-plugin/) whatever configuration
- Using FetchInstrumentation from @opentelemetry/instrumentation-fetch
- Using BatchSpanProcessor from @opentelemetry/sdk-trace-base
- Using XMLHttpRequestInstrumentation from @opentelemetry/instrumentation-xml-http-request

Screenshot from 2023-05-15 17-31-28

## Steps to reproduce

Using all informations said above to configure faro and your project.

## Expected behavior

Trace corretly exported.

## Actual behavior

Trace are not exported because of the error screenshot in description

## Environment

- **SDK version:** 1.0.2
- **SDK instrumentations:** 1.0.2
- **Device type:** computer
- **Device name:** Dell inspiron
- **OS:** Ubuntu 20.04
- **Browser:** Chrome latest

## Demo

It is working fine without customizing splitChunks on the same projet with a npm run serve in local. (Possibility that is not working with a build as it's the default configuration)

If I copy paste the load as curl in terminal it is working fine. 
If I copy past as fetch and use it directly in my code it's working fine.

If I use the SimpleSpanProcessor it's working.

The request block in pending even if the promise already rejected:
Screenshot from 2023-05-15 18-25-26

Other call to grafana-agent in the same page are working fine.

In the zone.js:975 file from the stack trace the line is: `return patchDelegate(this, arguments);`
If I do a `console.log(patchDelegate, this, arguments)`
I get:
Screenshot from 2023-05-15 18-34-26

Important to note the `Exception: TypeError:` in the callee do not appear just here but in other above call so it can be a wrong path to follow

In the fetch.js:233 file from the stack trace the line is: `return original.apply(this, args);` -> The issue STILL happen when we remove the fetch instrumentation
If I do a `console.log(original, original.apply, this, args)`
I get:
Screenshot from 2023-05-15 18-41-44

charsleysa:
## Why

Fetch limits payloads to 64 kibibytes when using the keepalive option by throwing a network error.

## What

Turn off keepalive when payload length is over 60_000. This allows the request to succeed with the caveat that it can still be canceled by the browser during page unload. This is a balanced compromise until a better solution is found.

## Links

Fixes #192

Standard
https://fetch.spec.whatwg.org/#http-network-or-cache-fetch

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next
 
-- Fix: disable keepalive in web-sdk fetch transport when the payload length is over 60_000
+- Fix: disable keepalive in web-sdk fetch transport when the payload length is over 60_000 (#353)
 
 ## 1.2.0 – 1.2.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next
 
+- Fix: disable keepalive in web-sdk fetch transport when the payload length is over 60_000
+
 ## 1.2.0 – 1.2.2
 
 - Feat: Detect if Faro is running inside K6 browser to distinguish between lab and field data (#263).
@@ -10,7 +12,6 @@
 - Deps: Dependency updates
 - Deps: CVE-2023-45133 - Babel vulnerable to arbitrary code execution when compiling specifically
   crafted malicious code
-- Fix: disable keepalive in web-sdk fetch transport when the payload length is over 60_000
 
 ## 1.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Deps: Dependency updates
 - Deps: CVE-2023-45133 - Babel vulnerable to arbitrary code execution when compiling specifically
   crafted malicious code
+- Fix: disable keepalive in web-sdk fetch transport when the payload length is over 60_000
 
 ## 1.2.0
 
--- a/packages/web-sdk/src/transports/fetch/transport.test.ts
+++ b/packages/web-sdk/src/transports/fetch/transport.test.ts
@@ -27,6 +27,17 @@ const item: TransportItem<LogEvent> = {
   meta: {},
 };
 
+const largeItem: TransportItem<LogEvent> = {
+  type: TransportItemType.LOG,
+  payload: {
+    context: {},
+    level: LogLevel.INFO,
+    message: Buffer.alloc(60_000, 'I').toString('utf-8'),
+    timestamp: new Date().toISOString(),
+  },
+  meta: {},
+};
+
 describe('FetchTransport', () => {
   beforeEach(() => {
     fetch.mockClear();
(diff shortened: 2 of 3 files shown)


codecapitano:
Hi @charsleysa,

Thanks a lot for this PR.

By using Fetch without the `keepalive` option we would not leverage the browsers beacon API which may lead to data loss because pending requests are not send if a visitor navigates away.  

So unfortunately we can't automatically decide for our users if Faro sends beacon requests or standard requests. 

An idea could be to provide a config option so that Faro users can set this as an deliberate decision with knowing the downsides in terms of data loss.


_**Quick note for readers who are not aware of what beacons are and what this limitation is about:**_
_By using Fetch with keepalive set, it uses the beacon API which ensures that we do not loose data by making pending requests outlive page navigations._ 
_Browsers define a max size limit for the the sum of all pending beacon requests which is 64kib._


===

**Question:**
May I ask if you hit the limit with the native Faro instrumentations or do you use the Web-Tracing instrumentation (which is a wrapper around otel)? 

Cheers,
Marco

cc @cedricziel

charsleysa:
Hi @codecapitano

I believe is your concern about data loss is moot because the data is already lost.

If you try to send more than 64 kibibytes in a single fetch request with keepalive enabled it fails and the data is lost.

This change simply allows data that would otherwise be a guaranteed loss an opportunity to be sent successfully. Also keep in mind that not all major browsers support keepalive anyway.

The only choices a user could have here is guaranteed loss or possible success, so it makes sense to make this an automatic decision as it's highly unlikely for a user to desire a scenario where the data is a guaranteed loss.

In regards to your question, I believe it is the wrapper. It's the TracingInstrumentation provided by @grafana/faro-web-tracing.

codecapitano:
Hi @charsleysa 

> This change simply allows data that would otherwise be a guaranteed loss an opportunity to be sent successfully.

> The only choices a user could have here is guaranteed loss or possible success, so it makes sense to make this an automatic decision as it's highly unlikely for a user to desire a scenario where the data is a guaranteed loss.

Oh yes, sure this makes absolutely sense. Didn't see the forest for the trees here.

> Also keep in mind that not all major browsers support keepalive anyway.

Yes this is a current limitation as we do not provide any fallback.

> In regards to your question, I believe it is the wrapper. It's the TracingInstrumentation provided by @grafana/faro-web-tracing.
All right thanks. May I ask for kind of data you nee/use the tracing insrumentation? Is it to instrument fetch/xhr requests or any other things we are currently missing in Faro?

codecapitano (commented):
As discussed it makes a lot of sense to go with this solution. 
Please update the Changelog and provide a test for the new functionality so I can approve the PR.


Thank you so much 🙏

charsleysa:
@codecapitano I've updated the changelog and added a test

codecapitano (approved):
Awesome and thanks a lot for fix 🥳  🙏

AMontagu:
Thanks a lot @charsleysa 

How can we expect a release with this fix ? I see a new version was just published yesterday but do not contains this one.

PR #377, opened 2023-11-09
Add x-faro-session header to identify client session to server

cedricziel:
## Why

In order to propagate a client-side session id to the server side without needing a trace necessarily.

## What

Propagate an `x-faro-session` header to the server containing the session ID.

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/experimental/instrumentation-fetch/src/instrumentation.test.ts
+++ b/experimental/instrumentation-fetch/src/instrumentation.test.ts
@@ -1,6 +1,6 @@
 import { initializeFaro } from '@grafana/faro-core';
 import { mockConfig } from '@grafana/faro-core/src/testUtils/mockConfig';
-import { SessionInstrumentation } from '@grafana/faro-web-sdk';
+import { makeCoreConfig, SessionInstrumentation } from '@grafana/faro-web-sdk';
 
 import { makeFaroRumHeaderValue } from './constants';
 import { FetchInstrumentation } from './instrumentation';
@@ -41,7 +41,9 @@ describe('FetchInstrumentation', () => {
 
     const sessionInstrumentation = new SessionInstrumentation();
 
-    const faroInstance = initializeFaro(mockConfig({ instrumentations: [instrumentation, sessionInstrumentation] }));
+    const faroInstance = initializeFaro(
+      makeCoreConfig(mockConfig({ instrumentations: [instrumentation, sessionInstrumentation] }))!
+    );
     const sessionId = faroInstance.api.getSession()!.id as string;
     expect(sessionId).not.toBe('');
 
--- a/experimental/instrumentation-fetch/src/instrumentation.test.ts
+++ b/experimental/instrumentation-fetch/src/instrumentation.test.ts
@@ -1,8 +1,8 @@
 import { initializeFaro } from '@grafana/faro-core';
 import { mockConfig } from '@grafana/faro-core/src/testUtils/mockConfig';
 
-import { FetchInstrumentation } from './instrumentation';
 import { makeFaroRumHeaderValue } from './constants';
+import { FetchInstrumentation } from './instrumentation';
 
 describe('FetchInstrumentation', () => {
   it('initialize FetchInstrumentation with default options', () => {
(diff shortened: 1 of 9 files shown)


experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -84,6 +85,11 @@ export class FetchInstrumentation extends BaseInstrumentation {
       initCopy.body = body;
     }
 
+    // add faroRumHeader to the request headers
+    if (faro.api.getSession() !== undefined && faro.api.getSession()!.id !== undefined) {
```
codecapitano, line 89: NIT:
```ts
const sessionId = faro.api.getSession()?.id
if (if sessionId != null) {...} // != null check for null and undefined values

File: experimental/instrumentation-fetch/src/instrumentation.test.ts [hunk 2]
```diff
@@ -36,7 +37,10 @@ describe('FetchInstrumentation', () => {
     const instrumentation = new FetchInstrumentation({
       testing: true,
     });
-    instrumentation.initialize();
+
+    const sessionId = 'test-session-id';
+
+    initializeFaro(mockConfig({ instrumentations: [instrumentation], session: { id: sessionId } }));
```
codecapitano, line 43: Soon we will remove the old session handling and only use the new session management. This will cause this test to fail.

I think the best way is to spy on the `createSession()` function and use `mockReturnValueOnce` to return a meta with a  fixed sessionId.

experimental/instrumentation-fetch/src/instrumentation.ts
```diff
@@ -84,6 +86,12 @@ export class FetchInstrumentation extends BaseInstrumentation {
       initCopy.body = body;
     }
 
+    // add Faro RUM header to the request headers
+    const sessionId = faro.api.getSession()?.id;
+    if (sessionId !== null) {
```
codecapitano, line 91: The id can only be undefined or string. 
Here the ` strict inequality operator (!==)` checks for `null` only.

To check for both `null` and `undefined` we use the ` inequality operator (!=)` 

```suggestion
    if (sessionId != null) {
```

experimental/instrumentation-xhr/src/instrumentation.ts
```diff
@@ -62,6 +63,15 @@ export class XHRInstrumentation extends BaseInstrumentation {
           return instrumentation.originalSend.apply(this, [body === undefined ? null : body]);
         }
 
+        // add Faro RUM header to the request headers
+        const sessionId = faro.api.getSession()?.id;
+        if (sessionId !== null) {
+          instrumentation.originalSetRequestHeader.apply(this, [
+            faroRumHeader,
+            makeFaroRumHeaderValue(sessionId as string),
```
codecapitano, line 71: Why don't need to cast `sessionId` to a string.

I think you saw an error before because  you checked for strict null instead for null and undefined so TS could not derive from the condition if session id is undefined or string.

codecapitano (changes requested):
Awesome Cedric! Just some nits.

codecapitano (changes requested):
Nice one, I forgot one thing in yesterday's review.
Please update the README  to reflect that the instrumentation automatically adds custom HTTP headers.

Because the additional headers will break CORS and users need to allow the new headers!

experimental/instrumentation-xhr/src/instrumentation.ts
```diff
@@ -65,10 +65,10 @@ export class XHRInstrumentation extends BaseInstrumentation {
 
         // add Faro RUM header to the request headers
         const sessionId = faro.api.getSession()?.id;
-        if (sessionId !== null) {
+        if (sessionId != null) {
           instrumentation.originalSetRequestHeader.apply(this, [
             faroRumHeader,
-            makeFaroRumHeaderValue(sessionId as string),
+            makeFaroRumHeaderValue(sessionId),
           ]);
```
codecapitano, line 72: Build fails because of a linter error.

this needs to be in one line.

```ts
          instrumentation.originalSetRequestHeader.apply(this, [faroRumHeader, makeFaroRumHeaderValue(sessionId)]);

```

codecapitano (approved):
Nice one 🔥

Starefossen:
For the reference, this change broke cors preflight for Grafana agent with default config.

PR #384, opened 2023-11-17
change: always add x-faro-session-id HTTP header

codecapitano:
## Why

Because of a backend issue—which got fixed—we needed to disable to always add the new x-faro-session-id header.

## What
* Always add x-faro-session-id header

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,13 @@
 
 ## Next
 
+## 1.2.8
+
+- Change: Custom x-faro-session-id header will now be added to legacy sessions as well (#384).
+
 ## 1.2.7
 
-- Hotfix: disable custom x-faro-session-id header for legacy sessions ()
+- Hotfix: disable custom x-faro-session-id header for legacy sessions (#383)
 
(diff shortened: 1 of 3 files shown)


cedricziel (approved):

Starefossen:
Should Grafana Agent default be updated as well? Currently only `content-type` header is allowed by default which causes preflight requests to fail. https://github.com/grafana/agent/blob/f1871dfd2a0acf5832d5f9d4aa155be34705b529/component/faro/receiver/handler.go#L72

codecapitano:
> Should Grafana Agent default be updated as well? Currently only `content-type` header is allowed by default which causes preflight requests to fail. https://github.com/grafana/agent/blob/f1871dfd2a0acf5832d5f9d4aa155be34705b529/component/faro/receiver/handler.go#L72

Hi @Starefossen,
Thanks a lot and yes it should. 
We'll update the agent. 

Cheers,
Marco

PR #385, opened 2023-11-21
Session based sampling V0

codecapitano:
## Why

With this PR we introduce a first version of session based sampling.
We use a head based sampling strategy based on sessions.
This means, if a session is not part of the sample we drop all events related to this session.

We calculate a sampling decision on the following points in time:
- On initialize, if either no session is persisted in web-storage or if the persisted session is invalid
- When we extend a session
- When the setSession() method is called outside of Faros' default session logic (if users "manually" call the function)
 
Settings:
- Users can either config a sample rate or use a sampler function to define a sample based on a custom calculation
  -  Faro inject a `context` parameter which current;y only contains the metas. 
  - The function returns a number
- If sampleRate or the result of the sampler function is not a number Faro sets the sample rate to 0 means we do not send any signals. This is to prevent costs rising up but will lead to data loss. _We may change this!_
- By default the sample rate is set to 1 (100%) means we always sent all events.

## What
* Provide first impl. of session based sampling

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/session/instrumentation.test.ts
+++ b/packages/web-sdk/src/instrumentations/session/instrumentation.test.ts
@@ -107,7 +107,7 @@ describe('SessionInstrumentation', () => {
     const transport = new MockTransport();
 
     const mockNewSessionId = '123';
-    mockStorage[STORAGE_KEY] = JSON.stringify(createUserSessionObject(mockNewSessionId));
+    mockStorage[STORAGE_KEY] = JSON.stringify(createUserSessionObject({ sessionId: mockNewSessionId }));
 
     initializeFaro(
       mockConfig({
@@ -130,7 +130,7 @@ describe('SessionInstrumentation', () => {
     const transport = new MockTransport();
 
     const mockNewSessionId = '123';
-    const mockUserSession = createUserSessionObject(mockNewSessionId);
+    const mockUserSession = createUserSessionObject({ sessionId: mockNewSessionId });
 
     mockStorage[STORAGE_KEY] = JSON.stringify(mockUserSession);
 
@@ -208,7 +208,7 @@ describe('SessionInstrumentation', () => {
   });
 
   it('creates session meta with id from persisted session for valid tracked session which is within maxSessionPersistenceTime.', () => {
-    const mockUserSession = createUserSessionObject('persisted-session');
+    const mockUserSession = createUserSessionObject({ sessionId: 'persisted-session' });
     mockUserSession.sessionMeta = {
       id: 'persisted-session',
       attributes: {
@@ -238,7 +238,7 @@ describe('SessionInstrumentation', () => {
   });
 
   it('creates new session meta with new sessionId for invalid tracked session which is within maxSessionPersistenceTime.', () => {
-    const mockUserSession = createUserSessionObject('persisted-session');
+    const mockUserSession = createUserSessionObject({ sessionId: 'persisted-session' });
     mockUserSession.sessionMeta = {
       id: 'persisted-session',
     };
@@ -265,7 +265,7 @@ describe('SessionInstrumentation', () => {
   });
 
   it('Deletes persisted session if maxSessionPersistenceTime is reached and creates new session meta.', () => {
-    const mockUserSession = createUserSessionObject('persisted-session');
+    const mockUserSession = createUserSessionObject({ sessionId: 'persisted-session' });
     mockUserSession.sessionMeta = {
       id: 'persisted-session',
     };
--- a/packages/web-sdk/src/instrumentations/session/instrumentation.test.ts
+++ b/packages/web-sdk/src/instrumentations/session/instrumentation.test.ts
@@ -20,7 +20,7 @@ import {
   SESSION_INACTIVITY_TIME,
   STORAGE_KEY,
 } from './sessionManager';
-import * as samplingModule from './sessionManager/sampling';
+import * as samplingModuleMock from './sessionManager/sampling';
 import { createUserSessionObject } from './sessionManager/sessionManagerUtils';
 
 describe('SessionInstrumentation', () => {
@@ -45,7 +45,7 @@ describe('SessionInstrumentation', () => {
   afterEach(() => {
     jest.clearAllMocks();
     jest.clearAllTimers();
-    jest.spyOn(samplingModule, 'isSampled').mockRestore();
+    jest.spyOn(samplingModuleMock, 'isSampled').mockRestore();
     mockStorage = {};
   });
 
@@ -302,7 +302,7 @@ describe('SessionInstrumentation', () => {
   it('Removes items which are not part of the sample.', () => {
     const transport = new MockTransport();
 
-    jest.spyOn(samplingModule, 'isSampled').mockReturnValue(true);
+    jest.spyOn(samplingModuleMock, 'isSampled').mockReturnValue(true);
 
     const { api } = initializeFaro(
       mockConfig({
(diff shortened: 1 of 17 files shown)


cedricziel (approved):
LGTM! Do we send `meta.session.attr.is_sampled` to the backend?

packages/web-sdk/src/instrumentations/session/instrumentation.ts
```diff
@@ -77,7 +62,10 @@ export class SessionInstrumentation extends BaseInstrumentation {
 
     if (isUserSessionValid(userSession)) {
       sessionId = userSession?.sessionId;
-      sessionAttributes = userSession?.sessionMeta?.attributes;
+      sessionAttributes = {
+        ...(userSession?.sessionMeta?.attributes ?? {}),
```
bfmatei, line 66: Fallbacking to `{}` is not needed in the case of `undefined` or `null`

packages/web-sdk/src/instrumentations/session/instrumentation.ts
```diff
@@ -103,17 +91,42 @@ export class SessionInstrumentation extends BaseInstrumentation {
...
+        })
+      );
+
+      this.notifiedSession = initialSessionMeta;
       this.api.setSession(initialSessionMeta);
 
-      const sessionManager = this.getSessionManagerInstanceByConfiguredStrategy(this.metas.value.session?.id);
+      const { updateSession } = new SessionManager();
+
+      this.transports?.addBeforeSendHooks(...this.transports.getBeforeSendHooks(), (item) => {
```
bfmatei, line 112: Q: is it really needed to re-add the existing hooks?

packages/web-sdk/src/instrumentations/session/instrumentation.ts
```diff
@@ -103,17 +91,42 @@ export class SessionInstrumentation extends BaseInstrumentation {
...
+        updateSession();
+
+        const attributes = item.meta.session?.attributes;
+
+        if (Boolean(attributes?.['isSampled'])) {
+          const { isSampled: _, ...restAttributes } = attributes as any;
```
bfmatei, line 118: `as any` 🫠

packages/web-sdk/src/instrumentations/session/sessionManager/index.ts
```diff
@@ -1,4 +1,7 @@
 export { PersistentSessionsManager } from './PersistentSessionsManager';
 export { VolatileSessionsManager } from './VolatileSessionManager';
+
 export * from './sessionConstants';
```
bfmatei, line 4: Nitpicky: I know it's a little cumbersome to write the exports again and again but usually it's best to avoid exporting everything.

packages/web-sdk/src/instrumentations/session/sessionManager/sampling.ts
```diff
@@ -0,0 +1,15 @@
+import { faro } from '@grafana/faro-core';
+
+export function isSampled() {
```
bfmatei, line 3:
```suggestion
export function isSampled(): boolean {
```

packages/web-sdk/src/instrumentations/session/instrumentation.test.ts
```diff
@@ -8,11 +14,13 @@ import * as createSessionMock from '../../metas/session';
 
 import { SessionInstrumentation } from './instrumentation';
 import {
+  FaroUserSession,
   MAX_SESSION_PERSISTENCE_TIME,
   SESSION_EXPIRATION_TIME,
   SESSION_INACTIVITY_TIME,
   STORAGE_KEY,
 } from './sessionManager';
+import * as samplingModule from './sessionManager/sampling';
```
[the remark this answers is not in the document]
codecapitano: In this case it's needed to make the spy work. 
I only need it to mock a return value. 
Let me give ir a better name which reflects its purpose.

packages/web-sdk/src/instrumentations/session/instrumentation.ts
codecapitano: Oh reall, nice. TIL!

packages/web-sdk/src/instrumentations/session/instrumentation.ts
codecapitano: Oh good catch, no it's not. 
Will remove it.

packages/web-sdk/src/instrumentations/session/instrumentation.ts
codecapitano: Yeah you're rigth it's smelly as hell. 💀
Let me change it.

packages/web-sdk/src/instrumentations/session/instrumentation.ts
```diff
@@ -103,17 +91,42 @@ export class SessionInstrumentation extends BaseInstrumentation {
...
 
-      if (sessionManager != null) {
-        this.transports?.addBeforeSendHooks(...this.transports.getBeforeSendHooks(), (item: any) => {
-          sessionManager?.updateSession();
           return item;
-        });
-      }
+        }
+
+        return null;
```
SimenB, line 128: I don't understand how this can work with batching

if batching is enabled, `execute` both enqueues it _and_ calls `instantSend`: https://github.com/grafana/faro-web-sdk/blob/9dca6e8e99e32cade68a700059cae515134902e5/packages/core/src/transports/initialize.ts#L147-L166

`instantSend` then applies this filter: https://github.com/grafana/faro-web-sdk/blob/9dca6e8e99e32cade68a700059cae515134902e5/packages/core/src/transports/initialize.ts#L121

It will then remove the `isSampled` attr (return on line 125)

It then loops over all transports, and only sends if the transport returns `false` for `isBatched()`

https://github.com/grafana/faro-web-sdk/blob/9dca6e8e99e32cade68a700059cae515134902e5/packages/core/src/transports/initialize.ts#L127-L134

We only have a batched transport, so nothing happens.

---

Then we wait for the batching interval to trigger: https://github.com/grafana/faro-web-sdk/blob/9dca6e8e99e32cade68a700059cae515134902e5/packages/core/src/transports/batchExecutor.ts#L54

This invokes `sendFn`, which is `batchedSend`: https://github.com/grafana/faro-web-sdk/blob/9dca6e8e99e32cade68a700059cae515134902e5/packages/core/src/transports/initialize.ts#L105-L118

It then does the _same_ `applyBeforeSendHooks` on the item that `instantSend` has already filtered, so it hits this `return null`, meaning it doesn't send the item to the backend.

However, `instantSend` never actually sent it to the backend, so the item is never actually sampled.

---

I think what needs to happen is that `item.meta.session.attributes.isSampled` is never removed until the item is actually sent to the backend, not during the _check_ for whether or not it's sampled.

SimenB: maybe something like this: https://github.com/SimenB/faro-web-sdk/commit/39f62ad5dee44a045a81ed79f1b5fb99be002e46

codecapitano: Hey Simen thanks a lot again and you are right, very good catch! 🙏 

I'll fix this and will change the cleanup logic.
The important thing I would like to avoid is that a specific transport need to take care about cleaning up the meta. 

Cheers,
Marco

codecapitano: But yes, letting batchExecuter be responsible to clone will prevent similar issues to arise.

codecapitano: > maybe something like this: [SimenB@39f62ad](https://github.com/SimenB/faro-web-sdk/commit/39f62ad5dee44a045a81ed79f1b5fb99be002e46)

Oh I am so sorry totally missed that.
Give me a few minutes to look at it.

And thanks a lot again, the contributions, reviews and input is great and super helpful!

codecapitano: Found another bug:

[`if (attributes && Boolean(attributes?.['isSampled']))`](https://github.com/grafana/faro-web-sdk/blob/9c729be7e4c315d9ee75ec0ae5678c8f4a2f355a/packages/web-sdk/src/instrumentations/session/instrumentation.ts#L117-L118) 
won`t work because attributes are always strings.

PR #387, opened 2023-11-22
chore: adjust maxSessionPersistenceTime an export session related functionality

codecapitano:
## Why

In persistence tracking mode Faro picks up the valid session it finds in local storage.

If multipe users would share the same browser this may lead to a situation where the new site visitor will still receive the sessionId of the previous user.

We should make it easy for customers to delete sessions at the point in time when they consider  a session as finsihed.

And we should provide a shorter maxPersistenceTime to mitigate the problem with a sane (TBD) default setting.
Currently it is the max inactivity time + one minute buffer (16 minutes).

## What
* Export the session manager as well as some other props to make it as easy as possible to manually delete sessions
* Change default value of `maxSessionPersistenceTime` to 16 minutes.

## Links

[Docs update](https://github.com/grafana/website/pull/16705)

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## 1.2.8
 
 - Change: Custom x-faro-session-id header will now be added to legacy sessions as well (#384).
+- Change: Lower maxSessionPersistenceTime for tracked session and export tooling to mak eit easier manually remove a persisted session (#387)
```
cedricziel, line 8:
```suggestion
- Change: Lower maxSessionPersistenceTime for tracked session and export tooling to make it easier manually remove a persisted session (#387)
```

cedricziel (approved):
build is failing :)

PR #391, opened 2023-11-23
chore: upgrade otel deps

codecapitano:
## Why
Some users of the Web-Tracing package reported problems due to incompatible OTEL versions.
To give the package more stability we declare the necessary versions as peerDependencies.

I also manually tested it with the Faro demo and could not find any problems.

## What
* Remove outdated resolution objects and merge packages back
* Upgrade OTEL dependencies
* Declare OTEL dependencies as peerDependencies

Note (**solved**):
Some otel packages depend on a broken otel.js version which sends time as an object instead of a timestamp.
Even with forcing the newest package version, which was mentioned that it fixed it, the  problem still persists.

See issues: 
* https://github.com/open-telemetry/opentelemetry-js/issues/4216
* https://github.com/open-telemetry/opentelemetry-js/issues/4202

Status 400 because of malformed time properties:
image

Example for package which references the broken version:
```
"@opentelemetry/instrumentation-document-load@^0.34.0":
  version "0.34.0"
  resolved "https://registry.yarnpkg.com/@opentelemetry/instrumentation-document-load/-/instrumentation-document-load-0.34.0.tgz#873075b3484fc8874f8aeb7ffcc8aac487f10dc0"
  integrity sha512-KvhQ7UDhHu/MCNeqm8rgeB82aM8D2sLYHPX65mamY9iJiXTe6cWzobZOgej5UjZqik8AE+z6mdiNupBiD7Y8lg==
  dependencies:
    **"@opentelemetry/core" "^1.8.0"**
    "@opentelemetry/instrumentation" "^0.45.1"
    "@opentelemetry/sdk-trace-base" "^1.0.0"
    "@opentelemetry/sdk-trace-web" "^1.15.0"
    "@opentelemetry/semantic-conventions" "^1.0.0"
```

**Update:**
The issue is fixed.
Screenshot 2023-11-27 at 11 54 53

We can either wait till the packages use the newest version or downgrade.
Thoughts?

cc @cedricziel 

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated


packages/web-tracing/package.json
```diff
@@ -53,22 +53,34 @@
...
+    "@opentelemetry/instrumentation-xml-http-request": "^0.45.1",
+    "@opentelemetry/otlp-transformer": "^0.45.1",
+    "@opentelemetry/resources": "^1.18.1",
+    "@opentelemetry/sdk-trace-base": "^1.18.1",
+    "@opentelemetry/sdk-trace-web": "^1.18.1",
+    "@opentelemetry/semantic-conventions": "^1.18.1"
   },
-  "resolutions": {
-    "@opentelemetry/api-metrics": "^0.29.1"
+  "peerDependencies": {
```
SimenB, line 70: if you use peer deps, they should be dev dependencies and not "real" dependencies, otherwise it does nothing.

Which would be a breaking change 😅

yarn.lock
```diff
@@ -1388,352 +1381,279 @@
   dependencies:
     "@opentelemetry/api" "^1.0.0"
 
-"@opentelemetry/api@^1.0.0", "@opentelemetry/api@^1.4.0", "@opentelemetry/api@^1.4.1":
+"@opentelemetry/api@^1.0.0", "@opentelemetry/api@^1.7.0":
+  version "1.7.0"
+  resolved "https://registry.yarnpkg.com/@opentelemetry/api/-/api-1.7.0.tgz#b139c81999c23e3c8d3c0a7234480e945920fc40"
+  integrity sha512-AdY5wvN0P2vXBi3b29hxZgSFvdhdxPB9+f0B6s//P9Q8nibRWeA3cHm8UmLpio9ABigkVHJ5NMPk+Mz8VCCyrw==
+
+"@opentelemetry/api@^1.4.0":
```
SimenB, line 1389: you can run https://www.npmjs.com/package/yarn-deduplicate (followed by a `yarn install`) to deduplicate the `/api` deps

yarn.lock
```diff
@@ -1388,352 +1381,279 @@
...
 
-"@opentelemetry/core@1.15.2":
-  version "1.15.2"
-  resolved "https://registry.yarnpkg.com/@opentelemetry/core/-/core-1.15.2.tgz#5b170bf223a2333884bbc2d29d95812cdbda7c9f"
-  integrity sha512-+gBv15ta96WqkHZaPpcDHiaz0utiiHZVfm2YOYSqFGrUaJpPkMoSuLBB58YFQGi6Rsb9EHos84X6X5+9JspmLw==
-  dependencies:
-    "@opentelemetry/semantic-conventions" "1.15.2"
-
-"@opentelemetry/core@1.17.1", "@opentelemetry/core@^1.1.0", "@opentelemetry/core@^1.15.2", "@opentelemetry/core@^1.8.0":
+"@opentelemetry/core@1.17.1", "@opentelemetry/core@^1.1.0":
```
SimenB, line 1412: would be good to figure out what pulls in the old `@opentelemetry/core@1.17.1` and update it

yarn.lock
codecapitano: Awesome thanks for the hint. Will give it a try.

yarn.lock
codecapitano: I updated all otel packages to their current versions but no luck.

Also tried yarn-deduplicate which works perfect. 
But even without any dependency pointing to an old version of @opentelemetry/core the problem still persists.

Will continue next week and keep you updated.

SimenB: `yarn why @opentelemetry/core` will tell you why it's there

SimenB: When I run `yarn-deduplicate` on this branch, the `1.17.1` disappears, so I'd guess the deduplication solves it 🙂 Push it?

packages/web-tracing/package.json
```diff
@@ -53,22 +53,24 @@
...
+    "@opentelemetry/sdk-trace-base": "^1.18.1",
+    "@opentelemetry/sdk-trace-web": "^1.18.1",
+    "@opentelemetry/semantic-conventions": "^1.18.1"
   },
   "resolutions": {
-    "@opentelemetry/api-metrics": "^0.29.1"
+    "@opentelemetry/core": "^1.18.1",
+    "@opentelemetry/sdk-trace-base": "^1.18.1",
+    "@opentelemetry/sdk-trace-web": "^1.18.1"
   },
```
SimenB, line 74: are these needed? removing doesn't seem to make a difference at least

yarn.lock
```diff
@@ -1388,396 +1381,299 @@
   dependencies:
     "@opentelemetry/api" "^1.0.0"
 
-"@opentelemetry/api@^1.0.0", "@opentelemetry/api@^1.4.0", "@opentelemetry/api@^1.4.1":
+"@opentelemetry/api@^1.0.0", "@opentelemetry/api@^1.4.0":
```
SimenB, line 1384: hmm, still duped. running `yarn-deduplicate` locally does remove it


the semver range doesn't lock things down, so won't affect consumers. but still 😀

codecapitano:
Hey @SimenB just a quick update:

The problem with sending timings in LongBits instead of a string/number is solved.
It was not enough to enable `useHex` but also to explicitly disable `useLongBits` in `createExportTraceServiceRequest`.

https://vscode.dev/github/grafana/faro-web-sdk/blob/upgrade-otel-deps/packages/web-tracing/src/faroTraceExporter.ts#L12

packages/web-tracing/package.json
codecapitano: I'll double check thank you Simen!

I appreciate your support a lot. 
This is really great, thank you so much ❤️

yarn.lock
codecapitano: Yes no that I could fix the issue regrading the data format in regards to timings I'll run `yarn-deduplicate` again, test the demo and push the change if everything is  working well 😃

packages/web-tracing/package.json
codecapitano: No they aren't needed.

SimenB (commented):
image

😍

packages/web-tracing/package.json
```diff
@@ -52,23 +52,37 @@
...
+    "@opentelemetry/instrumentation-document-load": "^0.34.0",
+    "@opentelemetry/instrumentation-fetch": "^0.45.1",
+    "@opentelemetry/instrumentation-xml-http-request": "^0.45.1",
+    "@opentelemetry/otlp-transformer": "^0.45.1",
+    "@opentelemetry/resources": "^1.18.1",
+    "@opentelemetry/sdk-trace-base": "^1.18.1",
+    "@opentelemetry/sdk-trace-web": "^1.18.1",
+    "@opentelemetry/semantic-conventions": "^1.18.1"
+  },
+  "peerDependencies": {
```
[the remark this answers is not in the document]
codecapitano: > Might be it makes sense to leave them all as deps for now, and then revisit once you're ready to make a v2 release?

Good hint! 
I thought about that for a bit but was a bit unsure. 
Also, I think that now where we removed all `resolutions` the version related problems some users had are gone.

So I like the idea. Let's remove the peerDeps and revisit it in the future.

SimenB (approved):
fwiw 😀

SimenB:
👏 

Would love to see a publish as well 🙏

codecapitano:
> 👏
> 
> Would love to see a publish as well 🙏

Hey Simen, we plan to release a new version tomorrow (Wednesday).

PR #401, opened 2023-11-29
fix: remove unused OTel `api-metrics` dependency

SimenB:
## Why

`@opentelemetry/api-metrics` is unused, and it is also deprecated: https://www.npmjs.com/package/@opentelemetry/api-metrics

image

## What

I removed the unused `@opentelemetry/api-metrics` dependency declaration.

In addition, I noticed that `@opentelemetry/sdk-trace-base` is used - this is not necessary. `sdk-trace-web` and `sdk-trace-node` both re-export the entirety of the `base` package, so there's no need for an explicit dependency.

## Links

N/A

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next
 
-- Deps: remove unused `@opentelemetry/api-metrics` dependency (#401).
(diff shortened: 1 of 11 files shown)


codecapitano (approved):
Perfect, thanks a lot ❤️

PR #402, opened 2023-11-30
chore: quick fix for sampling in sessions mode

codecapitano:
## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


SimenB:
Yeah, works as a quick fix I think 👍

cedricziel (approved):

codecapitano:
> Yeah, works as a quick fix I think 👍

Thanks for the review Simen.
Since you already showed proposal for a fix:
Do you plan to work on the fix or do you not have time at the moment and would like us to implement the fix?

packages/core/src/transports/initialize.ts
```diff
@@ -156,12 +156,10 @@ export function initializeTransports(
       return;
     }
 
-    if (!config.batching?.enabled) {
-      instantSend(item);
-      return;
+    if (config.batching?.enabled) {
```
codecapitano, line 159: This change is not related, but i cleaned it up for better readability.

codecapitano:
@cedricziel Note for you: with the quick fix we do not remove the `isSampled` attribute yet. 
When the real solution is implemented the `isSampled` attribute will be removed of course

SimenB:
>Do you plan to work on the fix or do you not have time at the moment and would like us to implement the fix?

I'm not planning to, no 🙂

codecapitano:
> > Do you plan to work on the fix or do you not have time at the moment and would like us to implement the fix?
> 
> I'm not planning to, no 🙂

All right, so I do not step on your toes when working on a solution 😄 👍

PR #419, opened 2023-12-06
Add user defined session attributes to the session meta and ensure that those attrinutes are not stored in web-storage. 

codecapitano:
## Why

1. Ensure that user provided session attributes are also send if a session is resumed or extended.

2. Ensure that we do not store user defined attributes in web-storage. Otherwise Faro could easily leak sensitive information to the web-storage which is a privacy risk and may be an adoption blocker for customers.
 
## What
* Ensure that we send user defined attributes.
* Statically defined attributes (think of them as constants) during the init phase always win over dynamic attributes with the same name added by calling the `setSession()` function.

* Add a cleanup function which is called before storing session meta data in web storage 

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/demo/src/client/faro/initialize.ts
+++ b/demo/src/client/faro/initialize.ts
@@ -43,13 +43,6 @@ export function initializeFaro(): Faro {
       version: env.package.version,
       environment: env.mode.name,
     },
-    sessionTracking: {
-      session: {
-        attributes: {
-          location: 'moon',
-        },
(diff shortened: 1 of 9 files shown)


codecapitano:
Closing this PR in favor of: https://github.com/grafana/faro-web-sdk/pull/420

PR #421, opened 2023-12-08
Custom-session-id-generator

codecapitano:
## Why

Add an option to provide a custom id generator function. 
This is for custoners who use their own session naming scheme.

## What
* Add optional `sessionIdGenerator` property to the session tracking object
* enhance `createUserSessionObject()` function to use the `sessionIdGenerator`. I sessionIdGenerator is not provided Faro uses the internal shortId genearot function.

## Links

[Documentation](https://github.com/grafana/website/pull/16964)

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -34,6 +34,7 @@ export interface Config<P = APIEvent> {
     onSessionChange?: (oldSession: MetaSession | null, newSession: MetaSession) => void;
     samplingRate?: number;
     sampler?: (context: SamplingContext) => number;
+    sessionIdGenerator?: () => string;
   };
 
   user?: MetaUser;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next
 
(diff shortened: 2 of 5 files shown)


packages/core/src/config/types.ts
```diff
@@ -34,6 +34,7 @@ export interface Config<P = APIEvent> {
     onSessionChange?: (oldSession: MetaSession | null, newSession: MetaSession) => void;
     samplingRate?: number;
     sampler?: (context: SamplingContext) => number;
+    sessionIdGenerator?: () => string;
```
kpelelis, line 37: Nitpicky: I would expect a function to be named as an imperative verb (i.e. `generateSessionId()`) instead of a noun where it might indicate that it is an object. Feel free to merge though 🪨

kpelelis (approved):

packages/core/src/config/types.ts
codecapitano: Argh sorry merged before i read the comment. 
Will rename it!

PR #425, opened 2023-12-13
fix: guard against missing `isSampled` attribute

SimenB:
## Why

I'm not sure why, but we sometimes get an error from Faro with `cannot read property 'toString' of undefined`. It doesn't happen with all users, only some

## What

Makes sure there's a valid value on which to call `toString`

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 ## Next
 
-- fix (`@grafana/faro-web-sdk`): guard against missing `isSampled` (#)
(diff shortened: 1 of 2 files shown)


codecapitano (approved):
Awesome, thanks a lot Simen. 
We'll release the fix tomorrow

codecapitano (changes requested):
Missed this one: Please update the CHANGELOG.

SimenB:
Done (from mobile, so hopefully not too wonky)

codecapitano (approved):
LGTM

Issue #393, opened 2023-11-23
Page-Load Instrumentation


cedricziel:
## Description

We should have a simple and quick way to record page-loads and resources loaded as part of a page load as distinct elements and record important KPIs about them like the navigation and resource timings.

Related: #388 

## Proposed solution

Use browser performance APIs to record the navigation timings AND
* record server-timing information so that we can match a page-load with a server-side trace
* make page loads identifyable
* make associated resource timings reference the page-load

Issue #389, opened 2023-11-23
Link between page navigations events 


codecapitano:
## Description

Link between page navigations.

## Proposed solution
Add a `from` and `to` property to each navigation event where `from` is the url and id of the current page we are leaving and `to` is the url and id of the destination we navigate to.

  
## Context
This issue is related to issue #388

Issue #388, opened 2023-11-22
Navigation timings and resource timings should have a causal relation


cedricziel:
## Description

In order to understand which specific resources with which specific timings were loaded as part of a page-load, there should be an identifier that can be used to find all the resource timings that are associated with a navigation timing.

Example:
- navigation-timing: page-load-a (id: 123, name: /shopping-cart)
  - resource-timing-z (name: app.js, navigation: 123)
  - resource-timing-y (name: styles.css, navigation: 123)
- navigation-timing: page-load-b (id: 456, name: /shopping-cart)
  - resource-timing-x (name: app.js, navigation: 456)
  - resource-timing-w (name: styles.css, navigation: 456)

Then we could look at each set of navigation timings and understand which page loads were involved and calculate if it's in line with expectations.

## Proposed solution

* calculate an identifier for each navigation timing
* calculate an identifier for each resource timing
* set navigation-timing event id on resource timings to make them discoverable

codecapitano:
They are sent as event which has respective information about the page etc. in it's metas.
So the correlation could be done in the backend instead of putting more load on the web-sdk.

Wdyt?

codecapitano:
Implemented with: #434

Issue #379, opened 2023-11-14
Page loads not properly recorded when redirecting to another page/sub-app 


eskirk:
## Description

A user report indicates that their application architecture has some unexpected behavior regarding number of page loads. They have a main app, and a sub-app. The main app loads and accurately counts the number of page loads and then redirects to the sub-app, which does not record the correct number of page loads. 

## Steps to reproduce

TODO

## Expected behavior

Page loads should be recorded on both main and sub-app. 

## Actual behavior

Page loads are only counted on the main app. 

## Environment

TODO

codecapitano:
@eskirk Do you have some more information about the sub-app?
Is it an own page or is the sub-app contained withing a page, maybe loaded inside an iFrame?

Cheers,
Marco

PR #434, opened 2023-12-19
Faro navigation and resource metrics part 1

codecapitano:
## Why

Capture `PerformanceNavigationTiming` and `PerformanceResourceTiming` events form the browsers performance-timeline and add the instrumentation to the Faro default instrumentations.

### This PR also closes the following issues

**Page-Load Instrumentation #393** 
[x] make page loads identifyable
[x] make associated resource timings reference the page-load
[ ] record server-timing information so that we can match a page-load with a server-side trace => This will be done with  issue #452

**Link between page navigation events #389** 
Faro adds a `faroNavigationId` to each navigation and stores it in session-storage.
Each subsequent navigation gets this id added as `faroPreviousNavigationId`.
The first navigation has `faroPreviousNavigationId: 'unknown'`

_A note on naming:
It is important to namespace things here, such we use `faroNavigationId` instead of `navigationId`.
For example the [chrome team works on measuring soft navigations](https://developer.chrome.com/docs/web-platform/soft-navigations-experiment?hl=en) withing SPAs and will set a `navigationId` on some timings. Even if `Resource` and `Navigation` timings are not in that list (yet) it's advisable to prevent naming conflicts_ 

 **Navigation timings and resource timings should have a causal relation #388** 
[x] calculate an identifier for each navigation timing. 
[x] calculate an identifier for each resource timing
[x] set navigation-timing event id on resource timings to make them discoverable

Faro waits for the first navigation timing before it reads all Resource entries from the `PerformanceObserver`. It reads from the buffer so we do not miss any entries.

The `ResourceTimingsObserver` receives the `faroNavigationId` and attaches it to each `faroResourceEntry`.
It also attaches a  unique `faroResourceId` to each  `faroResourceEntry`.

**Page loads not properly recorded when redirecting to another page/sub-app #379** 
As of now we derive the number page loads from the `ttfb` measurement.
This is a flawed metric to derive page load and especially problematic if multiple Faro instances are loaded on a pages.
If we load a Faro sub-instance after the TTFB event, it can send this data anymore because the event already happened.
With the new instrumentation we can can count real navigations to get the number of pageLoads.

_Note:
When calculating pageLoadTimes I would recommend to filter out `Back-Forward navigations` because they are veeeery fast and can skew the numbers and they are no "real page loads" in terms of user loads the page._

### How to enable the performance instrumentation
The performance instrumentation is part of the default instrumentation which we get via `getWebInstrumentations()`
The performance instrumentation will be enabled by default eventually but as long as it is in preview, it's disabled.

It can be enabled like this:
```ts
initializeFaro({
    instrumentations: [
      ...getWebInstrumentations({
        performanceInstrumentation: true,
      }),
    ],
  });
```

### How do we capture the timings 
The Performance Instrumentation will start the `navigation` and `resource` instrumentation when it's initialize method is called and the is or switches to the [complete](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState)    state because we want to ensure that the document and all resources have been loaded so we have all the numbers to calculate the timings.

**First it instruments the Navigation timings by registering a `PerformanceObserver` on the Navigation timeline.**
Once a navigation entry is emitted, we:
* check if it's emitted on a `ignoredUrl`
* Then we get the `faroNavigationId` of the previous navigation (important in MPAs to link between navigations) from the browsers session storage or `unknown` if it is the first navigation.
* Then Faro calculates important load timings etc. and creates it's on `FaroNavigationItem` object.
  - Each navigation object has a unique `faroNavigationId` attached as well as the `faroPreviousNavigationId`.
* Then Faro stores the new `faroNavigationId` in session storage.
* Then Faro sends an event wit the name `faro.performance.navigation` and the `FaroNavigationItem` as payload
* Then it resolves the promise returning the new `FaroNavigationItem`. 

**Then, once the promise for the `FaroNavigationItem` is resolved, it instruments the browsers resource timeline by registering the respective `PerformanceObserver`.**
Once a resource entry is emitted, we:
* check if it's emitted on a `ignoredUrl`
* Then we attach the `faroNavigationId` of the parent navigation so we can match a resource to the navigation it belongs to.
* Then Faro calculates important load timings etc. and creates it's on `FaroResourceItem` object.
  - Each resource object has a unique `faroResourceId` and the `faroNavigationId` of it's parent attached.

### Percalculated metrics 
Faro calculates Performance metrics based on the Navigation and Resource timings emitted by the browser, directly on the client so we can reduce the volume and number of properties we sent and male it easier for oss users to use and interpret of those timings

timestamp-diagram

* https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming
* https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming

#### Navigation Metrics
The `FaroNavigationItem` contains a few metrics which are specific to navigation, like rendering metrics.
Additionally it contains all  metrics contained in a `FaroResourceItem`.

`visibilityState`
The visibility of site during the navigation.
This is very useful to be able to remove noise and get more accurate results when filtering out hidden navigations because browsers prioritize visible / foreground work. So tabs loaded in the background are much slower.
So hidden navigation skew the results a bit. Here's a very nice article: [Page Visibility: If a tree falls in the forest…](https://calendar.perfplanet.com/2021/page-visibility-if-a-tree-falls-in-the-forest/).

`totalNavigationTime`
The time the complete navigation took.
Taken from the `duration` property which is the `loadEventEnd - startTime`

`pageLoadTime`
The time the page took too load.
This is the time between the browser starts fetching a resource and when the document and all sub-resources have finished loading (`domComplete - fetchStart`)

`domProcessingTime`
How long it takes to load resources like e. g. images and videos and to execute Javascript which has been loaded after DOMContentLoaded event. This is when the page is already loaded and a users can interact with the page but some processing is still going on (`domComplete - domInteractive`).

_This also contains the [time frameworks and libraries take to load because they typicall wait for the DOMContentLoaded event before starting to run their code (see MDN)](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/domContentLoadedEventEnd)._ 

`domContentLoadHandlerTime`
This is the time to execute deferred scripts. 
The duration should be kept at <= 50ms (see [DOMContentLoaded event](https://developer.mozilla.org/en-US/docs/Web/Performance/Navigation_and_resource_timings#domcontentloaded_event))

`onLoadTime`
The time it takes to execute Scripts which are delayed to be started after the `load` event.
(`loadEventEnd - loadEventStart`)

`ttfb`
The time between the browser requesting a page and when it receives the first byte from the server.
(`responseStart`) or for pererendered pages (`responseStart - activationStart`).

`type`
The type of the navigation which can be `back_forward` | `navigate` | `prerender` | `reload`;

#### Resource Metrics
`name`
The resource's URL

`initiatorType`
This is the element which triggered the resource download (`initiatorType`).
This is **NOT** the "Content Type" of the resource!

`protocol`
Network protocol used to fetch a resource, as identified by the [ALPN Protocol ID (RFC7301)](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids) (`nextHopProtocol`)

`tcpHandshakeTime`
TCP handshake time (`connectEnd - connectStart`)

`dnsLookupTime`
DNS lookup time (`domainLookupEnd - domainLookupStart`)

`tlsNegotiationTime`
TLS negotiation time (`requestStart - secureConnectionStart`)

`redirectTime`
 Redirection time (`redirectEnd - redirectStart`)

`requestTime`
Request time (`responseStart - requestStart`)

`fetchTime`
Time to fetch, without redirects (`responseEnd - fetchStart`)

`serviceWorkerTime`
ServiceWorker processing time (`fetchStart - workerStart`)

`decodedBodySize`
The size (in octets) received from the fetch (HTTP or cache) of the message body after removing any applied content encoding (like gzip or Brotli). If the resource is retrieved from an application cache or local resources, it returns the size of the payload after removing any applied content encoding (taken from [MDN decodedBodySize property](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/decodedBodySize)).

`encodedBodySize`
The size (in octets) received from the fetch (HTTP or cache) of the payload body before removing any applied content encodings (like gzip or Brotli). If the resource is retrieved from an application cache or a local resource, it must return the size of the payload body before removing any applied content encoding (taken from [MDN encodedBodySize](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/encodedBodySize) property). 

`cacheHitStatus`
If the resource is loaded from cache and what type of cache it is loaded from it has.
Possible values:
* `cache` direct cache hit, directly loaded from the browser cache
* `conditionalFetch` loaded via a 304
* `fullLoad`  resource is fully loaded from the server

`renderBlockingStatus`
The renderBlockingStatus status of the resource as automatically determined by the browser or resource which have the `blocking="render"` attribute added manually.

Possible values: 
`blocking` the resource potentially blocks rendering
`non-blocking` the resource does not block rendering
`unknown`: The `renderBlockingStatus` property is currently **NOT** available in Firefox and Chrome. For theses browsers the value is undefined.

### Content-Type of a resource
Resource Timings don't tell us the `content-type` of a property. 
So we need to come up with a solution where we either do this client side or backend side.
I would say this depends on how resource heavy (like number of rules etc.) it may be to derive the content-type from a `ResourceEntry`.

I think (not validated yet) that a combination of the `initiatorType` and the information we can get form the filename may help us. 

Once our own `XHR`/`Fetch` instrumentations support support `ResourceTimings` we can get related information from the request, for example from the [destination](https://developer.mozilla.org/en-US/docs/Web/API/Request/destination) property.
But I think this could not replace the custom logic mentioned before because of the buffered resource entries emitted by the browser before Faro is even initilalized.

### Matching signals to a specific navigation
We can mach signals to a navigation either in the UI or in Faro by matching them to the current navigation. 
However it can happen that we already send signals before we receive the complete navigation timing.
We need to decide if we need to tackle this and how much effort we want to put into this.

### Reduce payload
Page can emit hundreds of resource per navigation so it's strongly advisable to reduce the size of performance-entries before sending them. I would like to tackle this in a follow up PR. 
This article has some nice  suggestions: [Compressing ResourceTiming ](https://nicj.net/compressing-resourcetiming/)

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated


packages/web-sdk/src/instrumentations/performance/types.ts
```diff
@@ -0,0 +1,56 @@
...
+  serviceWorkerProcessingTime: string;
+  //For example: Checking if content was compressed (decodedBodySize !== encodedBodySize)
+  decodedBodySize: string;
+  encodedBodySize: string;
+  // Checking if local caches were hit (transferSize should be 0)
+  cacheHitStatus: 'cache' | 'conditionalFetch' | 'fullLoad';
+  // Checking if the correct resources are render-blocking (renderBlockingStatus)
+  renderBlockingStatus: 'blocking' | 'non-blocking' | 'unknown';
+  initiatorType: string;
+  // serverTiming: PerformanceServerTiming[];
```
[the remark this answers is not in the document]
codecapitano: The will be included in an extra PR. 
They are nested objects so I'll do thsi in an extra PR.
Respective issue: https://github.com/grafana/faro-web-sdk/issues/452

demo/src/client/faro/initialize.ts
```diff
@@ -33,11 +33,10 @@ export function initializeFaro(): Faro {
         },
       }),
     ],
-    batching: {
-      // Batching is enabled by default and there is normally no reason to disable it.
-      // We did it in the demo so users can inspect each single requests sent due to certain interactions.
-      enabled: false,
-    },
+    // batching: {
```
cedricziel, line 36: why is this commented out?

codecapitano: Batching is enabled by default.
The example in the demo is made with one request per signal in mind.
I left it there as commented out code to peeps can disable batching to see a single request per signal.

But when thinking twice it makes no sense because if batching is disabled we will get a bunch of errors because the Faro fetch buffers will fill to quick (too many event in a very short amount of time).

Let me remove it

packages/web-sdk/src/instrumentations/performance/navigation.ts
```diff
@@ -0,0 +1,48 @@
+import { genShortID } from '@grafana/faro-core';
+import type { EventsAPI } from '@grafana/faro-core';
+
+import { getItem, setItem, webStorageType } from '../../utils';
+
+import { NAVIGATION_ENTRY, NAVIGATION_ID_STORAGE_KEY } from './performanceConstants';
+import { calculateFaroNavigationTiming, calculateFaroResourceTiming, entryUrlIsIgnored } from './performanceUtils';
+import type { FaroNavigationItem } from './types';
+
+export function observeAndGetNavigationTimings(
```
kpelelis, line 10: I think the name is a bit confusing here as this will only be run once and it acts as a Promise while hiding the observer underneath. I would suggest either renaming it to `getNavigationTimings` or expose it as an observer and let the caller use it accordingly (I prefer the first case)

packages/web-sdk/src/instrumentations/performance/navigation.ts
```diff
@@ -0,0 +1,48 @@
...
+    const faroPreviousNavigationId = getItem(NAVIGATION_ID_STORAGE_KEY, webStorageType.session) ?? 'unknown';
+
+    const faroNavigationEntry: FaroNavigationItem = {
+      ...calculateFaroResourceTiming(navigationEntryRaw.toJSON()),
+      ...calculateFaroNavigationTiming(navigationEntryRaw.toJSON()),
+      faroNavigationId: genShortID(),
+      faroPreviousNavigationId,
+    };
+
+    setItem(NAVIGATION_ID_STORAGE_KEY, faroNavigationEntry.faroNavigationId, webStorageType.session);
```
kpelelis, line 35: Note: I just realized we are [silencing](https://github.com/grafana/faro-web-sdk/blob/cf0375799c19529319cdc22d4b28c7a31a8651bc/packages/web-sdk/src/utils/webStorage.ts#L57) the errors from the storage. Might be worth taking care of this in the future

packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export function entryUrlIsIgnored(ignoredUrls: Array<string | RegExp> = [], entryName: string): boolean {
+  return ignoredUrls.some((url) => entryName.match(url) != null);
+}
+
+export function onDocumentReady(handleReady: () => void) {
+  if (document.readyState === 'complete') {
+    handleReady();
+  } else {
+    document.addEventListener('readystatechange', () => {
```
kpelelis, line 15: Should this event listener be removed after the event?

packages/web-sdk/src/instrumentations/performance/instrumentation.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { observeAndGetNavigationTimings } from './navigation';
+import { onDocumentReady, performanceObserverSupported } from './performanceUtils';
+import { observeResourceTimings } from './resource';
+
+export class PerformanceInstrumentation extends BaseInstrumentation {
+  readonly name = '@grafana/faro-web-sdk:instrumentation-performance';
+  readonly version = VERSION;
+
+  initialize() {
```
kpelelis, line 11: As a side note, I suggest that we provide a way for the consumers of the instrumentation(s) to pause/resume the instrumentation. The way this works for now, it will only stop doing its thing after a refresh

kpelelis (changes requested):
Thanks for @codecapitano for that ton of work. I left some comments for discussion

packages/web-sdk/src/instrumentations/performance/instrumentation.ts
codecapitano: Sounds good!

packages/web-sdk/src/instrumentations/performance/navigation.ts
codecapitano: We check if web storage avialable and usable before we call the storage api directly `isWebStorageTypeAvailable(webStorageMechanism)`.
For performance this function uses static values which are calculated when Faro is initialized.

```
function isWebStorageTypeAvailable(webStorageMechanism: StorageMechanism) {
  if (webStorageMechanism === webStorageType.local) {
    return isLocalStorageAvailable;
  }

  if (webStorageMechanism === webStorageType.session) {
    return isSessionStorageAvailable;
  }

  return false;
}
```

Of course there is a very small chance that in the meantime something happens like local|session storage is full etc.

But yes let's have look.

packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
codecapitano: Good catch! let's remove after it has been executed.

kpelelis (approved):
LGTM 🪨

Issue #430, opened 2023-12-14
Add Body Field for Measurement Logs For OTLP Transport


bdschaap:
## Description
Faro web vitals measurements are sent without a `body` field in the OTLP transport.  The `body` field is optional in the [Otel Log spec](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-body) but can cause issues with Otel Collector components (e.g. Splunk HEC Exporter).  

## Proposed solution
Set the `body` field to something like "Measurement" and/or provide a mechanism for setting a custom value.

PR #437, opened 2023-12-21
430 add body field for measurement logs for otlp transport

codecapitano:
## Why

Enhancement request from the community

> Faro web vitals measurements are sent without a body field in the OTLP transport. The body field is optional in the [Otel Log spec](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-body) but can cause issues with Otel Collector components (e.g. Splunk HEC Exporter).

## What

* Keep the current behavior but provide options to the user to configure a custom string which will be attached as a body property.
* Provide this functionality also to logs of type `Exception`
* Do a small refactoring to reduce the number of parameter in `OtelPayload.ts`
* Update docs

## Links

Resolves: #430

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/demo/src/client/faro/initialize.ts
+++ b/demo/src/client/faro/initialize.ts
@@ -8,6 +8,7 @@ import {
   ReactRouterVersion,
 } from '@grafana/faro-react';
 import type { Faro } from '@grafana/faro-react';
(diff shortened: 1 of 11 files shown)


eskirk (approved):
looks solid, good change 👍🏼

PR #463, opened 2024-01-15
fix: cortex config for demo

steven-prybylynskyi:
## Why

I just cloned a fresh repo to my WSL ubuntu distro and was not able to run the `cortex` container

See:
image

## What

Cortex config according to official docs: https://cortexmetrics.io/docs/configuration/configuration-file/


codecapitano (approved):
Thanks a lot @steven-prybylynskyi. Great that you fixed this ❤️

PR #476, opened 2024-02-06
source map upload SDK changes

eskirk:
## Why

signals (especially errors) need to include a bundle ID so the Faro receiver can decode stack traces 

## What

use the app name to grab the build ID from the global object - attached with the Faro JS bundler plugins during compilation 

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Next
 
+- Enhancement (`@grafana/faro-core`): sourcemap uploads - add `buildId` to the `MetaApp` `Meta` object (#476).
```
codecapitano, line 5: Isn't it a feature?

packages/core/src/api/meta/initialize.ts
```diff
@@ -65,6 +66,9 @@ export function initializeMetaAPI(
 
   const getView: MetaAPI['getView'] = () => metas.value.view;
 
+  const getBuildId: MetaAPI['getBuildId'] = (appName) =>
```
codecapitano, line 69: Suggestion:
`appName` is optional. Let's attach random ID if `appName` is missing.

packages/core/src/metas/registerInitial.ts
```diff
@@ -8,6 +8,9 @@ export function registerInitialMetas(faro: Faro): void {
     sdk: {
       version: VERSION,
     },
+    app: {
```
codecapitano, line 11: Do you plan to provide the function to the outside or is it only for internal use?

If not, I would propose to move `faro.api.getBuildId(faro.config.app.name)` to a dedicated module, make it a utility function and remove it from the metas API.

packages/web-tracing/src/faroTraceExporter.ts
```diff
@@ -10,7 +10,7 @@ export class FaroTraceExporter implements SpanExporter {
   constructor(private config: FaroTraceExporterConfig) {}
 
   export(spans: ReadableSpan[], resultCallback: (result: ExportResult) => void): void {
-    const traceEvent = createExportTraceServiceRequest(spans, { useHex: true, useLongBits: false });
```
codecapitano, line 13: I've built the web-sdk again to double check. 
This is working as expected.

Can you try to run the cleanup task, remove node_modules, do a yarn install and built again?

codecapitano: Ah did you already update with the main branch?
Maybe some otel update are missing?

packages/core/src/utils/sourceMaps.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function getBundleId(appName: string) {
+  return (globalObject as any)?.[`__faroBundleId_${appName}`];
+}
+
+export function getBundleIdStackMap() {
+  return (globalObject as any)?.[`__faroBundleIds`] as Map<Error, string> | undefined;
+}
+
+export function getBundleIdFromError(error: Error) {
```
kpelelis, line 14: Clarification Question: What is this used for? Reading through the code, I am assuming having different bundles for different parts/slices of the app?

packages/core/src/utils/sourceMaps.ts
```diff
@@ -0,0 +1,21 @@
+import type { ExceptionStackFrame } from '..';
+import { globalObject } from '../globalObject';
+
+export const cachedBundleIdStackFrameMap = new Map<string, ExceptionStackFrame[]>();
```
kpelelis, line 4: Can you add a test to verify that this works as a cache indeed? Something like

1. Have a function that throws an error
2. Throw error twice
3. Verify cache size the same

I am fearing this can run the client out of memory if a lot of errors happen, in which case we might need to add an invalidation technique

kpelelis (changes requested):
Nice work! I am only requesting changes for clarifications but overall LGTM

experimental/instrumentation-fetch/package.json
```diff
@@ -48,7 +48,7 @@
   },
   "homepage": "https://github.com/grafana/faro-web-sdk",
   "dependencies": {
-    "@grafana/faro-core": "^1.4.2"
+    "@grafana/faro-core": "file:.yalc/@grafana/faro-core"
```
codecapitano, line 51: Why do we need yalc? 
Lerna takes care of the linking.

Does Yarc work well with lerna?

packages/core/src/utils/sourceMaps.test.ts
```diff
@@ -0,0 +1,20 @@
+import { getBundleId } from './sourceMaps';
+
+describe('sourceMapUpload utils', () => {
+  beforeAll(() => {
+    delete (global as any).__faroBundleId_foo;
+    delete (global as any).__faroBundleIds;
```
codecapitano, line 6: I sourceMaps.ts i only see `__faroBundleId_`.
Is the __faroBundleIds list a leftover?

packages/core/src/utils/sourceMaps.ts
```diff
@@ -0,0 +1,5 @@
+import { globalObject } from '../globalObject';
+
+export function getBundleId(appName: string) {
```
codecapitano, line 3: TS cant't infer teh correct return tyoe and falls back to `any`.
Let's add a return type. `string | undefined`

codecapitano (approved):

codecapitano (approved):
LGTM 🦖 just some nits

PR #494, opened 2024-02-21
fix: crash when navigator.userAgentData is undefined

Pajn:
## Why

We have gotten several Sentry reports about a crash `Cannot read properties of undefined (reading 'brands')` from this file. Apparently some browsers does have the field `userAgentData` defined, but its value is undefined.

## What
Adds an extra check that `userAgentData` is truthy before trying to read it.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next
 
 - Enhancement (`@grafana/faro-web-sdk`): add `duration` property to a `faro.performance.resource` timing (#490).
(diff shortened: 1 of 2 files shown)


codecapitano (approved):
Great @Pajn, thank you so much 🙏

codecapitano:
@Pajn would you do me a favor and add the bug fix to the changelog.

codecapitano:
Hey @Pajn we released a patch version including your fix.

PR #499, opened 2024-02-28
480 create opentelemetry trace sampler taking faro sampling decisions into account

codecapitano:
## Why

**1. Sampling**
Faro has a probabilistic sampler, but when tracing is enabled, the sampling decision is not promoted to the OpenTelemetry-JS tracer which lead to problems if backend traces are sampled as well. 

> The implication is that when the SDK has chosen not to sample, the Otel tracer will still propagate the desire to sample downstream, creating undesired side-effects for `parentbased_always_off` scenarios.

**2. Sending a Faro event for client spans** 

Open tasks:
- [ ] update documentation

## What

* Add sampler which reads the `isSampled` flag from the session meta and sets the otel sampling decision based on that flag.
* Send a Faro event for spans of `kind=client`

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+- Feat (`@grafana/faro-web-sdk`): Enable Faro Navigation and Resource timings instrumentation by default (#482).
+- Update (`@grafana/faro-web-tracing`): Send a dedicated Faro event for traces of kind=client (#499).
+
 ## 1.3.9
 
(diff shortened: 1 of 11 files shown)


codecapitano:
Point for discussion:

The Faro event is redundant because we already have a trace / span already.
To reduce data consumption, I would propose to remove the attributes from the faro event payload because they can be obtained from the trace/span stored in Tempo.

Thoughts @cedricziel @kpelelis @eskirk

PR #519, opened 2024-03-14
fix: infinite loop if session dom storage is turned off

codecapitano:
## Why

If DOM storage is diabled teh web-sdk got stuck in a loop.

This is because in the `updateSession` we fetch a session from DOM storage, validate it and then either update the last activity timestamp of the valid session or create a new session.

In case DOM storage is not available, we always get a `null` value for stored session which is invalid.
So we only update the in-memory session meta with the new session.
Creating a new session via this function is basically a session extend, so it causes a session extend event to be send, which in turn leads to that the update function is called again, which cuases teh inifnifte loop beacuse the above steps are executed over and over again. 

*Responsible part of the update function*

```ts

    if (isUserSessionValid(sessionFromStorage)) {
      storeUserSession({ ...sessionFromStorage!, lastActivity: dateNow() });
    } else {
      let newSession = addSessionMetadataToNextSession(
        createUserSessionObject({ isSampled: isSampled() }),
        sessionFromStorage
      );

      storeUserSession(newSession);

      faro.api?.setSession(newSession.sessionMeta);
      sessionTrackingConfig?.onSessionChange?.(sessionFromStorage?.sessionMeta ?? null, newSession.sessionMeta!);
    }
```
  

## What
Check if DOM storage is availble and if not don't execute the update (session_extend).

This will still gives send data and sessionID, but: 
* the session will not be auto-extended if timed out.
* A new session will be created fo reach page load

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/demo/src/client/faro/initialize.ts
+++ b/demo/src/client/faro/initialize.ts
@@ -3,6 +3,7 @@ import { createRoutesFromChildren, matchRoutes, Routes, useLocation, useNavigati
 import {
   initializeFaro as coreInit,
   getWebInstrumentations,
+  InternalLoggerLevel,
   ReactIntegration,
   ReactRouterVersion,
 } from '@grafana/faro-react';
@@ -38,6 +39,7 @@ export function initializeFaro(): Faro {
       version: env.package.version,
       environment: env.mode.name,
     },
+    internalLoggerLevel: InternalLoggerLevel.VERBOSE,
   });
 
   faro.api.pushLog(['Faro was initialized']);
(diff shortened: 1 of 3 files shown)


demo/src/client/faro/initialize.ts
```diff
@@ -38,6 +39,7 @@ export function initializeFaro(): Faro {
       version: env.package.version,
       environment: env.mode.name,
     },
+    internalLoggerLevel: InternalLoggerLevel.VERBOSE,
```
kpelelis: I think this slipped in from debugging

kpelelis (changes requested):
LGTM! Nice catch. One small thing noted

demo/src/client/faro/initialize.ts
codecapitano: Yes, removed it 👍

kpelelis (approved):

PR #541, opened 2024-03-28
Add missing parsing time

codecapitano:
## Why

When comparing page-load time with time spent in the browser, let's say dom and render timings + response time, there is notable delta between the sum of the latter timings and page-load time.

After some investigation, the results are:

Faro is working correct and we derive all timings as they should be, but why the delta?

We did some manual math and it turns out that the delta is _mostly_ related to the time the browser needs to parse the document.

IMG_5973

So the plan is calculate parser time as well.

### How to calculate the parser time
Unfortunately the current performance timeline doesn't provide and event to retrospectively get the timestamp when the parser is starting it's work. In the image below you can see that we have `responseEnd` followed by `domInteractive`.
One could argue  that we simply can use `domInteractive - responseEnd`, but this can not be used because the browser may start parsing even before the last bit of the response has arrived.

image

When reading the [docs for domInteractive](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/domInteractive) we see that this property represents the time immediately before the documents ready state is set to "interactive"

> The domInteractive read-only property returns a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) representing the time immediately before the user agent sets the document's [readyState](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) to "interactive". ——MDN

So let's have a look if we can utilize the documents `readyState` to get that timestamp.
Turns out: yes, but!

When the parsers start it's work the`document.readyState` property changes to 'loading' (see [mdn Document: readyState property](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState)). 

This could be a suitable strategy in case we fetch the `readyState` at the beginning of the `<head>` of the `index.html` file.
But this is not working for us because we can't control when Faro is initialized. Some user may defer loading and so on. 
Also we need a good mechanism to ensure that fetching the `readyState` is done before everything else. 
Another option would be to put this work on the user side and have user adding extra logic specific to enable use fetching that timestamp. Even if we provide copy&paste code this is no option and very error prone because people will likely forget to do that.

image

### PerformanceTiming: domLoading property to the rescue
The old, deprecated, `performance.timing` object provides the `domLoading` property which is moment when the parser is starting its work.

> The legacy PerformanceTiming.domLoading read-only property returns an unsigned long long representing the moment, in milliseconds since the UNIX epoch, when the parser started its work, that is when its [Document.readyState](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) changes to 'loading' and the corresponding [readystatechange](https://developer.mozilla.org/en-US/docs/Web/API/Document/readystatechange_event) event is thrown. ——MDN

performance.timing.domLoading is deprecated but still supported in all browsers.

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming/domLoading#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

According to the [w3 docs for domLoading](https://www.w3.org/TR/2015/WD-navigation-timing-2-20150527/#widl-PerformanceNavigationTiming-domLoading) the reason for the deprecation is:

> The domLoading attribute is deprecated and may be removed in future versions of this specification. Due to differences in when a Document object is created in existing user agents, the value returned by the domLoading is implementation specific and should not be used in meaningful metrics.——W3 Docs

### Solution
The [`domLoading` property can be used to calculate the parser time](https://web.dev/articles/critical-rendering-path/measure-crp#navigation-timing). 

image

Even if the metric is different across user agents I still see value in providing the information and close the gap in numbers mentioned above.

One downside is that we are using a deprecated API, so we need to:
* ensure to not break Faro in case a browser vendor removes it 
* keep an eye on browser api updates to react early
* keep researching a better solution  

## What
* Calculate parser time by using the `domLoading` property

## Links

* [Documentation PR](https://github.com/grafana/website/pull/19129)

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
+++ b/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
@@ -126,11 +126,6 @@ export function createFaroNavigationTiming(navigationEntryRaw: PerformanceNaviga
 
 function getDocumentParsingTime(domInteractive: number): string {
   let parserStart;
-  console.log('performance.timeOrigin :>> ', performance.timeOrigin);
-  console.log('performance :>> ', performance);
-  console.log('performance.timing :>> ', performance.timing);
-  console.log('performance.timing.domLoading :>> ', performance.timing?.domLoading);
-
   if (performance.timing?.domLoading != null) {
     // the browser is about to start parsing the first received bytes of the HTML document.
     // This property is deprecated but there isn't a really good alternative atm.
--- a/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
+++ b/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
@@ -23,7 +23,7 @@ export function onDocumentReady(handleReady: () => void) {
   }
 }
 
-export function calculateFaroResourceTiming(resourceEntryRaw: PerformanceResourceTiming): FaroResourceTiming {
+export function createFaroResourceTiming(resourceEntryRaw: PerformanceResourceTiming): FaroResourceTiming {
   const {
     connectEnd,
     connectStart,
@@ -92,7 +92,7 @@ export function calculateFaroResourceTiming(resourceEntryRaw: PerformanceResourc
   }
 }
 
-export function calculateFaroNavigationTiming(navigationEntryRaw: PerformanceNavigationTiming): FaroNavigationTiming {
+export function createFaroNavigationTiming(navigationEntryRaw: PerformanceNavigationTiming): FaroNavigationTiming {
   const {
     activationStart,
     domComplete,
@@ -119,7 +119,7 @@ export function calculateFaroNavigationTiming(navigationEntryRaw: PerformanceNav
 
     type: type,
 
-    ...calculateFaroResourceTiming(navigationEntryRaw),
+    ...createFaroResourceTiming(navigationEntryRaw),
   };
 }
 
--- a/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
+++ b/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
@@ -109,6 +109,7 @@ export function createFaroNavigationTiming(navigationEntryRaw: PerformanceNaviga
   return {
     visibilityState: document.visibilityState,
     pageLoadTime: toFaroPerformanceTimingString(domComplete - fetchStart),
+    documentParsingTime: getDocumentParsingTime(domInteractive),
     domProcessingTime: toFaroPerformanceTimingString(domComplete - domInteractive),
     domContentLoadHandlerTime: toFaroPerformanceTimingString(domContentLoadedEventEnd - domContentLoadedEventStart),
     onLoadTime: toFaroPerformanceTimingString(loadEventEnd - loadEventStart),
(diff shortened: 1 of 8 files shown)


packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
```diff
@@ -119,10 +120,22 @@ export function calculateFaroNavigationTiming(navigationEntryRaw: PerformanceNav
...
+function getDocumentParsingTime(domInteractive: number): string {
+  let parserStart;
+  if (performance.timing?.domLoading != null) {
+    // the browser is about to start parsing the first received bytes of the HTML document.
+    // This property is deprecated but there isn't a really good alternative atm.
+    // For now we stick with domLoading and keep researching a better alternative.
+    parserStart = performance.timing.domLoading - performance.timeOrigin;
+  }
+
+  return toFaroPerformanceTimingString(parserStart ? domInteractive - parserStart : undefined);
```
kpelelis, line 136: Q: Is there a chance that the parser start is greater than the domInteractive?

packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
```diff
@@ -119,10 +120,22 @@ export function calculateFaroNavigationTiming(navigationEntryRaw: PerformanceNav
 
     type: type,
 
-    ...calculateFaroResourceTiming(navigationEntryRaw),
+    ...createFaroResourceTiming(navigationEntryRaw),
   };
 }
 
+function getDocumentParsingTime(domInteractive: number): string {
```
kpelelis, line 127: nit: From a design PoV it makes sense to return the number with it being formatted with `toFaroPerformanceTimingString` so that it is consistent with the other attributes

packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
kpelelis: I think this is meant to return null so that the `toFaroPerformanceTimingString` returns `unknown`. Maybe cover it with a testcase

kpelelis (changes requested):
Short questions but overall this is very well written! 🪨

packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
codecapitano: Very good hint. Let me change it 🙏

packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
codecapitano: Q: Is there a chance that the parser start is greater than the domInteractive?
No (at least what I've read in the docs)

kpelelis (approved):

PR #554, opened 2024-04-12
553 allow to exclude requests to certain urls from tracking on a global level

codecapitano:
## Why

Faro does not provide an easy to use option to globally ignore endpoints from being tracked. 
This only applies to requests to endpoints, not page URLs of a website.

This helps to reduce noise and save resources (# of requests, amount of data sent) of visitors of the web site and can remove noise from RUM data.

For example tracking the load performance to common analytics libraries all the time may not provide much value.

Currently this affects the following instruments:
* Performance
* XHR (experimental)
* Fetch (experimental)
* Performance Timeline (experimental)

## What
* Provide the `ignoreEndpoints` property to define strings and regex patterns for URLs to be excluded
* Fix some tests in XHR and Fetch instrumentation

## Links

[Cloud docs, Performance Instrumentation](https://github.com/grafana/website/pull/19386)

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/performance/instrumentation.test.ts
+++ b/packages/web-sdk/src/instrumentations/performance/instrumentation.test.ts
@@ -130,6 +130,7 @@ describe('Performance Instrumentation', () => {
       transports: [fetchTransport],
       instrumentations: [new PerformanceInstrumentation()],
       ignoreUrls: [/.*foo-analytics/, /.*.analytics.com/, 'http://example.com/awesome-image'],
+      trackResources: true,
     });
 
     const faro = initializeFaro(config);


experimental/instrumentation-performance-timeline/src/instrumentation.test.ts
```diff
@@ -280,38 +281,42 @@ describe('PerformanceTimelineInstrumentation', () => {
...
     });
 
-    const config = mockConfig({ dedupe: true, instrumentations: [instrumentation] });
-
     // Mocking the Performance object doesn't work. Skip testing its usage for now
     jest.spyOn(instrumentation, 'configureResourceTimingBuffer' as any).mockImplementation(() => {});
-    instrumentation.initialize();
-    const { api } = initializeFaro(config);
+
+    // instrumentation.initialize();
```
eskirk, line 292:
```suggestion
```

eskirk (approved):
lots of good tests, thanks for replacing the ad-hoc type with `Patterns`

codecapitano:
Too ignore endpoints for all URLs but current domain.

```ts
const currentDomain = window.location.hostname;

ignoreUrls: [new RegExp(`^(?!.*${currentDomain})`)],
```

PR #580, opened 2024-05-02
feat: LogArgsSerializer to customize how args are converter to string

### LINKED ISSUE #564: `pushLog` argument serializer
## Description

Currently the `args` array for the `pushLog` function [maps each arg to a string](https://github.com/grafana/faro-web-sdk/blob/b76c7766a1b1ad9dcd4294ef9794e55648fe8cd0/packages/core/src/api/logs/initialize.ts#L32). 

```js
{
...
  message: args
    .map((arg) => {
      try {
        return String(arg);
      } catch (err) {
        return '';
      }
    })
    .join(' '),
}
```

This works but makes logs ugly as soon as someone write `console.log("something happened" , {foo: "bar"})` as it yields `something happened [object Object]`

## Proposed solution

Extend the config with a `LogMessageSerializer` which is used instead of the hard coded `String(arg);`.
```js
{
...
  message: args
    .map((arg) => {
      return config.logMessageSerializer.serialize(arg)
    })
    .join(' '),
}
```

 Use current behaviour as a default `StringLogMessageSerializer` in the config but it could be overwritten with a e.g. `JSONLogMessageSerializer` that would do something like `JSON.stringify(arg)`

mxab:
## Why

The default behaviour in `pushLogs` stringifies all arguments in a very simple way. This results in potential `[object Object]` log messages if some one or some third party library decides to put complex logs arguments e.g. `console.info({foo: "bar"})`

## What

This PR extends the faro config with a `logArgsSerializer` paremeter that allows to override the default behaviour and put in a custom args renderer

fixes #564

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next
 
+- Enhancement (`@grafana/faro-core`): Config has now a parameter `logArgsSerializer` to set a custom serializer for log arguments (#564). This is usuful if log message args are complex and might produce `[object Object]` in the logs.
+
 ## 1.7.0
 
 - Enhancement (`@grafana/faro-web-sdk`): provide option to globally
--- a/docs/sources/developer/architecture/components/api.md
+++ b/docs/sources/developer/architecture/components/api.md
@@ -37,6 +37,14 @@ Methods:
 
 - `pushLog` - sends a log
 
+Logging arguments (`console.info('Hello', {foo: 'bar'})`) are by default each transformed into a string and concatenated by a space. If you want to use a different transformation, you can provide a custom `logArgsSerializer` function in the config object.
+
+```javascript
+initializeFaro({
+  logArgsSerializer: (args) => args.map((arg) => JSON.stringify(arg)).join(' ')
+});
+```
+
 ### Measurements
 
 Measurements are signals which can be seen as metrics.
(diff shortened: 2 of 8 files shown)


codecapitano:
Hi @mxab can you also add the new property to the core readme please.

https://github.com/mxab/faro-web-sdk/blob/155bc67cc79ce1ce60a86403e35c226c4b723ff2/packages/core/README.md#L43

mxab:
Done, do you want me to squash the changes?

codecapitano:
> Done, do you want me to squash the changes?

Hey Max you don't need to squash them yourself.
We'll squash & merge them via GH.

codecapitano (approved):
@mxab Great addition thank you so much. 🙏 
Thanks for opening the initial issue and implementing the solution as well 🥳

codecapitano:
@mxab I am so sorry but the Linter hates us.

``````javascript
initializeFaro({
  logArgsSerializer: (args) => args.map((arg) => JSON.stringify(arg)).join(' ')
});
```

You wants to see a `,` at the of the property line

mxab:
yeah I only executed the linter in the core package. thx

mxab:
at last the linter approves 😅

PR #581, opened 2024-05-03
fix: fixing import of Patterns in web tracing package

eskirk:
## Why

customer reported issue with build that was reported in this [slack thread](https://raintank-corp.slack.com/archives/C01T8C5CNTE/p1714773077383609)


codecapitano (approved):
Thanks for fixing this 🙏

PR #584, opened 2024-05-06
Update types to work with v6 data routers

codecapitano:
## Why

There was type issue with newest react (data) router.
Add specific types for `Index` and `NonIndex` routes.
We took the necessary properties from the `RouteObject` type as defined by the remix/react router.  

## What
Update types in `withFaroRouterInstrumentation` to satisfy the React Router `RouteObject` type.

### Before

Screenshot 2024-05-06 at 12 41 41

### After
Screenshot 2024-05-06 at 12 41 10

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/react/src/router/v6/types.ts
+++ b/packages/react/src/router/v6/types.ts
@@ -2,6 +2,24 @@ import type { ReactElement, ReactNode } from 'react';
 
 import type { ReactRouterLocation } from '../types';
 
+interface IndexRouteObjectV6DataRouter {
+  caseSensitive?: boolean;
+  children?: undefined;
+  element?: React.ReactNode | null;
+  index?: true;
+  path?: string;
+}
+
+export interface NonIndexRouteObjectV6DataRouter {
+  caseSensitive?: boolean;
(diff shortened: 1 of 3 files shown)


cedricziel (approved):

packages/react/src/router/v6/types.ts
```diff
@@ -2,6 +2,24 @@ import type { ReactElement, ReactNode } from 'react';
 
 import type { ReactRouterLocation } from '../types';
 
+interface IndexRouteObjectV6DataRouter {
+  caseSensitive?: boolean;
+  children?: undefined;
+  element?: React.ReactNode | null;
+  index?: true;
```
domasx2, line 9: I think `index` shouldn't be optional here?

domasx2 (approved):

packages/react/src/router/v6/types.ts
codecapitano: Good catch thank you 🙏

PR #589, opened 2024-05-08
feat(web-sdk): forward logArgsSerializer in initializeFaro to the core

LeonEck:
## Why

https://github.com/grafana/faro-web-sdk/pull/580 introduced the `logArgsSerializer` option in the core. But it isn't usable through `initializeFaro` from the web-sdk since the `makeCoreConfig` doesn't forward the option.

## What

I tried to follow the patterns of the surrounding code, which is why I introduced the new `defaultLogArgsSerializer` to use as a fallback. If `logArgsSerializer` is not set on the browser config this default is used.

I didn't find any documentation that could be updated alongside with this change. If there is a place I should document this, please let me know.

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Feature (`@grafana/faro-core`): sourcemap uploads - add `bundleId` to the `MetaApp` `Meta` object
   (#476).
 - Feature (`@grafana/faro-web-sdk`): track window dimensions via the `browser meta` (#594).
+- Feature (`@grafana/faro-web-sdk`): If `logArgsSerializer` is set in the config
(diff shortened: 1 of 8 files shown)


codecapitano (approved):
Great work thank you so much @LeonEck 🙏

codecapitano:
Hi @LeonEck 
And thanks a lot for your contribution 🙏 

> I didn't find any documentation that could be updated alongside with this change. If there is a place I should document this, please let me know.

No worries, currently we only have the cloud docs for Faro which are not oss. I'll take care of updating them.

PR #594, opened 2024-05-16
Track window dimensions

codecapitano:
## Why
Track layout viewport dimension.

This can e. g. provide insights about different behavior in different viewports, enables to filter signals by viewport size and more.

## What
* add two new properties to the browser meta: `viewportWidth` and `viewportHeight`

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [x] Documentation updated


eskirk (approved):
lgtm

Issue #452, opened 2024-01-05
Record server-timing information so that we can match a page-load with a server-side trace 


codecapitano:
## Description

Record server-timing information so that we can match a page-load with a server-side trace.

## Proposed solution
Enable this in the new Performance instrumentation

PR #608, opened 2024-06-06
feat: collect tracecontext through server-timing for nav and resource timings

cedricziel:
## Why

Create correlation between faro events for navigations by attaching a server-sent identifier

## What

Extract server-timing contents and use it to add span context.

Enhance demo to send trace parent via server-timing.

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

Fixes: #452

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -35,3 +35,4 @@ yarn-error.log*
 # App plugin
 infra/grafana/plugins/
 infra/grafana/plugins-provisioning/*.yaml
+.yarn/releases/*.js
--- a/packages/web-sdk/src/instrumentations/performance/navigation.ts
+++ b/packages/web-sdk/src/instrumentations/performance/navigation.ts
@@ -33,13 +33,18 @@ export function getNavigationTimings(
       spanId: '',
     };
 
-    for (const { name, description } of navEntryJson) {
-      if ("traceparent" === name) {
-        const [version, traceId, spanId, sampled] = description.split("-");
-
-        spanContext.traceId = traceId;
-        spanContext.spanId = spanId;
-      }
+    if ('serverTiming' in navEntryJson) {
+        const serverTiming = navEntryJson.serverTiming;
+        if (serverTiming.length > 0) {
+          for (let i = 0; i < serverTiming.length; i++) {
+            if (serverTiming[i][0] === 'traceparent') {
+              const [version, traceId, spanId, sampled] = serverTiming[i][1].split('-');
+              spanContext.traceId = traceId;
+              spanContext.spanId = spanId;
+              break;
+            }
+          }
+        }
     }
 
     const faroPreviousNavigationId = getItem(NAVIGATION_ID_STORAGE_KEY, webStorageType.session) ?? 'unknown';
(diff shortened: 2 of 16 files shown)


packages/web-sdk/src/instrumentations/performance/navigation.ts
```diff
@@ -1,12 +1,15 @@
 import { genShortID } from '@grafana/faro-core';
 import type { EventsAPI } from '@grafana/faro-core';
+import type {PushEventOptions} from "@grafana/faro-core/src/api/events/types";
```
codecapitano, line 4: `PushEventOptions` is not exported by the main barrel file file (index.ts) in Faro core.

packages/web-sdk/src/instrumentations/performance/navigation.ts
```diff
@@ -23,17 +26,38 @@ export function getNavigationTimings(
...
     }
 
+    const navEntryJson = navigationEntryRaw.toJSON();
+
+    let spanContext: Pick<SpanContext, 'traceId' | 'spanId'> = {
+      traceId: '',
+      spanId: '',
+    };
+
+    if ('serverTiming' in navEntryJson) {
```
codecapitano, line 36: This doesn't match with the data structure of `PerformanceServerTiming` wich is an object in the following structure (see [w3c docs](https://w3c.github.io/server-timing/#the-performanceservertiming-interface)).

codecapitano, line 36: `PerformanceServerTiming` objects have n explicit structure. How did you receive an array instead?

codecapitano: I think we can simmplify this function and be more expliciut. 

```suggestion
    const { serverTiming = [] }: { serverTiming: PerformanceServerTiming[] } = navEntryJson;

    for (const serverEntry of serverTiming) {
      if (serverEntry.name === 'traceparent') {
        const [traceId, spanId] = serverEntry.description.split('-');
        spanContext.traceId = traceId;
        spanContext.spanId = spanId;
        break;
      }
    }
```

codecapitano: Does thge API contract guarantee that we always get traceID and spanID?

codecapitano: Please add a test

codecapitano (changes requested):

packages/web-sdk/src/instrumentations/performance/navigation.ts
```diff
@@ -1,12 +1,15 @@
 import { genShortID } from '@grafana/faro-core';
 import type { EventsAPI } from '@grafana/faro-core';
+import type {PushEventOptions} from "@grafana/faro-core/src/api/events/types";
```
codecapitano, line 3: As per comment above, please add it to the main index.ts file so we can import it form `'@grafana/faro-core'`

codecapitano:
@cedricziel please add a proper PR description. 
Also check the todo items if applicable

packages/web-sdk/src/instrumentations/performance/navigation.ts
cedricziel: Exported

packages/web-sdk/src/instrumentations/performance/navigation.ts
cedricziel: done

.prettierignore
```diff
@@ -35,3 +35,4 @@ yarn-error.log*
 # App plugin
 infra/grafana/plugins/
 infra/grafana/plugins-provisioning/*.yaml
+.yarn/releases/*.js
```
codecapitano, line 38: Thanks ♥️

codecapitano (approved):

PR #617, opened 2024-06-18
fix: mark peer deps `react-dom` and `react-router-dom` as optional

SimenB:
## Why

`react-router-dom` is an optional dependency, and should be marked as such.

`react-dom` only exists since `react-router-dom` has its own peer dep on it.

## What

Use `peerDependenciesMeta` (works in all package managers)

## Links

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated


codecapitano (approved):
Great thanks a lot @SimenB 🙏

codecapitano:
@SimenB please add a changelog entry :)

SimenB:
@codecapitano added 👍

PR #622, opened 2024-06-19
feat: add helper functions to initialize React Router integration

SimenB:
## Why

Instead of passing a `version` field, just call a function with the version you want. Removes a nesting level, and a coupling between fields in the object (from the user's perspective)

## What

Added helper functions for v4, v5, v6 and v6 data routers

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [x] Documentation updated


codecapitano (approved):
A great usability improvement!
Thanks a lot.

The linter ails, I think it is because of line length in the changelog

SimenB:
it was in the react package's readme. fixed! 👍

codecapitano:
For reference: [pr to update cloud doc [private repo] ](https://github.com/grafana/website/pull/20407)

PR #658, opened 2024-08-01
652 navigation timing is off set

codecapitano:
## Why

The timestamp of a signal sent via the Faro API represents the exact moment when the respective API was called.
This precision is sufficient for most cases but for some it's not.

**Example:**
Faro's performance instrumentation collects the relevant data after the document's readyState changes to 'complete' [1]. When the performance item is sent via the Faro API, the timestamp reflects the moment the Faro signal is created, not the actual time when the event occurred. The option to provide a timestamp adjustment resolves this discrepancy.

To solve above issue we:
* Provide an option to to pass a timestamp overwrite to all Faro APIs (except pushTrace API)
* Calculate the actual timestamp of when a resource or navigation event occurred and pass this value as the correction timestamp

We calculate the actual occurrences of a navigation / resource item with the following code:  
`const actualTimestamp = performance.timeOrigin + entry.startTime;`

* [performance.timeOrigin](https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin) is the time of the start of the navigation context
* [entry.startTime](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/startTime): is the start time of the navigation/resource relative to the `performance.timeOrigin`.

[1] [document.readyState](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) complete denotes that the document and all sub-resources have finished loading

## What 
* Update Faro APIs to allow to inject an adjustment timestamp
* Update performance and resource instrumentation to calculate and provide an adjustment timestamp to use the timestamp when the event occurred instead of the timestamp when Faro created the Signal.

## Links

* [Docs PR (internal)](https://github.com/grafana/website/pull/21093)

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/demo/src/client/faro/initialize.ts
+++ b/demo/src/client/faro/initialize.ts
@@ -13,6 +13,7 @@ import { env } from '../utils';
 
 export function initializeFaro(): Faro {
   const faro = coreInit({
+    // url: 'https://faro-collector-prod-us-central-0.grafana.net/collect/2b420d8dde37317f7458f034cf3e3e44',
     url: `http://localhost:${env.faro.portAppReceiver}/collect`,
     apiKey: env.faro.apiKey,
     trackWebVitalsAttribution: true,
--- a/demo/src/client/faro/initialize.ts
+++ b/demo/src/client/faro/initialize.ts
@@ -13,7 +13,6 @@ import { env } from '../utils';
 
 export function initializeFaro(): Faro {
   const faro = coreInit({
-    // url: 'https://faro-collector-prod-us-central-0.grafana.net/collect/2b420d8dde37317f7458f034cf3e3e44',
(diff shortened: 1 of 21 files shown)


cedricziel (approved):

demo/src/client/faro/initialize.ts
```diff
@@ -13,6 +13,7 @@ import { env } from '../utils';
 
 export function initializeFaro(): Faro {
   const faro = coreInit({
+    // url: 'https://faro-collector-prod-us-central-0.grafana.net/collect/2b420d8dde37317f7458f034cf3e3e44',
```
cedricziel, line 16:
```suggestion
```

Issue #691, opened 2024-09-27
It seems Faro web SDK has broken mux player manifest downloads on Chrome.


discardedhunk:
## Description

Faro web SDK has broken mux player manifest downloads on Chrome.
Suddenly, at 3 a.m. on September 26, we started seeing errors from our users unable to play videos on our platform. We reached out to Mux, and they suggested removing the faro-web-sdk. Once we did, the problem was resolved.

## Steps to reproduce

Install faro-web-sdk and faro-web-tracing in javascript, along with mux-player 3.0. Trying playing a video with these installed and you should see an error in the chrome console that mux is not able to download the manifest file.
Here's a snippet from the error:
text: "t.match is not a function" (this is the error from the manifest download)
url: "https://manifest-gcp-us-east1-vop1.cfcdn.mux.com/Hz3Sqwrwts00wd6CUAW02AKKoqlg6h6cNg4YaOrR9oM01YmjV01rJKqmnqFRI02MIO1i4Kd88QXxZ4Z7EgoSpZDG13LzuaD5M4Ulj/rendition.m3u8?cdn=fastly&expires=1727985600&skid=default&signature=NjZmZWY3YzBfYTUyNDUzNTAyNDIwYzllMzk2YmM2ODUwN2ExNWFkNDAyYWEzZWRkNWFkMTllNjhiYzY5MTQ4MGIwZDhkNGExMw==&vsid=TX2DtxjimI4r1MUlLNLKSQigRpwTSW3oNw7ZJZnkWDfK01D6bI8TBLMXhxBFGwlAyn0

## Expected behavior

faro-web-SDK and web-tracing work with mux-player 3.0, on Chrome, without error. This worked in Safari. 

## Actual behavior

Errors from mux downloading manifest file

## Environment

- **SDK version:**
- https://unpkg.com/@grafana/faro-web-sdk@^1.9.0/dist/bundle/faro-web-sdk.iife.js"
- **SDK instrumentations:**
- https://unpkg.com/@grafana/faro-web-tracing@^1.9.0/dist/bundle/faro-web-tracing.iife.js"
- **Device type:**
- **Device name:**
- **OS:** MacOS, Windows
- **Browser:**
- Google Chrome

codecapitano:
Hi @discardedhunk 

I can reproduce the issue.
Seems that the `web-tracing` package causes the problem.

Now investigating why it causes the problem.

codecapitano:
[...]

PR #695, opened 2024-10-01
XHR instrumentation: properly handle URL objects and url strings

codecapitano:
## Why

we got a [report that the last Faro release broke the download of manifest files of the mux player](https://github.com/grafana/faro-web-sdk/issues/691).

In the meantime we got similar reports about failing add downloads etc.

The issue was related to the web-tracing instrumentation and happened when a URL object was provided instead of a string.
This instrumentation uses the otel.js xhr instrumentation the hood which expects to get string injected. 
For some reasons, when xhr streaming is uses, the url is URL object and not a string.

https://github.com/open-telemetry/opentelemetry-js/blob/3007d3e749029deb9dbb27dede6408a8d27a4786/packages/opentelemetry-core/src/utils/url.ts#L16

We could verify the behavior by:
- removing the web-tracing package => everything works
- Adding it back and by exchanging the regex we add by default with a string. The expression !!url.match(urlToMatch); in this case evaluates to  `undefined === urlToMach` (otel code above) is valid JS and won't cause an error.
- When we manually add a regex to `ignoreUrls` the problem appears again. So the problem is not directly related to the last Faro release. It only made us aware of it.

The underlying xml-http-instrumentation extracts the url in the following function.

```ts
protected _patchOpen() {
    return (original: OpenFunction): OpenFunction => {
      const plugin = this;
      return function patchOpen(this: XMLHttpRequest, ...args): void {
        const method: string = args[0];
        const url: string = args[1];
        plugin._createSpan(this, url, method);

        return original.apply(this, args);
      };
    };
  }
 ```
[https://github.com/open-telemetry/opentelemetry-js/blob/3007d3e749029deb9dbb27dede[…]kages/opentelemetry-instrumentation-xml-http-request/src/xhr.ts](https://github.com/open-telemetry/opentelemetry-js/blob/3007d3e749029deb9dbb27dede6408a8d27a4786/experimental/packages/opentelemetry-instrumentation-xml-http-request/src/xhr.ts#L380)

As we can see, it expects URL to be a string.
But, in the current case the url is a URL class and this case isn't handled.

So the URL object is passed down to the `urlMatches` function which also expect url to be a string.

```ts
export function urlMatches(url: string, urlToMatch: string | RegExp): boolean {
  if (typeof urlToMatch === 'string') {
    return url === urlToMatch;
  } else {
    return !!url.match(urlToMatch);
  }
}
```

The first condition evaluates to false because URL is not a string but an URL object instead.
So the else branch is executed which calls `url.match`. 
The URL class doesn't provide a match function so we get the error  `t.match is not a function`.

As we see in the screenshots, a URL object is passed down.
The other screenshots showing console.logs for both standard requests and the streaming requests
Screenshot 2024-10-01 at 11 41 33
Screenshot 2024-10-01 at 11 43 32
Screenshot 2024-10-01 at 11 38 09

## What
Extend `XMLHttpRequestInstrumentation` and override the `_patchOpen` function to handle both, url as string and as an Obejct

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-tracing/src/faroXhrInstrumentation.ts
+++ b/packages/web-tracing/src/faroXhrInstrumentation.ts
@@ -17,7 +17,7 @@ export class FaroXhrInstrumentation extends XMLHttpRequestInstrumentation {
     this.parentCreateSpan = self._createSpan.bind(this);
   }
 
-  // Patching the private method to take handle url type string or URL
+  // Patching the parent's private method to handle url type string or URL
   protected override _patchOpen() {
     return (original: OpenFunction): OpenFunction => {
       const plugin = this;
--- a/packages/web-tracing/src/faroXhrInstrumentation.ts
+++ b/packages/web-tracing/src/faroXhrInstrumentation.ts
@@ -17,6 +17,7 @@ export class FaroXhrInstrumentation extends XMLHttpRequestInstrumentation {
     this.parentCreateSpan = self._createSpan.bind(this);
   }
 
+  // Patching the private method to take handle url type string or URL
   protected override _patchOpen() {
     return (original: OpenFunction): OpenFunction => {
       const plugin = this;
--- a/packages/web-tracing/src/faroXhrInstrumentation.ts
+++ b/packages/web-tracing/src/faroXhrInstrumentation.ts
@@ -1,13 +1,13 @@
 import type { Span } from '@opentelemetry/api';
 import { XMLHttpRequestInstrumentation } from '@opentelemetry/instrumentation-xml-http-request';
 import type { XMLHttpRequestInstrumentationConfig } from '@opentelemetry/instrumentation-xml-http-request';
-import { OpenFunction } from '@opentelemetry/instrumentation-xml-http-request/build/src/types';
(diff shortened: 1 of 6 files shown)


packages/web-tracing/src/faroXhrInstrumentation.ts
```diff
@@ -0,0 +1,42 @@
...
+
+        plugin.parentCreateSpan(this, url, method);
+
+        return original.apply(this, args);
+      };
+    };
+  }
+}
+
+function isInstanceOfURL(item: any): item is URL {
```
kpelelis, line 40: I think this is a bit redundant as a function. `url instanceof URL` is explanatory enough IMO

kpelelis (approved):

packages/web-tracing/src/faroXhrInstrumentation.ts
codecapitano: Unfortunately we need a type guard here, otherwise Typescript won't derive the type.

cedricziel (approved):

Issue #702, opened 2024-10-09
[Bug] page_id is set to "/*" for URLs other the "english words"


undefinedhuman:
## Description

When integrating the very basic Faro implementation into my VueJS frontend (see sample config below) - and have routes defined via "vue-router" other then english words (such as /servers, /networks, /houses), the page_id setting is always set to "/*". Is there any way to configure this?

```typescript filename=sample.ts
initializeFaro({
  url: 'https://faro-collector-prod-eu-west-2.grafana.net/collect/<app-id>',
  app: {
    name: 'Sample',
    version: '1.0.0',
    environment: 'developement'
  },

  instrumentations: [
    ...getWebInstrumentations({
      enablePerformanceInstrumentation: true
    }),

    // Tracing package to get end-to-end visibility for HTTP requests.
    new TracingInstrumentation()
  ]
});
```

## Environment

- **SDK version:**
- **SDK instrumentations:** Web SDK
- **Device type:** Desktop
- **Device name:** Desktop PC, Manjaro
- **OS:** Manjaro
- **Browser:** Firefox 131

## Screenshots

![Image](https://github.com/user-attachments/assets/b47d8713-dc39-499f-a46a-4fdd584a7ad6)

codecapitano:
Hi @undefinedhuman need to investigate whats going on there. 

There's a way to set own pageIDs but it's not that easy to find. We'll improve this in future releases. 

Currently you can archive this with overwriting the page meta.

**Example**
```ts
initializeFaro({
   // ....
    metas: [
      () => ({
        page: {
          id: parsePageId(location.pathname),
          url: location.href,
          attributes: {
            foo: 'bar',
          },
        },
      }),
    ],
// ...
})
```

Maybe this can help you as a workaround for now?

Cheers,
Marco

undefinedhuman:
That works! Thank you very much!

As for my research: I found out if I use "regular" words it works, I first noticed it because my Entity URLs contain the UUID of the entity. So for example:

`/servers` => page_id: /servers
`/servers/UUID` => page_id: /servers/*

And for non english nouns I would say it would also convert the page_id to /*

If you need more information about the setup or if I should setup a minimal example please let me know!

codecapitano:
Hey @undefinedhuman thank you for opening the issue  and your PR as well 🙏 

The page_id parsing happens in the Faro receiver.

[...]

PR #704, opened 2024-10-10
fix(web-sdk): include pathname for page id

undefinedhuman:
## Why

I found a potential fix for https://github.com/grafana/faro-web-sdk/issues/702

As I could not find the code that actually propagates the id in the first place (I assume it's done in Grafana?)

## What

I included a default value for the parameter id of the `page` meta object.
 
## Links

https://github.com/grafana/faro-web-sdk/issues/702

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/metas/page/meta.ts
+++ b/packages/web-sdk/src/metas/page/meta.ts
@@ -2,6 +2,7 @@ import type { Meta, MetaItem } from '@grafana/faro-core';
 
 export const pageMeta: MetaItem<Pick<Meta, 'page'>> = () => ({
   page: {
+    id: location.pathname,
     url: location.href,
   },
 });


codecapitano:
Hi @undefinedhuman thank you so much for the PR but unfortunately we can't solve it that way.

The pageID set in Faro overrides the pageID creation which is done in the receiver.
Even when using `location.pathname` we can end up with long  paths for exampe which may contain "gibberish" which can't be shortened when overridden via Faro.

**A shallow description how the page_id parser's gibberish detection works:**

The gibberish detector is based on a likeliness of two letters being next to each other in western European languages, so for example "abc" is gibberish because it doesn't happen often in words. 

Note:
In a future release we want make it easier to override page_ids in Faro but it needs some sync and internal discussion before.

Cheers and thank you so much again.
Marco

undefinedhuman:
Sounds good - looking forward! And thanks a lot for the explanation!

PR #712, opened 2024-10-23
Allow additionalInstrumentations in TracingInstrumentation

### LINKED ISSUE #711: Allow additional TracingInstrumentations without re-implementing getIgnoreUrls() or faro.initialize()
## Description

When attempting to insert additional `TracingInstrumentations` other than the default OTEL tracing instrumentations, Faro web-tracing currently requires that we call `getDefaultOtelInstrumentations()` with the same arguments as https://github.com/grafana/faro-web-sdk/blob/main/packages/web-tracing/src/instrumentation.ts#L102. However, this is not possible as `getIgnoreUrls()` is a private function, so calling `faro.getIgnoreUrls()` is not possible from outside of the class. 

As an example, today I was trying to add a couple of `opentelemetry/instrumentation` into our app (which does not already use opentelemetry on the frontend). To do this I expected I could just add them to the instrumentation list, like this:

```
initializeFaro({
  instrumentations: [
    ...getWebInstrumentations(),
    new TracingInstrumentation({
      instrumentations: [
        ...getDefaultOtelInstrumentations(),
        new DocumentLoadInstrumentation(),
        new UserInteractionInstrumentation(),
      ],
    })
  ],
})
```

I found as a result of this that Faro does not get the correct values for `ignoreUrls` or `propagateTraceHeaderCorsUrls`. 

## Proposed solution

I think one way around this -- while maintaining backwards compatibility -- would be to allow an array of additional instrumentations to be given to Faro at initialization time, so that we don't have to re-implement `getIgnoreUrls`, make `getIgnoreUrls` public, or re-implement the entire `initialize()` function.

Then, `options.instrumentations` would clearly be an _override_, while `options.additionalInstrumentations` would be to add more tracing instrumentations that are not included by default.

### Example

```
initializeFaro({
  instrumentations: [
    ...getWebInstrumentations(),
    new TracingInstrumentation({
      instrumentationOptions: {
         propagateTraceHeaderCorsUrls: [ ... ],
      }
      additionalInstrumentations: [
        new DocumentLoadInstrumentation(),
        new UserInteractionInstrumentation(),
      ],
    })
  ],
})
```

## Context

This code block here is responsible for the default setup of OTEL instrumentations and configuration: https://github.com/grafana/faro-web-sdk/blob/main/packages/web-tracing/src/instrumentation.ts#L99-L108

pshahid:
## Why

This PR adds a small feature to the `faro.initialize()` API  that was opened in #711 

## What

Allow `additionalInstrumentations` for non-default tracing instrumentations to be added via `faro.initialize()` instead of re-implementing in a subclass or being forced to re-implement `getIgnoreUrls`.

## Links

Fixes #711 

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-tracing/src/instrumentation.ts
+++ b/packages/web-tracing/src/instrumentation.ts
@@ -104,6 +104,7 @@ export class TracingInstrumentation extends BaseInstrumentation {
           propagateTraceHeaderCorsUrls,
           fetchInstrumentationOptions,
           xhrInstrumentationOptions,
+          ...options.additionalInstrumentations ?? [],
         }),
     });
 
--- a/packages/web-tracing/src/types.ts
+++ b/packages/web-tracing/src/types.ts
@@ -23,6 +23,7 @@ export interface TracingInstrumentationOptions {
   instrumentations?: InstrumentationOption[];
   spanProcessor?: SpanProcessor;
   instrumentationOptions?: Omit<DefaultInstrumentationsOptions, 'ignoreUrls'>;
+  additionalInstrumentations?: Instrumentation[];
 }
 
 export type MatchUrlDefinitions = Patterns;


codecapitano:
Hi @pshahid thanks a lot for contribution, very good idea!

We need to adjust some things.

The `getDefaultOTELInstrumentations` function doesn't initialize additional instruments.

I would propose something like the following.

1. We can change the `additionalInstrumentations` property to a function and and inject the `ignoreUrls`. So users are able to instantiate the additional instruments inside that function, apply instrumentation specific config and can add the `ignoreUrls` as well.

Example:
```ts
// types.ts

export interface TracingInstrumentationOptions {
  // ...
  additionalInstrumentations?: ({ignoreUrls}: {ignoreUrls: Patterns}) => Instrumentation[];
}
```

2. Then we can add the instruments in the instrumentation class, for example like so:


```ts
// instrumenation.ts

 // ...
  const instrumentations = [
      ...getDefaultOTELInstrumentations({
        ignoreUrls: this.getIgnoreUrls(),
        propagateTraceHeaderCorsUrls,
        fetchInstrumentationOptions,
        xhrInstrumentationOptions,
      }),
    ];

    if (options.additionalInstrumentations) {
      instrumentations.push(...options.additionalInstrumentations({ignoreUrls: this.getIgnoreUrls()}));
    }

    registerInstrumentations({
      instrumentations: options.instrumentations ?? instrumentations,
    });
// ...
```

Usage will then be like this:

```ts
// in initiliazeFaro()

//...

new TracingInstrumentation({
        additionalInstrumentations({ignoreUrls}) => ([new FooInstrumentation({ignoreUrls})])
      }),

//...

```

Wdyt?

pshahid:
@codecapitano thanks so much for the reply, these are great adjustments! I'll get them incorporated into the change and re-push

Bruno-DaSilva:
One thing to double check - are there any side effects of instantiating otel instrumentations but then not using them?
We may want to only `getDefaultOTELInstrumentations()` if options.instrumentations is undefined.

codecapitano:
> One thing to double check - are there any side effects of instantiating otel instrumentations but then not using them? We may want to only `getDefaultOTELInstrumentations()` if options.instrumentations is undefined.

Th

codecapitano:
Hi @pshahid, @Bruno-DaSilva 

Thank you so much for the contribution.
While looking and testing the PR it came to me that the PR revealed some sub optimal API design of the `TracingInstrumentation` class which I think it may make sense to tackle.

With the current API design we have properties which affect default and custom instruments as well as only custom instruments. 

I think this is not easy to reason off, creates ambiguity and clutters the API in the long term.

But your idea is super valuable so I have some ideas I would like to discuss. 
And don't worry, we can either work together on this or if it's to much time to spend for you on top of your daily work, then we can take care of it.


Look at the following example:
* It's not clear that `instrumentationOptions` only affects the default instruments. Maybe we can deprecate the name, provide a second prop which is call something like `defaultInstrumentationOption` and read them both under the good.
* `instruments: []` and `additionalInstruments: () => []` are very similar. Why not export `getIgnoreURLs` or make the property `instrumentations[] | ({ignoreUrls}: {ignoreUrls: Patterns}) => Instrumentation[];`. If a function is provided then call it with the `ignoreUrls` a param. This would help us to get rid of an extra config option and helps.

I bet you have some good ideas as well so please don't hesitate and add them :) 

```ts
      new TracingInstrumentation({
        // Affect both, default and custom instruments
        contextManager: ...,
        propagator: ...,
        resourceAttributes: {},
        spanProcessor: ...,

        // Affects only default instruments
        // -- maybe deprecate and rename to `defaultInstrumentationOptions`?
        instrumentationOptions: {
          fetchInstrumentationOptions: {
            applyCustomAttributesOnSpan: () => {},
            ignoreNetworkEvents: true,
          },
          xhrInstrumentationOptions: {
            applyCustomAttributesOnSpan: () => {},
            ignoreNetworkEvents: true,
          },
        },

        // affects custom and default instruments
        // -- it's not directly clear what's the difference between `instrumentations` and `additionalInstrumentations`. When do I need to use what?
        // ---- get rid of it.
        additionalInstrumentations: (ignoreUrls) => [],

        // Overwrites default instruments allows to add more instruments via the `additionalInstrumentations` function
        // -- Why not mak instrumentation to be working with both:
        // --- Tou can provide ad array of instruments or use a function like additionalInstrumentations() to provide an array of instruments
        instrumentations: [], // instrumentations[] | instrumentations: (ignoreUrls) => instrumentations[],
      }),
```
@cedricziel, @kpelelis, @eskirk Would love to get your input as well.

Wdyt and again big thanks,
Marco

Bruno-DaSilva:
@codecapitano quick q - how do y'all feel about breaking changes? Do we need to only consider changes that are backwards compatible with existing behaviour?

codecapitano:
> @codecapitano quick q - how do y'all feel about breaking changes? Do we need to only consider changes that are backwards compatible with existing behaviour?

It depends a bit on the change and it's impact.
Usually we try to avoid breaking changes of course but on rare occasions we still had them.

For example in our experimental modules or if they inevitable, like when we introduced the new session manager or if the impact is super low but impact is high. All of this need discussion with the team first.

Introducing breaking changes is especially problematic for users using the CDN version because, in it's default config, it auto fetches minor and patch updates on page load. 

--- 

So by the end it needs to be done on a case by case decision.

pshahid:
@codecapitano I agree that the approach suggested creates some ambiguity about how to handle override vs additional instrumentation. If you already have ideas on how to move forward with changing the API then please proceed without this contribution, or feel free to rework it as you see fit. 

For me personally, I need to keep chugging on daily work to meet some monthly/quarterly goals, so I likely won't have enough time to collaborate on this change. 

If it's as simple as changing instrumentations from an array to an optional array or function I can do that in this PR, but that'd probably be as far as I go for now due to time constraints.

codecapitano:
@pshahid @Bruno-DaSilva 

I think the easiest solution for now would be to expose the ignoreURLs to make it easier to use them in additional otel instrumentations.

It's not the 100% best solution but I think it can already make the setup a bit easier.

Thoughts?

Bruno-DaSilva:
sounds reasonable to me! Minimizes change to the existing API for now.

codecapitano:
Hi @pshahid and @Bruno-DaSilva PR to export `ignoreUrls` is in review.

https://github.com/grafana/faro-web-sdk/pull/732

codecapitano:
@pshahid We released a new Faro version today which exports a `getIgnoreUrls` function.

```ts
import  { getIgnoreUrls } from '@grafana/faro-web-sdk';

```

Bruno-DaSilva:
sickkkk thank you @codecapitano!!

pshahid:
Amazing, thank you @codecapitano!!!

PR #763, opened 2024-12-27
fix: bump otel

Netail:
## Why

Bump OTEL packages

## What

Bump otel packages

- ~Set target to ES2016, else `.includes` wasn't supported yet~
- Add `node` to tsconfig types, for `global` type

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


codecapitano:
Hey @Netail thank you so much for the contribution.
Can give a quick update on what OTEL bug is fixed due to that update?

Netail:
> Hey @Netail thank you so much for the contribution. Can give a quick update on what OTEL bug is fixed due to that update?

Oops, not necessarily a bug in here actually. But in nextjs bundling seems to have an issue (https://github.com/vercel/next.js/issues/74331), where it always picks the latest version. The rest of our apps uses 0.57.0, while faro uses 0.53.0, between these versions a function got removed which is now throwing an error

codecapitano:
Aaaah got it @Netail thanks that's fine. 🙏

Netail:
Could have reverted our versions ofc, but the dependencies have to updated here at some point, so why not xd

tsconfig.base.cjs.json
```diff
@@ -2,6 +2,6 @@
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "module": "CommonJS",
-    "target": "ES5"
```
codecapitano, line 5: Any specific reason to change the target here?

Note:
We have users (at least back in the days) who need ES5 support.

codecapitano:
> Could have reverted our versions ofc, but the dependencies here had to updated at some point, so why not xd

It's already time to update OTEL deps so you had perfect timing :) 
Thanks a lot for that 🙏

Netail:
After some ci/cd fixes, it should be good xd

codecapitano:
@Netail I can't promise that we release a new version this week so it might take till Monday.

Netail:
No worries, thanks :)

PR #893, opened 2025-01-21
feat: provide dynamic service name overrides

codecapitano:
## Why

As a user of Frontend O11y and Application O11y in Grafana Cloud, i would like to override the service name
for a subset of telemetry as the user progresses through the UX.

Example:
User lands on the login page, the telemetry should be mapped to the login service
User progresses through the app, goes to checkout, the associated telemetry should be associated with the checkout service
User manages their newsletter preferences, associated telemetry should be tagged with the profile service name
The override should be persisted in the user session to allow maximum flexibility.

TODO:
- [ ] update docs

## What

Faro provides two APIs to dynamically set overrides.
Under the hood it's nothing else as adding an overrides object to the session meta.
The receiver (for Grafana it's cloud only atm) reads the contents of the overrides object and triggers respective actions.
Currently it only provides a service name override.

The overrides can be set via the `setView(view, overrides)` and `setSession(session, overrides)` APIs.
When calling set session without a session object (falsy or empty)m the the session manager creates a new session.

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/demo/src/client/faro/initialize.ts
+++ b/demo/src/client/faro/initialize.ts
@@ -21,20 +21,7 @@ export function initializeFaro(): Faro {
         captureConsole: true,
       }),
 
-      new TracingInstrumentation({
-        instrumentationOptions: {
-          fetchInstrumentationOptions: {
-            applyCustomAttributesOnSpan: () => {
-              console.log('fetchInstrumentationOptions: applyCustomAttributesOnSpan');
-            },
-          },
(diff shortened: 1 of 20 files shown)


eskirk (approved):

eskirk (approved):
looks great to me, no comments. let me know when the receiver PR is done.

Issue #901, opened 2025-01-24
Add option for ConsoleInstrumentation args serializer


eskirk:
## Description

The default method for capturing console errors with the ConsoleInstrumentation often times serializes objects with the default serializer, leaving us with `[object Object]`. Being able to provide our own serializer to the ConsoleInstrumentation would help with debugging console errors. 

## Proposed solution

Add a `errorSerializer` option to the ConsoleInstrumentation constructor options which is used to parse the args of console errors before pushing them to the endpoint.

codecapitano:
@eskirk can we close this?

PR #904, opened 2025-01-25
feat: properly serialize the properties of a console error

eskirk:
solves #901 

# why

when using the `ConsoleInstrumentation` and capturing console errors, the instrumentation will strip the details including error source and stack frames from the original error. often times the captured console errors will appear as
```
console.error: [object Object]
```
which provides no details as to the actual error. 

# what

this PR changes the console instrumentation to properly pass along the details of the error to the API as well as providing an option for users to provide their own error serializing function. 

```js
intializeFaro({
   // ...
   instrumentations: [new ConsoleInstrumentation()],
   consoleInstrumentation: {
      errorSerializer: (args) => {
         // my custom serializer logic here
      },
   }
})
```


packages/web-sdk/src/instrumentations/console/instrumentation.ts
```diff
@@ -25,7 +32,9 @@ export class ConsoleInstrumentation extends BaseInstrumentation {
         console[level] = (...args) => {
           try {
             if (level === LogLevel.ERROR && !this.options?.consoleErrorAsLog) {
-              this.api.pushError(new Error('console.error: ' + defaultLogArgsSerializer(args)));
+              const { value, type, stackFrames } = getDetailsFromErrorArgs(args);
```
codecapitano, line 35: Q: shouldn't we still provide an option to serialize object in error messages or is this solved with the current functionality?

eskirk: yes we should, I just wanted to break these features up into two PRs. I will reconcile these changes once the serializer is complete.

eskirk: I do not think we need a custom serializer anymore. by properly parsing the details of the console error I do not think we will see the same issue with `[object Object]` anymore. if that isnt the case we can revisit a custom serializer.

codecapitano: Console Instrumentation creates an error when console.error is called. So we do not get an error but args any[].

I think unless an JS error is logged by console.error the initial problem of [object Object] still exists when a standard object is logged.

Not sure if you would like to tackle this as well by a customer serializer? 

Only want to mention it. No need to change yet.

packages/core/src/api/exceptions/const.ts
```diff
@@ -1 +1,30 @@
...
+
+function tryParseJSONObject (jsonString: string) {
+  try {
+    const o = JSON.parse(jsonString);
+
+    // Handle non-exception-throwing cases:
+    // Neither JSON.parse(false) or JSON.parse(1234) throw errors, hence the type-checking,
+    // but... JSON.parse(null) returns null, and typeof null === "object",
+    // so we must check for that, too. Thankfully, null is falsey, so this suffices:
+    if (o && typeof o === "object") {
```
codecapitano, line 13: Faro provides a file with convenience functions for check like this. It's `is.ts`

In this case you can use the [isObject(object)](https://github.com/grafana/faro-web-sdk/blob/8bb966ceb1fb2c5ab131cd9a4f1ea8a08f90acd9/packages/core/src/utils/is.ts#L34) function.

codecapitano: Please don't forget to add a test 😬

packages/core/src/api/exceptions/const.ts
```diff
@@ -1 +1,30 @@
+import { stringifyExternalJson } from "../../utils";
+
 export const defaultExceptionType = 'Error';
+
+function tryParseJSONObject (jsonString: string) {
```
codecapitano, line 5: Please don't forget to add a test 😬

packages/core/src/api/exceptions/const.ts
```diff
@@ -1 +1,30 @@
+import { stringifyExternalJson } from "../../utils";
+
 export const defaultExceptionType = 'Error';
```
codecapitano, line 3: Maybe you can use the [LogLevel](https://github.com/grafana/faro-web-sdk/blob/8bb966ceb1fb2c5ab131cd9a4f1ea8a08f90acd9/packages/core/src/utils/logLevels.ts#L1) enum?

But it has a lower case string (`error`). Maybe this fit's as well?

packages/web-sdk/src/instrumentations/console/instrumentation.ts
```diff
@@ -32,9 +35,18 @@ export class ConsoleInstrumentation extends BaseInstrumentation {
         console[level] = (...args) => {
           try {
             if (level === LogLevel.ERROR && !this.options?.consoleErrorAsLog) {
-              const { value, type, stackFrames } = getDetailsFromErrorArgs(args);
+              const { value, type, stackFrames } = getDetailsFromConsoleErrorArgs(args, this.errorSerializer);
```
codecapitano, line 38: Please add a test for the different cases

packages/web-sdk/src/utils/errors.ts
```diff
@@ -1,14 +1,20 @@
...
+  value?: string;
+  type?: string;
+  stackFrames?: ExceptionStackFrame[];
+}
+
+export function getDetailsFromErrorArgs(args: [any?, ...any[]]): ErrorDetails {
   const [evt, source, lineno, colno, error] = args;
 
-  console.log('getDetailsFromErrorArgs', {evt, source, lineno, colno, error});
+  console.log('getDetailsFromErrorArgs', { evt, source, lineno, colno, error, args });
```
codecapitano, line 17: console.log

packages/web-sdk/tsconfig.spec.json
```diff
@@ -6,6 +6,6 @@
     "rootDir": "..",
     "tsBuildInfoFile": "../../.cache/tsc/webSdk.spec.tsbuildinfo"
   },
-  "include": ["./src"],
+  "include": ["./src", "../core/src/utils/json.test.ts", "../core/src/utils/json.ts"],
```
codecapitano, line 9: Why did we need to update the icludes?

packages/web-sdk/src/instrumentations/console/instrumentation.ts
codecapitano: Do we rally want to use a default serializer?
when big object are console.error logged this increase payloads a lot.

Readability in frontend-o11y is affected by this as well since we use teh value/error message as the title for teh error.
Let's ensure that the UX can handle these cae properly and keeping things readable.

Ate least we should how to change the serializer in case error logs become too big.

codecapitano: Maybe you can test calling console.error with a bigger object and see how it looks in the UI.

codecapitano (approved):

PR #913, opened 2025-01-27
chore: define supported Node engines 

codecapitano:
## Why

Define supported Node engines.

## What
* Define Node engines in package.json engines object
* Update Github action to  build/test against Node 18 as well. 
* Update Github action to  build/test against Node version ranges instead of a fixed version number
* Downgrade rimraf to V5 because starting [with V6 rimraf needs Node version >= 20](https://www.npmjs.com/package/rimraf). 
   - [This rimraf version has no known CVEs](https://security.snyk.io/package/npm/rimraf/5.0.10)

## Links

- [Node EOL table](https://endoflife.date/nodejs) 

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -2,6 +2,9 @@
   "name": "@grafana/faro-demo",
   "version": "1.12.3",
   "description": "Demo of Faro",
+  "engines": {
+    "node": ">=20.0.0"
+  },
   "license": "Apache-2.0",
   "author": "Grafana Labs",
   "homepage": "https://github.com/grafana/faro-web-sdk",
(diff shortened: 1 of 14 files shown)


demo/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@grafana/faro-demo",
   "version": "1.12.3",
   "description": "Demo of Faro",
+  "engines": {
+    "node": ">=20.0.0"
```
cedricziel, line 6:
```suggestion
    "node": ">=18.0.0"
```

cedricziel (changes requested):
as discussed, let's find out if we build with node 18 first, then bump to 20

cedricziel (approved):

PR #923, opened 2025-01-29
feat(page-meta): provide options to override automatically created pageId's

codecapitano:
## Why

With this PR we provide a convenient way to the user to change the page-meta.
This is mostly useful to give users and option to provide a custom template for pageID parsing.

If the page-meta contains an `id` attribute, the Faro receiver will use this ID instead.

### API

**Page tracking can be configured during initialisation via the Faro options**

```ts
pageTracking?: {
    page?: MetaPage;
    generatePageId?: (location: Location) => string;
  };
```

* The `page` property can be used to set the initial page-meta.
* The `generatePageId()` function is used to define a template parsers which consistently  updates the page id. It is to satisfy advanced use cases

** Explicitly set a the page via the `setPage(page: string)` function.

`setPage` is an overloaded function. If the `page` parameter is of type string, Faro updates the current active page meta with the given page id.

## What
* Provide a `setPage(meta)` function to set/override the page-meta and `getPage()` function to receiver it.
* Provide an option to define a pageID template parser which is called on every URL change. It receiver the browser Location object as a parameter and is expected to return a string.
* The `defaultMetas` object is removed and such not exported anymore because pageMeta creation changed. This can break if users used that export for some reasons. It was never meant to be used outside Faro so we decided to remove it anyways.

## Links

* [Documentation PR](https://github.com/grafana/website/pull/23763)

## Screenshots

Manually testing the pageId parser using a custom parser which attaches a random id to the pageId URL string.
Jest doesn't play well with manually changing the URLs. 
Location isn't updated on the code side 

https://github.com/user-attachments/assets/89e27b50-f7cb-4584-9691-1e67da09e8fe

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/demo/src/client/faro/initialize.ts
+++ b/demo/src/client/faro/initialize.ts
@@ -42,16 +42,6 @@ export function initializeFaro(): Faro {
       version: env.package.version,
       environment: env.mode.name,
     },
-    pageTracking: {
-      generatePageId(location) {
-        return location.pathname + ' ' + genShortID();
-      },
-      page: {
-        attributes: {
-          foo: 'bar',
-        },
-      },
-    },
   });
 
   faro.api.pushLog(['Faro was initialized']);


packages/web-sdk/src/metas/page/meta.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    expect(global.window.location.href).toBe(initialUrl);
+
+    // @ts-expect-error
+    const meta = createPageMeta((location) => location.href + '_' + faroModule.genShortID())();
+    const pageMeta = meta.page;
+
+    expect(pageMeta.url).toBe(initialUrl);
+    expect(pageMeta.id).toBe(initialUrl + '_' + initialPagePostfix);
+
+    // TODO: It seems that jest doesn't work well with JS proxies. At leas the proxied location object doesn't update.
```
codecapitano, line 21: It seems that somehow in Jest the JS proxy won't trigger.

Tested the functionality manually with Firefox, Safari and Chrome and it works as expected.



https://github.com/user-attachments/assets/66b5cb9f-862f-4bca-8b93-2d96a07941e0

packages/web-sdk/src/metas/page/meta.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    // Logic works well when manually tested in the browser. See attached video in the PR.
+
+    // const newPagePostfix = 'new-page-postfix';
+    // jest.spyOn(faroModule, 'genShortID').mockReturnValueOnce(newPagePostfix);
+
+    // const newUrl = 'http://localhost/new-page/';
+    // global.window.history.pushState({}, '', new URL(newUrl));
+
+    // expect(pageMeta.url).toBe(newUrl);
+    // expect(meta.page.id).toBe(newUrl + '_' + newPagePostfix);
```
eskirk, line 44: I went through a similar issue with the frontend o11y plugin and found that I could make this work by doing the following: https://stackoverflow.com/a/54021633/4607958

not sure if that is applicable to this case but worth a shot

eskirk (approved):
lgtm, one comment about a potential solution to your testing issue

cedricziel (approved):

cedricziel:
The desired API usage would be:

Ad hoc usage:

```ts
faro.api.setPage('product-list');
```


Init usage:

```ts
initializeFaro({
  page: {
    id: 'product-list'
  }
});
```

packages/web-sdk/src/metas/page/meta.test.ts
codecapitano: Thanks Elliot. 
I changed the tests and could use this pattern well in the new tests 👍

demo/src/client/faro/initialize.ts
```diff
@@ -41,6 +42,16 @@ export function initializeFaro(): Faro {
       version: env.package.version,
       environment: env.mode.name,
     },
+    pageTracking: {
+      generatePageId(location) {
+        return location.pathname + ' ' + genShortID();
```
cedricziel, line 47: this seems to be a really bad example 😓 - can we come up with something that doesnt result in super-high cardinality?

codecapitano: It's a leftover form manually testing.
Let me remove it from the demo.

It's not meant to be an example

cedricziel (approved):

PR #933, opened 2025-02-03
fix(setPage): use meta.page if locally cached page meta is not set.

codecapitano:
## Why

Calling setPage with the a string or an object which is missing the URL, causes an auto merge with teh currently active meta. 

This is done by merging the new pageID property or string wih the current atcive meta from the cache.

When `setPage()` is called the first time, the locally cached meta is not set thus we need to fetch the active meta with `getPage()`.

This behaviour happens only after a page is loaded. 

## What

* Fetch active page meta if locally chached meta is not set. 
* Update Faro test config

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated


rrmf (approved):
🚀

Issue #947, opened 2025-02-07
Provide configuration to enable the "reportAllChanges" property in the web-vitals package


eskirk:
## Description

the [web-vitals package has a property called "reportAllChanges"](https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#report-the-value-on-every-change). providing users with the option to enable this feature would provide more data if desired.

## Proposed solution

provide an option in the `WebVitalsInstrumentation` to enable `reportAllChanges` in the `web-vitals` package

Aha! Link: https://grafana-labs.aha.io/features/FEO11Y-143

codecapitano:
Duplicate of https://github.com/grafana/faro-web-sdk/issues/485.

I think we can close or merge this one @eskirk

eskirk:
@codecapitano this is slightly different - this doesn't seem to involve soft navigations, this just makes it so web vitals are reported at a greater frequency which was a request one of our users has made.

codecapitano:
Argh yes didn't read carefully and confused it with `{reportSoftNavs: true}` sorry.

PR #981, opened 2025-02-18
feat: adding ability to track all changes for web vitals

eskirk:
## Why

solves #947 

provides the option to enable the `trackAllChanges` feature in the web-vitals package 

## What

add `reportAllWebVitalChanges` to the config object to enable that feature in the web-vitals package

### DIFF
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -164,20 +164,28 @@ export interface Config<P = APIEvent> {
   trackWebVitalsAttribution?: boolean;
 
   /**
-   * Report all changes for web vitals (default: false)
-   *
-   * In most cases, you only want the callback function to be called when the metric is ready to be reported.
-   * However, it is possible to report every change (e.g. each larger layout shift as it happens)
-   * by setting reportAllChanges to true.
+   * Configuration for the web vitals instrumentation
    */
-  reportAllWebVitalChanges?: boolean;
+  webVitals?: {
+    /**
+     * Report all changes for web vitals (default: false)
+     *
+     * In most cases, you only want the callback function to be called when the metric is ready to be reported.
+     * However, it is possible to report every change (e.g. each larger layout shift as it happens)
+     * by setting reportAllChanges to true.
+     *
+     * This can be useful when debugging, but in general using reportAllWebVitalChanges is not needed (or recommended)
+     * for measuring these metrics in production.
+     */
+    reportAllWebVitalChanges?: boolean;
+  };
 
   /**
    * Configuration for the console instrumentation
    */
   consoleInstrumentation?: {
     /**
-     * Configure what console levels should be captured by Faro. By default the follwoing levels
+     * Configure what console levels should be captured by Faro. By default the following levels
      * are disabled: console.debug, console.trace, console.log
      *
      * If you want to collect all levels set captureConsoleDisabledLevels: [];
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -166,7 +166,7 @@ export interface Config<P = APIEvent> {
   /**
    * Configuration for the web vitals instrumentation
    */
-  webVitals?: {
+  webVitalsInstrumentation?: {
     /**
      * Report all changes for web vitals (default: false)
      *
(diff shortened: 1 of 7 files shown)


packages/web-sdk/src/instrumentations/webVitals/webVitalsWithAttribution.ts
```diff
@@ -28,114 +31,132 @@ export class WebVitalsWithAttribution {
...
+        this.addIfPresent(values, 'largest_shift_value', largestShiftValue);
+        this.addIfPresent(values, 'largest_shift_time', largestShiftTime);
+
+        const context = this.buildInitialContext(metric);
+        this.addIfPresent(context, loadStateKey, loadState);
+        this.addIfPresent(context, 'largest_shift_target', largestShiftTarget);
+
+        this.pushMeasurement(values, context);
+      },
+      { reportAllChanges: this.reportAllChanges }
```
eskirk, line 48: a lot of changes in these files, but all that I really added was `{ reportAllChanges: this.reportAllChanges }`

packages/core/src/config/types.ts
```diff
@@ -163,6 +163,15 @@ export interface Config<P = APIEvent> {
    */
   trackWebVitalsAttribution?: boolean;
 
+  /**
+   * Report all changes for web vitals (default: false)
+   *
+   * In most cases, you only want the callback function to be called when the metric is ready to be reported.
+   * However, it is possible to report every change (e.g. each larger layout shift as it happens)
```
codecapitano, line 170: Let's add a comment that `reportAllChanges` isn't recommended to use in production and more for debugging purposes. See [web-vitals docs](https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#report-the-value-on-every-change)

codecapitano: Suggestion: 
since we now have two props to configure web vitals we can add a config object bundling web-vitals updates. We can still keep the `trackWebVitalsAttribution` for quick enabling/disabling.
Then we would follow the shape of other config object which have multiple params.


Exmple:

```ts
webVitals {
  enabled: boolean
  reportAllWebVitalChanges: boolean
}
```

packages/web-sdk/src/instrumentations/webVitals/instrumentation.ts
```diff
@@ -15,8 +15,8 @@ export class WebVitalsInstrumentation extends BaseInstrumentation {
 
   private intializeWebVitalsInstrumentation() {
     if (this.config.trackWebVitalsAttribution) {
-      return new WebVitalsWithAttribution(this.api.pushMeasurement);
+      return new WebVitalsWithAttribution(this.api.pushMeasurement, this.config.reportAllWebVitalChanges);
```
codecapitano, line 18: You need to update [makeCoreConfig](https://github.com/grafana/faro-web-sdk/blob/f910c9586cf5f9ec5c7056a2b2fa842f1b03fc9d/packages/web-sdk/src/config/makeCoreConfig.ts#L24) so the new property is added to the core config

Please add a small test to ensure that the `trackWebVitalsAttribution` is available in the config

codecapitano:
Please also update the [web-vitals cloud docs](https://github.com/grafana/website/blob/007e7633f198fac41c6e20c30b2254b987a65ba7/content/docs/grafana-cloud/monitor-applications/frontend-observability/instrument/web-vitals.md#L16-L17).

codecapitano:
please update the changelog

packages/core/src/config/types.ts
eskirk: I think I understood what you wanted here - I did not want to create a breaking change by adding `trackWebVitalsAttribution` into this object, but wasnt too sure what else to add in here?
image

packages/web-sdk/src/config/makeCoreConfig.test.ts
```diff
@@ -83,6 +83,22 @@ describe('config', () => {
...
+        reportAllWebVitalChanges: true,
+      },
+    };
+    const config = makeCoreConfig(browserConfig);
+
+    expect(config).toBeTruthy();
+    expect(config?.trackWebVitalsAttribution).toBe(true);
+    expect(config?.webVitals?.reportAllWebVitalChanges).toBe(true);
+  });
+
```
eskirk, line 101: @codecapitano is this what you had in mind?

eskirk:
made changes to cloud docs [here](https://github.com/grafana/website/pull/24122)

packages/core/src/config/types.ts
codecapitano: Yes exactly. 

Nit:
for consistency we can add an enabled attribute to the `webVitals`object as well

codecapitano: One last thing:

Let's add the "Instrumentation" prefix to the config object for consistency with the other instrumentation options like the [consoleInstrumentation](https://github.com/grafana/faro-web-sdk/blob/43be0f525d3abc28110a2a22922221a121fa7a4f/packages/core/src/config/types.ts#L169) .

And remove the `webVitals` part from the nested property bc it is already namespaced by the parent object.


```ts
webVitalsInstrumentation: {
  reportAllChanges: true
}
```

codecapitano (approved):
lgtm but please adjust the property names for consistency before merging 🙏

PR #990, opened 2025-02-20
chore(web-tracing): attach user attributes to spans

codecapitano:
## Why

The web-tracing instrumentation now takes care to add user attributes to spans.

The `UserMeta` was updated with more properties to be aligned with [otel semantic conventions for user](https://opentelemetry.io/docs/specs/semconv/attributes-registry/user/).

## What
* Add missing attributes to user meta to be compatible with otel semantic attributes
* create new span exporter which takes care of adding session and user attributes to spans

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/core/src/metas/types.ts
+++ b/packages/core/src/metas/types.ts
@@ -35,9 +35,33 @@ export interface MetaApp {
 }
 
 export interface MetaUser {
+  /**
+   * User email address.
+   */
   email?: string;
+  /**
+   * Unique identifier
+   */
   id?: string;
+  /**
+   * User’s full name
+   */
   username?: string;
+  /**
+   * Short name or login/username of the user.
+   */
+  shortName?: string;
+  /**
+   * comma separated list of user roles. "admin",editor" etc.
+   */
+  roles?: string;
+  /**
+   * Unique user hash to correlate information for a user in anonymized form.
+   */
+  hash?: string;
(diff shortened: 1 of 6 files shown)


packages/core/src/metas/types.ts
```diff
@@ -35,9 +35,33 @@ export interface MetaApp {
...
+   */
   id?: string;
+  /**
+   * User’s full name
+   */
   username?: string;
+  /**
+   * Short name or login/username of the user.
+   */
+  shortName?: string;
```
rrmf, line 53: This is a bit confusing for me, to have a `shortName` that means what is broadly know as `username`. I would suggest something different like: `fullName` that maps to the user full name and `username` or `shortName` that maps to the actual login/username. 
wdyt?

codecapitano: Yes good point.

The otel semantic attribute for the short name is `[user.name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/user/#user-name)` and the attribute for fullName i[user.full_name](https://opentelemetry.io/docs/specs/semconv/attributes-registry/user/#user-full-name).

Long story short. Your suggestion makes total sense. 🙏 💯

rrmf (approved):
LGTM 🚀

PR #1000, opened 2025-02-25
Search-all-exception-properties-when-matching-error-excludes

codecapitano:
## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/docs/sources/developer/architecture/components/transports.md
+++ b/docs/sources/developer/architecture/components/transports.md
@@ -15,7 +15,6 @@ Methods and properties:
 
 - `add()` - adds a new transport
 - `addBeforeSendHook()` - adds a hook that is called before a signal is sent to each transport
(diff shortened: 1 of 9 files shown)


docs/sources/developer/architecture/components/transports.md
```diff
@@ -15,7 +15,6 @@ Methods and properties:
 
 - `add()` - adds a new transport
 - `addBeforeSendHook()` - adds a hook that is called before a signal is sent to each transport
-- `addIgnoreErrorsPatterns()` - adds a pattern that matches errors which should be ignored
```
cedricziel, line 18: is this a cleanup or a BC break?

cedricziel (approved):

docs/sources/developer/architecture/components/transports.md
codecapitano: `addIgnoreErrorsPatterns` is removed from Faro

PR #1033, opened 2025-03-13
feat(PoC): track user actions 

codecapitano:
# Why

This feature implement a way to track "User Actions".
A user action is basically a unit of work done as the result of a user action.
They can help providing insights into users behaviour and give other important insights like:
* the overall performance of a user action (action duration)
* related errors
* resources loaded by the action
* and more

### User Action Instrumentation

Listens to `keydown` and `pointerdown` events and tracks actions if the following  point are fulfilled:

* A focusable element need to be marked as a user action by adding the `data-faro-user-action-name` to the element.
     Faro will use the attribute value as the name (`data-faro-user-action-name="my-action"`).  
   
 * If one of the above events is triggered and the element emitting the event is marked as a user event, one of the 
   follow up actions need to happen within 100ms: a HTTP request or a DOM mutation or a Resource load. 
   Note 1: We aligned theses settings with what other vendors have specified as well to be compliant with the industry 
   Note 2: By default Faro only tracks resource timings for http requests and no others (can be changed by config). This is 
   not unnecessarily increase the volume of data sent and stored. We'll update this in the future so we can safely track 
   more timings by default.

NOTE: PR description is still WIP
 

# What

# Links

# Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated


packages/core/src/api/initialize.ts
```diff
@@ -21,14 +25,27 @@ export function initializeAPI(
...
-    ...initializeExceptionsAPI(unpatchedConsole, internalLogger, config, metas, transports, tracesApi),
-    ...initializeMetaAPI(unpatchedConsole, internalLogger, config, metas, transports),
-    ...initializeLogsAPI(unpatchedConsole, internalLogger, config, metas, transports, tracesApi),
-    ...initializeMeasurementsAPI(unpatchedConsole, internalLogger, config, metas, transports, tracesApi),
-    ...initializeEventsAPI(unpatchedConsole, internalLogger, config, metas, transports, tracesApi),
+    ...initializeExceptionsAPI(props),
+    ...initializeMetaAPI(props),
+    ...initializeLogsAPI(props),
+    ...initializeMeasurementsAPI(props),
+    ...initializeEventsAPI(props),
```
eskirk, line 49: good change 👌🏼

packages/core/src/api/logs/initialize.test.ts
```diff
@@ -167,4 +171,54 @@ describe('api.logs', () => {
       });
     });
   });
+
+  describe('User action', () => {
+    it('buffers the error if a user action is in progress', () => {
```
eskirk, line 176: what error is being buffered here? 

should this say:
```suggestion
    it('buffers the item if a user action is in progress', () => {
```

packages/core/src/api/measurements/initialize.test.ts
```diff
@@ -236,4 +247,54 @@ describe('api.measurements', () => {
       });
     });
   });
+
+  describe('User action', () => {
+    it('buffers the error if a user action is in progress', () => {
```
eskirk, line 252: same here, should this just say 

```suggestion
    it('buffers the item if a user action is in progress', () => {
```
?

packages/core/src/config/types.ts
```diff
@@ -238,6 +238,16 @@ export interface Config<P = APIEvent> {
...
+
+  /**
+   * This is a preview feature.
+   * We have tested it thoroughly, but it is possible that it might not work as expected in all cases.
+   */
+  trackUserActions?: boolean;
+  /**
+   * Configure your own attribute name for tracking user actions. Default is 'data-faro-user-action-name'
+   */
+  trackUserActionsDataAttributeName?: string;
```
eskirk, line 250: just adding a new line for style 
```suggestion

  /**
   * Configure your own attribute name for tracking user actions. Default is 'data-faro-user-action-name'
   */
  trackUserActionsDataAttributeName?: string;
```

packages/web-sdk/src/instrumentations/userActions/processUserActionEventHandler.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      )
+    );
+  });
+
+  afterAll(() => {
+    jest.restoreAllMocks();
+    jest.clearAllTimers();
+  });
+
+  it('Emits a user-action-end message if a user action has follow upa activity within 100ms', () => {
```
eskirk, line 37:
```suggestion
  it('Emits a user-action-end message if a user action has follow up activity within 100ms', () => {
```

packages/web-sdk/src/instrumentations/userActions/processUserActionEventHandler.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (key === parsedDataAttributeName) {
+      return dataset[key];
+    }
+  }
+
+  return undefined;
+}
+
+function startTimeout(timeoutId: number | undefined, cb: () => void) {
+  const maxTimeSpanTillUserActionEnd = 100;
```
eskirk, line 129: lets consider making this configurable in the future?

eskirk (commented):
few comments and questions but otherwise you killed it 🔥

packages/core/src/api/events/initialize.ts
```diff
@@ -59,7 +71,12 @@ export function initializeEventsAPI(
 
       internalLogger.debug('Pushing event\n', item);
 
-      transports.execute(item);
+      const msg = getMessage();
+      if (msg && msg.type === 'user-action-start') {
```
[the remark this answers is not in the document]
codecapitano: Oh damns yes, I added the type but no constants

packages/web-sdk/src/instrumentations/userActions/processUserActionEventHandler.ts
codecapitano: Let's keep an eye on it.
When we go down this route we need more option to exclude unrelated signals.
The longer the timeframe the more events unrelated to the action will occur.

But we need to add that functionality anyways.I prepared an option for to exclude items form actions.
It's in the UserActionLifecyleHandler 
https://github.com/grafana/faro-web-sdk/blob/32e52aec41fe8aca863cfcbb5db6f333886b8d67/packages/core/src/api/userActionLifecycleHandler.ts#L63

cedricziel (approved):

rrmf (approved):
🚀

PR #1057, opened 2025-03-19
fix(quick-start-browser docs): update default otel instruments

codecapitano:
## Why

rom quick-start-browser docs: Remove instruments fwe do longer include by default 

## What

See above

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [x] Documentation updated


rrmf (approved):
LGTM 🚀

Issue #1123, opened 2025-04-11
Errors auto instrumentation: Read error context from errors captured


codecapitano:
# Why

sometimes customers add extra context information to am error which bubbles to a global handler.
In some cases the global handler bubbles as well to ensure to stop execution.
In such cases Faro's auto error instrumentation captures the error and there's no option to inject context data.

# What
- Update Faro errors instrumentation so it's able to pick up context data from a defined property and adds it to the Faro exception event
- For convenience provide a TS interface / JS class to extend to make it easier to create errors

Aha! Link: https://grafana-labs.aha.io/features/FEO11Y-139

PR #1133, opened 2025-04-15
chore(error tracking): provide original error to beforeSend hook

codecapitano:
## Why

Provide an option to instruct Faro to provide the original error in a Faro error payload.
This is useful to handle special cases in uncontrolled environments/libraries or other enhanced use cases.
I allows to enrich the Faro error before sending it to the collector.

Examples:
* `throw undefined`
* `throw ''`
* Enriching errors while they bubble up to a global error handler which rethrows itself. 
*...

To not unnecessary increase resource user for users who don't use the feature, the option needs to enabled via the `preserveOriginalError: true` property.

If enabled, Faro adds a `orginalError` property to the Faro exception payload holding a ref to the original JS Error.

The `orginalError` is not handled by the Faro receiver so it is removed before sending the data to not unnecessarily increase request payload sizes.

Faro uses the `ExceptionEventExtended` for in that case.

Example
```
beforeSend(item) {
      if (item.type === 'exception') {
        const originalError = (item as TransportItem<ExceptionEventExtended>).payload.originalError;
        // ...
      }

      return item;
    },
```

## Links
[Issue](https://github.com/grafana/faro-web-sdk/issues/1123)
[Docs PR (internal)](https://github.com/grafana/website/pull/25256)

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/packages/core/src/api/exceptions/types.ts
+++ b/packages/core/src/api/exceptions/types.ts
@@ -26,7 +26,7 @@ export interface Stacktrace {
 
 export type ExceptionContext = Record<string, string>;
 
-export interface ExceptionEvent {
+export interface ExceptionEventDefault {
   timestamp: string;
   type: string;
   value: string;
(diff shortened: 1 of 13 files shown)


packages/core/src/api/exceptions/types.ts
```diff
@@ -38,6 +38,19 @@ export interface ExceptionEvent {
   action?: UserAction;
 }
 
+/**
+ * The ExceptionEventExtended type is used to represent an exception event with an additional error
+ * property ans is only meant for client side use.The Additional property is removed by Faro before
```
eskirk, line 43:
```suggestion
 * property and is only meant for client side use. The additional property is removed by Faro before
```

packages/core/src/transports/initialize.ts
```diff
@@ -189,3 +190,15 @@ export function initializeTransports(
     unpause,
   };
 }
+
```
eskirk, line 193:
```suggestion
  /**
   * Removes the `payload.originalError` property from the provided `TransportItem[]` parameter. 
   */ 
```

eskirk (approved):
couple things mostly surrounding the comments and docs, but looks good

Issue #1160, opened 2025-04-25
ignoreErrors not working properly in some cases


edvard-falkskar:
## Description
In our application we have configured Faro to ignore `chrome-extension` by setting `ignoreErrors: ['chrome-extension']`, but from time to time we see these chrome extension errors in our list of errors anyway.

[Example](https://appo11y.grafana.net/a/grafana-kowalski-app/errors?frameIndex=0&from=1745410801590&to=1745583601590&var-app_id=78&timezone=browser&var-loki=grafanacloud-logs&var-error_app_id=78&var-error_text=FallbackRequested&var-error_hash=16436379023565278787&var-show_errors_drawer=1&var-error_session_id=0wDmp5rw4K)
![Image](https://github.com/user-attachments/assets/963059d1-d1d1-4335-8927-8e71c74da650)

After spending too much time investigating this I think I know one way these errors get past the ignore list. But there might be more since in the example I see zone.js and I’m not sure it takes the route I will show in the next section.

## Steps to reproduce
Since I don't have a good chrome extension to create errors with I'm faking it instead.

First thing is to make sure you have `ignoreErrors: ['chrome-extension']` set.

Second thing is to throw a error that will be caught by faro, I first tried from the console but that does not seem to get picked up by faro. My solution is to add a event handler to an element from the developer tools and then from there throw a fake error. Select an appropriate element in the Elements tab and run this code in the console.

```
$0.onmouseover = () => {
  const mockErrorWithStacktrace = new Error('Mock error for testing');
  mockErrorWithStacktrace.name = 'MockError';
  mockErrorWithStacktrace.stack = `MockError: Mock error for testing
at mockFunction (chrome-extension://mock-extension-id/mock-file.js:10:15)
at anotherFunction (chrome-extension://mock-extension-id/mock-file.js:20:5)
at Object.<anonymous> (chrome-extension://mock-extension-id/mock-file.js:30:3)`;

  throw mockErrorWithStacktrace;
};
```

Now add breakpoints at these locations:
* `instrumentations/errors/registerOnerror.js` - line 8 - `api.pushError`
* `api/exceptions/initialize.js` - line 20 - `if (isErrorIgnored(ignoreErrors, error))`

Now move your mouse over the element you attached the `onmouseover` event to.

In `registerOnerror.js` the original error will look something like this

![Image](https://github.com/user-attachments/assets/d8515e96-e3ff-44bb-974b-4bf162d33020)

[...]

PR #1168, opened 2025-04-29
fix(ignoreErrors): ensure to match on the original error and not on the parsed error

codecapitano:
## Why

Ignoring errors didn't work properly when matching stack only strings.
This was because the parsed error was used which contains parsed stack frames and not the original stack string anymore. 

## What
Open up the options object of the `pushError` API to inject the original error as well.
The original error is then used to check if the error has a match in the `ignoreErrors` list.

Additionally we no (correctly) assign original error to the  `originalError` object in the exception signal payload.
This is used if [preserveOriginalError](https://grafana.com/docs/grafana-cloud/monitor-applications/frontend-observability/instrument/error-tracking/#handle-other-edge-cases) is enabled.

## Links
Fixes: #1160 

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+- Fix (`@grafana/faro-web-sdk`): Fixed the ignored errors matching logic to properly identify
+  errors based on stack trace content (#1168).
+
 ## 1.17.2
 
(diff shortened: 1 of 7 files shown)


edvard-falkskar (approved):
LGTM 👍

PR #1183, opened 2025-05-05
chore(web-tracing): add resource attrs. for browser and distribution

codecapitano:
## Why

Add resource attributes for runtime and distro yo make it more staright forward to query respective spans.

## What

Add above attr. 

Note:
They are not stable and the OTEL suggested way is to create own files to hold the attributes to prevent breakage.

## Screenshots

Screenshot 2025-05-05 at 17 35 32

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-tracing/src/instrumentation.ts
+++ b/packages/web-tracing/src/instrumentation.ts
@@ -7,10 +7,11 @@ import { BatchSpanProcessor, WebTracerProvider } from '@opentelemetry/sdk-trace-
 import {
   ATTR_SERVICE_NAME,
   ATTR_SERVICE_VERSION,
+  ATTR_USER_AGENT_ORIGINAL,
   SEMRESATTRS_DEPLOYMENT_ENVIRONMENT,
 } from '@opentelemetry/semantic-conventions';
 
-import { BaseInstrumentation, unknownString, VERSION } from '@grafana/faro-web-sdk';
+import { BaseInstrumentation, isArray, VERSION } from '@grafana/faro-web-sdk';
 import type { Transport } from '@grafana/faro-web-sdk';
 
 import { FaroMetaAttributesSpanProcessor } from './faroMetaAttributesSpanProcessor';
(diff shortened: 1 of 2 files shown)


packages/web-tracing/src/instrumentation.ts
```diff
@@ -59,6 +71,34 @@ export class TracingInstrumentation extends BaseInstrumentation {
...
+      attributes[ATTR_BROWSER_MOBILE] = Boolean(browserMeta.mobile);
+    }
+
+    if (browserMeta?.os != null) {
+      attributes[ATTR_BROWSER_PLATFORM] = browserMeta.os;
+    }
+
+    if (browserMeta?.userAgent) {
+      attributes[ATTR_USER_AGENT_ORIGINAL] = browserMeta.userAgent;
+    }
```
eskirk, line 94: do the `mobile` and `os` properties have a default value that require you to explicitly check for `null`?  

wondering why only some of these ifs check for null

packages/web-tracing/src/instrumentation.ts
```diff
@@ -59,6 +71,34 @@ export class TracingInstrumentation extends BaseInstrumentation {
...
 
+    const browserMeta = this.metas.value.browser;
+
+    if (isArray(browserMeta?.brands)) {
+      attributes[ATTR_BROWSER_BRANDS] = browserMeta.brands.map((entry) => entry.brand);
+    }
+
+    if (browserMeta?.userAgent) {
+      attributes[ATTR_BROWSER_LANGUAGE] = browserMeta.language;
+    }
```
eskirk, line 82:
```suggestion
    if (browserMeta?.language) {
      attributes[ATTR_BROWSER_LANGUAGE] = browserMeta.language;
    }
```

eskirk (changes requested):
just one question and one change request otherwise looks good

packages/web-tracing/src/instrumentation.ts
codecapitano: Good catch Elliot.

browserMeta?.os is a copy paste thingy


browserMeta?.mobile is a boolean so i check for undefined and null (== null does check for both).
But you're right. Let me be explicit here and check for boolean.

packages/web-tracing/src/instrumentation.ts
codecapitano: Oh phew thanks for finding this!

eskirk (approved):

PR #1214, opened 2025-05-12
chore: add tests for stack trace generation

eskirk:
## Why

opening PR for an orphaned branch adding more tests

## What

more tests for stack trace functionality

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


codecapitano (approved):
LGTM. 
Let's double check what tests we may want to remove. 
For example Firefox 3 which was released in 2008

PR #1249, opened 2025-05-19
chore(user actions): don't attach user action context to http request when in halt mode

codecapitano:
## Why

When a action has long pending http requests it starts an extended timer if the regular action timeout is reached while waiting for the pending requests to resolve and the action goes in `halt` state .

The `halt` state indicates that the regular action has been finished bc no further eligible event happened within the respective timeframes and it waits for the pending requests to finish.

This can already happen on slow connection (can be tested with throttling in the browser).

Because we forgot to respect the `halt` state in the web tracing package it could happen that any requests sent during the `halt` phase (remember they don't belong to the action anymore) were still mapped to the action.

Note:
As long as the action is in `halt` state no new action can be started. 

## What

* add the missing `halt` state to the [FaroUserActionSpanProcessor](https://github.com/grafana/faro-web-sdk/blob/447907eb87781a86c68b53e1ad0a43de0e681cb1/packages/web-tracing/src/faroUserActionSpanProcessor.ts#L23)

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/index.ts
+++ b/packages/web-sdk/src/index.ts
@@ -169,7 +169,7 @@ export type {
   UserActionStartMessage,
   UserActionEndMessage,
   UserActionCancelMessage,
-  UserAction,
+  UserActionHaltMessage,
 } from '@grafana/faro-core';
 
 export {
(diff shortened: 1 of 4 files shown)


kpelelis (approved):
LGTM 👍

PR #1259, opened 2025-05-20
chore(deps): remove node 18 support, add node 24 support

codecapitano:
## Why

Remove the node engine object and update the supported node versions in the test matrix. 

## What

See above

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


eskirk (approved):
this feels like the right way to handle things 🤝

Issue #759, opened 2024-12-09
React v19 support


mathisfo:
## Description
@grafana/faro-react has a conflicting peer dependency in our project starting with react v19 

```json
"peerDependencies": {
    "react": "^16.8.0 || ^17.0.0 || ^18.0.0",
    "react-dom": "^16.8.0 || ^17.0.0 || ^18.0.0",
    "react-router-dom": "^4.0.0 || ^5.0.0 || ^6.0.0"
  },
```

Is there any plans to support react 19 in the near future?

Aha! Link: https://grafana-labs.aha.io/features/FEO11Y-145

codecapitano:
Hey @mathisfo yes we'll support React 19.
I can't 100% tell you when exactly we'll start but I expect it to be this quarter.

Cheers

hbj:
Hi @codecapitano, any updates on the timeline? Thanks!

codecapitano:
Hi @hbj It hasn't made it to our roadmap and I can provide a timeline unfortunately.

noelparisi:
Any updates on this thread? Thank you!

dlo:
Hi team, React 19 has been the latest stable release for over 6 months. I'd love to try Grafana Faro but this is a hard blocker.

BrendanBuono:
@codecapitano, as mentioned above, it's a peer dependency issue with the faro-web-sdk. I'm not sure if that's the responsibility of the maintainers of this repository or the faro-react repository. Still, it will require changes to both, and as mentioned, React 19 has been out a while and is continuing to gain adoption. It would be really great to be able to use the entirety of the faro ecosystem inside these projects

ValentinGurkov:
The peer dependencies on the React package have been updated to list React 19, but I have not done any tests with it.

PR #1293, opened 2025-06-03
chore(deps): add react 19 to peer deps

AlexanderYW:
## Why
Need react 19 added to peer deps to be able to install package in my react 19 application. Manipulating package-lock file manually solves the issue, but i consider that a short term solution, since it would require manual update on every release.

## What
* Updated the `react` peer dependency to include version `^19.0.0`.
* Updated the `react-dom` peer dependency to include version `^19.0.0`.

This PR does not include updates to react-router-dom.

## Links
https://github.com/grafana/faro-web-sdk/issues/759

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


codecapitano (approved):
Thank's a lot @AlexanderYW for the compatibility update. 🎉 

We'll update the README and docs for more clarity on the React Version support situation so people are aware that real R19 support is still missing atm.

PR #1312, opened 2025-06-11
feat(web-sdk): add Content Security Policy Instrumentation

kpelelis:
## Why

In order to support Content Security Policy violation payloads

## What

Create a new CSPInstrumenetation that registers to the `securitypolicyviolation` event and forwards the payload as a Faro event. Although this is not pedantic in terms of using the `Report-To` header etc, it is a way of getting this event into the pipeline and utilising all the metadata Faro SDK provides.

## Links

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP

## Checklist

- [X] Tests added
- [X] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/core/src/instrumentations/initialize.ts
+++ b/packages/core/src/instrumentations/initialize.ts
@@ -65,7 +65,7 @@ export function initializeInstrumentations(
         null
       );
 
-      if (!existingInstrumentationIndex) {
+      if (existingInstrumentationIndex === null) {
         internalLogger.warn(`Instrumentation "${instrumentationToRemove.name}" is not added`);
 
         return;
--- a/packages/web-sdk/src/instrumentations/csp/instrumentation.test.ts
+++ b/packages/web-sdk/src/instrumentations/csp/instrumentation.test.ts
@@ -1,0 +1,71 @@
+import { initializeFaro, TransportItem } from '@grafana/faro-core';
+import type { EventEvent } from '@grafana/faro-core';
+import { mockConfig, MockTransport } from '@grafana/faro-core/src/testUtils';
+
+import { makeCoreConfig } from '../../config';
+
+import { CSPInstrumentation } from './instrumentation';
+
+describe('CSPInstrumentation', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+    jest.restoreAllMocks();
+  });
+
+  it('sends a faro event when securitypolicyviolation event is fired', () => {
+    const mockTransport = new MockTransport();
+
+    const instrumentation = new CSPInstrumentation();
+    initializeFaro(
+      makeCoreConfig(
+        mockConfig({
+          transports: [mockTransport],
+          instrumentations: [instrumentation],
+        })
+      )!
+    );
+
+    const fakeEvent = {
+      blockedURI: 'https://evil.com/script.js',
+      documentURI: 'https://my.app/',
+      sourceFile: 'https://my.app/index.js',
+      statusCode: 200,
+      lineNumber: 10,
+      columnNumber: 5,
+      disposition: 'enforce',
+      effectiveDirective: 'script-src',
+      violatedDirective: 'script-src-elem',
+      originalPolicy: "default-src 'self'; script-src 'self'",
+      referrer: 'https://referrer.app/',
+      sample: 'alert("xss")',
+    } as SecurityPolicyViolationEvent;
+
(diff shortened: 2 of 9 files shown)


packages/core/src/instrumentations/initialize.ts
```diff
@@ -65,7 +65,7 @@ export function initializeInstrumentations(
         null
       );
 
-      if (!existingInstrumentationIndex) {
+      if (existingInstrumentationIndex === null) {
         internalLogger.warn(`Instrumentation "${instrumentationToRemove.name}" is not added`);
```
codecapitano, line 70: Good catch!

packages/web-sdk/src/instrumentations/csp/instrumentation.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  constructor() {
+    super();
+    this.boundHandler = this.securitypolicyviolationHandler.bind(this);
+    document.addEventListener('securitypolicyviolation', this.boundHandler);
+  }
+
+  initialize() {}
+
+  destroy() {
```
codecapitano, line 16: Who's calling teh destroy function?
We don't have any teardown mechanism like that in Faro.

I see it's in the type but can't find a service calling the destroy function.

Edit: 
Found it, it's part of `instrumentation.remove()`. TIL!
So it is correct, no action needed :)

codecapitano: When do we need to unmount the handlers?
I think it's either to save resources when tab is in background or when an instrumentation is manually removed.

I fear in this case you would need an own implementation.

packages/web-sdk/src/instrumentations/csp/instrumentation.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { Instrumentation } from '@grafana/faro-core';
+
+export class CSPInstrumentation extends BaseInstrumentation implements Instrumentation {
+  readonly name = '@grafana/faro-web-sdk:instrumentation-csp';
+  readonly version = VERSION;
+
+  constructor() {
+    super();
+    this.boundHandler = this.securitypolicyviolationHandler.bind(this);
+    document.addEventListener('securitypolicyviolation', this.boundHandler);
```
codecapitano: What is the reason to not setup in the initialize function? I assume it is to track CSP violation as early as possible.

If the handler fires during the init phase I assume it can: cause a crash or not send all data or not send anything at all bc Faro services aren't  completely setup.
Let's test this first.

packages/web-sdk/src/instrumentations/csp/instrumentation.ts
```diff
@@ -0,0 +1,38 @@
...
+  initialize() {}
+
+  destroy() {
+    document.removeEventListener('securitypolicyviolation', this.boundHandler);
+  }
+
+  private boundHandler: (ev: SecurityPolicyViolationEvent) => void;
+
+  public securitypolicyviolationHandler(ev: SecurityPolicyViolationEvent) {
+    this.api.pushEvent('securitypolicyviolation', {
```
codecapitano: Nit (no need to change): we have a convenience function to transform all object values to string.
`stringifyObjectValues(myObject)`

packages/web-sdk/src/instrumentations/csp/instrumentation.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        mockConfig({
+          transports: [mockTransport],
+          instrumentations: [instrumentation],
+        })
+      )!
+    );
+    faro.internalLogger.warn = jest.fn();
+
+    faro.instrumentations.remove(instrumentation);
+    expect(removeSpy).toHaveBeenCalledWith('securitypolicyviolation', instrumentation['boundHandler']);
```
codecapitano: Test is fails locally

```
CSPInstrumentation > ensures listener gets removed on teardown
-----
Error: expect(jest.fn()).toHaveBeenCalledWith(...expected)

Expected: "securitypolicyviolation", [Function bound ]

Number of calls: 0
```

packages/web-sdk/src/instrumentations/csp/instrumentation.ts
kpelelis: TIL

codecapitano (approved):
lgtm 🔥

Issue #844, opened 2025-01-15
webkitLineRegex can lock the page up on certain errors


elee1766:
We noticed that some users pages would "randomly" freeze after certain actions, but it would be sparingly, and inconsistent

we were able to eventually get a reproduction on chrome, showing a big trace block where one thing was running and hanging the entire page

![Image](https://github.com/user-attachments/assets/94c6aa68-3e74-4a84-84b9-0f49b923635b)

this i was able to trace back to https://github.com/grafana/faro-web-sdk/blob/main/packages/web-sdk/src/instrumentations/errors/stackFrames/const.ts

I don't really know what sort of conditions that are causing this, nor do i have the error that is causing this, sorry that this post is not more informational or convincing, but I thought I would post in case other people are having issues with their pages intermittently freezing after sending errors to faro, who can maybe provide some more information

i suspect it is some combination of errors from extensions or other packages that end up causing the regex to perform badly

eskirk:
hey @elee1766 thanks for the valuable report. I have seen something similar with regexes of this complexity - I think we will need to take another look at these expressions and see how we can simplify them. with a long enough line of a stack trace, the runtime could definitely behave like this.

elee1766:
thank you. i will try to figure out how to get some possible blobs that reproduce.

in the mean time @eskirk what would be the easiest way using the web+react sdk to disable stack trace parsing so peoples pages stop freezing?

eskirk:
@elee1766 a temporary workaround to avoid any stack trace parsing would be to remove the `ErrorsInstrumentation` from your Faro config. 

if you are using `getWebInstrumentations` in your Faro config, that is a shorthand for including the following instrumentations:
```js
[
    new ErrorsInstrumentation(),
    new WebVitalsInstrumentation(),
    new SessionInstrumentation(),
    new ViewInstrumentation(),
];
```

you could replace `getWebInstrumentations` with the subset of instrumentation you wanted to keep enabled until we fix this issue. 

hopefully this explanation makes sense and works for you in the short term.

elee1766:
> [@elee1766](https://github.com/elee1766) a temporary workaround to avoid any stack trace parsing would be to remove the `ErrorsInstrumentation` from your Faro config.
> 
> if you are using `getWebInstrumentations` in your Faro config, that is a shorthand for including the following instrumentations:
> 
> [
>     new ErrorsInstrumentation(),
>     new WebVitalsInstrumentation(),
[...]

PR #1316, opened 2025-06-12
feat(web-sdk): Properly allow users to override stack frame parsing settings

elee1766:
## Why

Certain extensions and libraries may throw exceptions with very long line lengths. These lines can cause the stack trace parser to lock up the browser thread, which is undesirable. 

It would be desirable for sites that require the use of these extensions or libraries to be able to override stack trace parsing settings, such that they may sacrifice their stack trace quality for site performance. 

## What

1. the stackFrames package has been moved to web-sdk/src/utils. this is a breaking change, but consumers of the existing functions if they are importing from the top level package will not have to update their imports.

I did this because it currently lives under the error instrumentation package, but that doesn't make sense, as it is not only use by the error instrumentation package, but also the stack trace parser that is passed in the makeCoreConfig that ultimately gets used by the core config. it seems more correct for it therefore to live in utils.

2. i've added the ability to pass in ones own stack trace parser to the browser config. this will be respected and this function is what will passed as parseStacktrace in the faro config.

3. create function `newStackTraceParser` which creates a StacktraceParser using new type StackframeParserOptions, which includes a single optional field for maximumLineLength

4. make  ErrorInstrumentation  use the stacktraceParser from the faro global config so all this stuff is configured from one place

## Links

#844 

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated


packages/web-sdk/src/config/getWebInstrumentations.ts
```diff
@@ -15,7 +15,7 @@ import type { GetWebInstrumentationsOptions } from './types';
 export function getWebInstrumentations(options: GetWebInstrumentationsOptions = {}): Instrumentation[] {
   const instrumentations: Instrumentation[] = [
     new UserActionInstrumentation(),
-    new ErrorsInstrumentation(),
+    new ErrorsInstrumentation(options.errorInstrumentationOptions),
```
codecapitano: Let's not use the options object here and move the property to the init function.
E. g. something like 

```
initializeFaro({
  stackframeParserOptions: {...}    
})
```

packages/web-sdk/src/config/types.ts
```diff
@@ -9,4 +11,5 @@ export interface GetWebInstrumentationsOptions {
   captureConsole?: boolean;
   captureConsoleDisabledLevels?: LogLevel[];
   enablePerformanceInstrumentation?: boolean;
+  errorInstrumentationOptions?: ErrorInstrumentationOptions;
```
codecapitano: Let's remove the ErrorInstrumentationOptions here and provide them via the Faro config only

packages/web-sdk/src/instrumentations/errors/instrumentation.ts
```diff
@@ -2,16 +2,21 @@ import { BaseInstrumentation, VERSION } from '@grafana/faro-core';
 
 import { registerOnerror } from './registerOnerror';
 import { registerOnunhandledrejection } from './registerOnunhandledrejection';
+import type { ErrorInstrumentationOptions } from './types';
 
 export class ErrorsInstrumentation extends BaseInstrumentation {
   readonly name = '@grafana/faro-web-sdk:instrumentation-errors';
   readonly version = VERSION;
 
+  constructor(private readonly options: ErrorInstrumentationOptions = {}) {
```
codecapitano: Let's read the settings form teh Faro config and remove the constructor.

The Faro config object is automatically attached via to teh instrumentation classes

packages/web-sdk/src/instrumentations/errors/instrumentation.ts
```diff
@@ -2,16 +2,21 @@ import { BaseInstrumentation, VERSION } from '@grafana/faro-core';
...
+    super();
+  }
+
   initialize(): void {
     this.logDebug('Initializing');
 
     registerOnerror(this.api);
 
-    registerOnunhandledrejection(this.api);
+    registerOnunhandledrejection(this.api, this.options);
```
codecapitano: Read the options from the faro config.

```
initialize(): void {

// ...

registerOnunhandledrejection(this.api, this.config.stackframeParserOptions);
}
```

packages/web-sdk/src/config/getWebInstrumentations.ts
elee1766: how can i do this?

the most i have from that context is "this.api", except the type for API does not expose the settings. 

it doesn't make sense to add this to the global config because this setting does not change the behavior of the "parseStacktrace", but only modifies the behavior of the hard coded stack trace parsing that is written into the errors instrumentation

one possible solution perhaps is that the errors instrumentation, instead of calling the function to parse the error stack directly, would use the stack trace parser provided by the config. what do you think? this would simplify the configuration

eskirk: similar to how it is done in other instrumentations (like the `WebVitalsInstrumentation`) you can look at this.config to see what global config options have been set. 

```js
...
  private intializeWebVitalsInstrumentation() {
    if (
      this.config?.trackWebVitalsAttribution === false ||
      this.config?.webVitalsInstrumentation?.trackAttribution === false
    ) {
      return new WebVitalsBasic(this.api.pushMeasurement, this.config.webVitalsInstrumentation);
    }
    return new WebVitalsWithAttribution(this.api.pushMeasurement, this.config.webVitalsInstrumentation);
  }
```

elee1766: i think it makes more sense then to just use the stacktraceparser in the errors instrumentation. i will refactor to do this

codecapitano:
> yeah so i didnt use it because that function wasnt respected by auto instrumentation (and also wasnt exposed, i think?)

Yes that's 100% correct.
* The custom parser was only respected on manual api call but not by the auto instrumentation
* The type was excluded on the BrowserConfig so it wasn't available via the Faro config.

codecapitano:
> @codecapitano
> 
> so you said
> 
> > Looks like there is a bug in the types for the Faro config.
> 
> digging deeper - it seems that this parseStackTrace is the initial setting for the Exceptions apis value for stacktraceparser, and the exceptions api then stores its own stacktraceparser i think in the function object scope or whatever for modification.

Yes that's a bit weird. If chnaging the parser it applies to API calls only but not to the auto instrumentation.

> maybe it is correct to allow passing this parameter in from the web configuration?

Yes this should be the goal. Provide a property in the config to defined a custom parser.

> its a bit confusing that the property here that supposedly would be passed to other things only modifies the exceptions api, and then wont necessarily match because of changeStackTraceParser.

That's correct. It's imho a bug.

> changeStacktraceParser is sort of weird to be honest existing in the first place. 

You mean having a function AND a config option at the same time right?
Yes I am all in to consolidate this into a single property for example.
Tbh I don't have any use case in mind where users need to change the stacktrace parser dynamically during a session.
+1 from my side to get rid of the function.

> its not even exposed right now. does it make more sense to just.. remove changeStacktraceParser?

That's what I meant with "bug in the Config type"

cheers

elee1766:
> Tbh I don't have any use case in mind where users need to change the stacktrace parser dynamically during a session.
+1 from my side to get rid of the function.

ok. this makes sense to me. I think that dynamically changing the stacktrace parser might have been used to make testing simpler or something.

I will make these changes then and see how it looks. I like the idea of having a single source of truth configuration variable for the stack trace parser.

elee1766:
hi @codecapitano sorry for the delay.

let me know how this looks.

codecapitano:
> hi @codecapitano sorry for the delay.
> 
> let me know how this looks.

@elee1766 many thanks ad also sorry from my side for the delay. 
Not sure if I've time this week for the review and it'll likely be next week.

Cheers

packages/web-sdk/src/instrumentations/errors/getErrorDetails.ts
```diff
@@ -1,12 +1,16 @@
...
 
 import { domErrorType, domExceptionType, objectEventValue, unknownSymbolString } from './const';
 import { getValueAndTypeFromMessage } from './getValueAndTypeFromMessage';
-import { buildStackFrame, getStackFramesFromError } from './stackFrames';
 import type { ErrorEvent } from './types';
 
-export function getErrorDetails(evt: ErrorEvent): [string | undefined, string | undefined, ExceptionStackFrame[]] {
+export function getErrorDetails(
+  evt: ErrorEvent,
+  stacktraceParser: StacktraceParser = newStackTraceParser()
```
codecapitano: Let's use `parseStacktrace` from the config? under the hood it's the same parsing logic.
We can update `parseStacktrace` to take the optional options object.

packages/web-sdk/src/instrumentations/errors/getErrorDetails.ts
```diff
@@ -39,7 +44,10 @@ export interface ErrorDetails {
   stackFrames?: ExceptionStackFrame[];
 }
 
-export function getDetailsFromErrorArgs(args: [any?, ...any[]]): ErrorDetails {
+export function getDetailsFromErrorArgs(
+  args: [any?, ...any[]],
+  stacktraceParser: StacktraceParser = newStackTraceParser()
```
codecapitano: Same here, why not use `parseStacktrace`?

packages/web-sdk/src/instrumentations/errors/getErrorDetails.ts
codecapitano: Another option could be to receive the parser from `api.getStacktraceParser`
Looks like we currently have three functions doing teh same thing.

elee1766:
@codecapitano are you asking why not just use the function instead of the um, stacktracer parser interface?

yeah i dont know really. I'm happy to get rid of the stack trace parser interface and unify everything to just use the parseStackTrace function type.

i only did it like this because StacktraceParser interface existed so i thought there was some reason you guys wanted it to exist so i kept using it. not sure if there is some reason its there

packages/web-sdk/src/instrumentations/errors/getErrorDetails.ts
elee1766: in actual usage the parser is received here by api.getStacktraceParser

kpelelis:
Hey @elee1766, I am doing some grooming on the open issues and I was wondering if this something that you still want to pursue

elee1766:
@kpelelis hi

in all honesty I dropped faro for posthog on our more critical sites. for a few months we were using a hacked up branch to get rid of the bugs but faro never moved forward much on session replay, and it wasn't worth the continued effort fixing the patch every faro update on my pvt fork


happy to try to wrap up this pr and finish the features, just I'm not using it anymore so it will be hard for me to test in field

domasx2:
oof, sorry for dropping this. Let's close for now and return if we find strong motivation to solve this

Issue #485, opened 2024-02-16
Support tracking measuring of Soft Navigations within SPAs 


codecapitano:
## Why

Instead of loading individual web pages per user navigation, Single Page Applications (SPAs) use  _soft-navigations_ when a user navigates the app.
Here, the content of a page is changed by dynamically altering the browser URL by leveraging the history API and by replacing a pages content through JavaScript. 

To get insights into navigation performance and UX metrics within in Single Page Applications Faro should support tracking soft-navigations.

## What

The [Chrome team is currently experimenting with tracking soft navigations](https://developer.chrome.com/docs/web-platform/soft-navigations-experiment) which are emitted as a dedicated PerformanceEntry form a PerformanceObserver hooking into the browsers performance timeline. 

Once this is activated in chrome we can start using this api to get the data. Given the big market share of Chrome we can get a good amount of data even if other browser need some more time to implement the api.

In the meantime we need to find out what else we can do.
For example [Google's Web-Vitals library provides an experimental mode to  report Web-Vials for soft navigations.](https://github.com/GoogleChrome/web-vitals/tree/soft-navs?tab=readme-ov-file#report-metrics-for-soft-navigations-experimental)

cedricziel:
Increment 0: Highlight views in the UX set via `setView` API

PR #1317, opened 2025-06-12
feat(web-sdk): add Navigation event instrumentation

eskirk:
## Why

very requested feature for JS apps that do not use React/react-router instrumentations. 

## What

this instrumentation will track page navigation and URL changes in any instrumented applications. 

## Links

#485 

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/navigation/instrumentation.ts
+++ b/packages/web-sdk/src/instrumentations/navigation/instrumentation.ts
@@ -1,4 +1,4 @@
-import { BaseInstrumentation, VERSION } from '@grafana/faro-core';
+import { BaseInstrumentation, VERSION, type Config } from '@grafana/faro-core';
 
 import {
   NAVIGATION_BACK,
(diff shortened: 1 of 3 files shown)


packages/web-sdk/src/instrumentations/navigation/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+  readonly name = '@grafana/faro-web-sdk:instrumentation-navigation';
+  readonly version = VERSION;
+  readonly originalHistory: History = window.history;
+  readonly originalNavigation: Navigation = window.navigation;
+
+  private currentUrl: string | null = null;
+  private currentHash: string | null = null;
+
+  initialize() {
+    console.log('NavigationInstrumentation initialized', {
```
codecapitano: console.log

codecapitano (changes requested):

packages/web-sdk/src/instrumentations/navigation/instrumentation.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    instrumentation.originalHistory.go = function (delta) {
+      originalGo.apply(instrumentation.originalHistory, [delta]);
+      instrumentation.api.pushEvent(NAVIGATION_GO, {
+        type: NAVIGATION_EVENT_TYPE,
+        delta: delta ? String(delta) : '',
+      });
+    };
+
+    window.addEventListener('popstate', (event) => {
```
codecapitano, line 146: Popstate will be triggered for `history.back(),` `.forward()` and `.go()`.
So we always send two events.

I assume these are considered as something like sidecar events to provide more context ? 
Let's not enable them by default. It's technical browser events but not really related to the trace of the user.
For example we can provide a setting to let users enable them if they really need it. 

Do you have an example how they enhance the RCA flow?

PR #1343, opened 2025-06-23
chore(build): don't build against node 23 anymore because it is EoL

codecapitano:
## Why

Node 23 is EoL and we have first dependencies failing to build.

## What

Remove Node 23 from GH action 

## Checklist

- [ ] Tests added
- [x] Changelog updated
- [ ] Documentation updated


rrmf (approved):
🚀

PR #1418, opened 2025-07-15
feat(user-actions): Add user aciton severity

kpelelis:
## What

Add a new severity field to the user action entity, that allows users to tag and distinguish between different user flows (e.g., critical vs. normal) for more granular handling and monitoring.


codecapitano:
What to the severity mean in regards to user actions?
What is the difference between none and normal or between high and critial ?

User actions are already events which have higher precedence then usual events.
It's too many levels and only critical should be fine bc a user action is already an event with level high.

Usually everything not critical or high is a standard event, but i think I get your idea to also allow to downgrade severity if users desire to use actions for other use cases.

Suggestion: let's remove the levels which are very close to others like normal and high.
  None = 'none',
  Normal = 'normal',
  High = 'high',
  Critical = 'critical',

Note:
I think it's important to not overuse user actions because it may degrade their meaning and it adds more load to the browser then usual events.

PR #1467, opened 2025-08-06
feat(meta): make SDK version validation configurable (#571)

### LINKED ISSUE #571: validate meta.sdk value before sending payload
## Description

the web SDK should enforce a constraint on the meta.sdk value such that it confirms to semver standards. 

## Context

arbitrary SDK version values are causing issues in faro collector metrics cardinality

anglerfishlyy:
## Why

Currently, the meta.sdk value is sent in the payload without validation. If it’s undefined or malformed, it can lead to issues downstream when consuming the payload. Validating it ensures data consistency and prevents potential errors.

## What

- **Added** a **check** to **ensure meta.sdk exists** and is c**orrectly structured before sending it** **with** the **payload**.
- Improved **type safety** around the SDK metadata to avoid silent failures.

## Links
Closes #571

## Checklist

- [y] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

**Let me know if anything else needs to be updated or improved!**


anglerfishlyy:
@eskirk, would appreciate your review when you have time.

anglerfishlyy:
Hey @eskirk  , while working on the meta validation changes I noticed some of the SDK setup/validation logic looks pretty similar in a few spots. Is it usually kept that way for clarity, or do we try to centralize it into a shared utility in Grafana projects? Just wondering if that’s a performance/maintainability call or more of a readability preference here.

eskirk:
@anglerfishlyy can you point out what you are talking about? probably mostly due to readability/maintainability. 

could you also add some tests for this functionality you added?

eskirk:
also - lets make this configurable. I was just speaking to a customer today who specifically mentioned that they would **not** like to have the version field validated using semver standards. this adds a bit of overhead to the PR, but if we could add this as a configurable option in the core config, that would be great. 

a to-do list for this PR:
- [ ] add option to enable/disable semver validation
- [ ] add tests for this new functionality
- [ ] remove added dependencies since we are just using regex now

anglerfishlyy:
Hi, This improves SDK version validation and makes it configurable.
It uses a better check for version format and shows a warning if the version looks wrong, instead of stopping the program.
I also added tests to make sure the validation works.

packages/web-sdk/src/metas/sdk/meta.ts
```diff
@@ -1,10 +1,22 @@
...
-    name: '@grafana/faro-core',
-    version: VERSION,
-    integrations: faro.config.instrumentations.map(({ name, version }) => ({ name, version })),
-  },
-});
+// Official semver v2.0.0 regex from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+const fullSemverRegex =
+  /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|[0-9A-Za-z-][0-9A-Za-z-]*)(?:\.(?:0|[1-9]\d*|[0-9A-Za-z-][0-9A-Za-z-]*))*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?$/;
+
+export const sdkMeta: MetaItem<Pick<Meta, 'sdk'>> = (currentVersion?: string) => {
```
codecapitano: I can't see where in Faro the `currentVersion` parameter is injected. Can you elaborate how this works?

packages/core/src/config/types.ts
```diff
@@ -46,6 +46,11 @@ export interface Config<P = APIEvent> {
    */
   isolate: boolean;
 
+  /** enable or disable semantic version validation for the SDK metadata.
+   *
+   */
+  validateSdkMeta?: boolean;
```
codecapitano, line 52: What drives the decision to make version check optional?

anglerfishlyy: Making it optional lets users with custom instrumentation provide their own versioning without breaking Faro, while standard semver validation still works for normal usage.

codecapitano:
Thank you for the contribution, but we won't be merging this PR.

The reason is that it addresses a hypothetical edge case that would only affect a very small subset of users who are building custom implementations on top of Faro. From a product perspective, this doesn't justify adding a new setting and the associated changes.

If you need custom versioning, there are multiple options available:

**Option 1: Set custom SDK metadata in the Faro config**

```ts
initializeFaro({
  metas: [
    () => ({
      sdk: {
       name: 'foo',
       version: '123',
      },
    }),
   ],
 })
```

**Option 2: Use the `beforeSend` hook to modify version numbers**

This approach allows you to rewrite version numbers as needed during the send process.

PR #1475, opened 2025-08-08
fix: prevent filename from being badly extracted when no function name is resolved

alizeait:
## Why

`filename` is not being extracted properly when no function name or method is provided on chrome, and the URL includes parentheses, causing the filename to be wrongly resolved, as shown in this test snapshot

```
       {
          "colno": 27,
    -     "filename": "http://localhost:8080/[brackets]/(paranthesis)/file.js",
    -     "func": undefined,
    +     "filename": "paranthesis)/file.js",
    +     "func": "http://localhost:8080/[brackets]/",
          "lineno": 41,
        }, 
```
## What

To fix this, the `webkitLineRegex` was changed to have a negative lookahead for function name capture group preventing it from matching URLs

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Next
 
 - Chore (`@grafana/faro-*`): set default node version to lts/jod
+- Fix (`@grafana/faro-web-sdk`): Update `webkitLineRegex` to prevent the
+  function name capture group from matching URLs
 
 ### Breaking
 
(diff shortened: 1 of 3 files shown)


codecapitano (approved):
Awesome thank you so much for the contribution

eskirk (approved):

PR #1500, opened 2025-08-21
Import faro from correct package

zackman0010:
## Why

After updating to 2.0.0-beta, our API calls started failing with an exception being thrown from within Faro. Upon investigation, we found that Faro was uninitialized in the OpenTelemetry interceptor, causing `faro.api` to be undefined. After looking at the minified code and comparing it to other usages, we noticed that the existing code that referenced the `faro` variable imported it from faro-web-sdk, which led to the minified code `e.faro`. But the code that was throwing an exception imported it from `faro-core`, which was minified to `lo`, an empty object.

## What

Changed the faro-core imports to instead import from faro-web-sdk.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


codecapitano:
Good catch @zackman0010 and many thanks for the contribution.
I remember we [fixed a related issue](https://github.com/grafana/faro-web-sdk/pull/1139) a while ago. 

Note to the Faro team:
* [GH Issue to prevent wrong imports in the future](https://github.com/grafana/faro-web-sdk/issues/1506).

zackman0010:
Added the missing semicolon that the linter was complaining about and rebased onto latest main.

packages/web-tracing/src/faroUserActionSpanProcessor.ts
```diff
@@ -1,7 +1,8 @@
 import { type Context, SpanKind } from '@opentelemetry/api';
 import type { ReadableSpan, Span, SpanProcessor } from '@opentelemetry/sdk-trace-web';
 
-import { faro, UserActionState } from '@grafana/faro-core';
+import { UserActionState } from '@grafana/faro-core';
```
codecapitano, line 4: We should re-export user action state from the web-tracing package as well to get rid of all imports from core. 
If the enum is not correctly resolved the user actions feature won't work correctly.

zackman0010: @codecapitano I don't think that's actually necessary due to the way TS enums are compiled to JS. The UserActionState enum is compiled to a JS object `{"Started": 0, "Halted": 1, "Cancelled": 2, "Ended": 3}`. So in the compiled JS, enums are stored as plain integers and enum comparisons are just integer comparisons.

martinkuba: Thanks @zackman0010 , this is true that the enum values are compared as plain numbers in the compiled code. We might still want to update the import to be consistent and not duplicate the enum definition in multiple bundles, but we can do that in a separate PR.

codecapitano: Thanks @zackman0010 and @martinkuba.
Absolutely fine for me as well to adjust in a follow up.

codecapitano (approved):

PR #1507, opened 2025-08-25
fix(web-tracing): Prevent web-tracing from importing faro from core

### LINKED ISSUE #1506: web-tracing package: ensure that the faro object can only be imported from the web-sdk package
# Why 

The web-sdk package pulls faro-core as a dependency an re-exports the Faro object. 
But since the faro-core also exports the faro object, it is possible to import is from the core package which can happen easily. 
The problem is that this import doesn't work when building the iife bundle for the CDN version causing the faro object to be undefined.

# What 

Ensure that the web-tracing package can be only use faro-core exports which are re-exported by the web-tracing package. 

# Links
Related issues: 
* #1139 
* #1500

zackman0010:
## Why

Importing `faro` from `core` in `web-tracing` will cause runtime failures when using the iife bundle due to the way minification works. The correct import should be from `web-sdk`.

## What

Added an ESLint rule that will cause the quality:lint check to fail if the wrong import is used.

A couple of notes:
1) The new rule will cause CI failures until #1500 is merged.
2) ESLint 9 does not support nested files like this uses, so this rule will need to be moved into the flat file if ESLint is upgraded. To make sure that won't get missed, I also wrote a test that validates the error is printed when running ESLint, but I kept it in [another branch](https://github.com/zackman0010/faro-web-sdk/commit/230c87f6f07436ca73e234f4ebdf77d1b462931b) due to it requiring some dependency changes to run ESLint in a Jest.

## Links

Resolves #1506 

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,6 +8,7 @@
 # App-specifics
 .cache
 .eslintcache
+**/.eslintrc.js
 .husky
 coverage
 cypress/videos
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+  root: true,
   extends: ['@grafana/eslint-config', 'plugin:import/errors', 'plugin:import/warnings', 'plugin:import/typescript'],
   plugins: ['eslint-plugin-no-only-tests'],
   settings: {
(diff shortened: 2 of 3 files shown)


zackman0010:
Closing in favor of the other PR

PR #1508, opened 2025-08-26
chore: add eslint rule to not allow importing core from web-tracing

martinkuba:
## Why

Importing from core in the web-tracing package can result in duplicated code between the web-sdk and web-tracing bundles. This was specifically an issue when importing the `faro` global singleton.

## What

Since this is easy to introduce and hard to detect, we can disallow any import from core. This change adds an eslint rule for this purpose. 

## Links

Fixes https://github.com/grafana/faro-web-sdk/issues/1506

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,4 +43,21 @@ module.exports = {
     project: 'tsconfig.base.json',
     tsconfigRootDir: __dirname,
   },
+  overrides: [
+    {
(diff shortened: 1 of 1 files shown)


codecapitano (approved):
Some imports need to updated. Indicates rule is working as expected 🔥 
Many thanks 🙏

PR #1551, opened 2025-09-05
fix(user-actions): custom severity wasn't added to user action attributes

codecapitano:
## Why

Custom triggers and severity weren't were missing in user action attributes.

## What

* Ensure trigger and severity are added
* Export `UserActionSeverity` from web-sdk and faro-react so users can import and use it more easy

## Before
_Action attributes contain the default values for trigger and severity_
Screenshot 2025-09-05 at 16 41 34

## After
_Action attributes contain the custom values for trigger and severity_

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next
 
+- Fix (`@grafana/faro-web-sdk`): Fixed an issue where custom severity and custom trigger properties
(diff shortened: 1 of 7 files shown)


kpelelis (approved):
LGTM

PR #1674, opened 2025-10-20
refactor(user actions): move user actions tracking to sdk

martinkuba:
## Why

The mechanism for how user actions are tracked should be transparent to users of the API, and it should be possible to replace the implementation of user actions tracking without changing the API.

## What

- removed extend() and halt() methods from the API
- added UserActionController class to the SDK, which encapsulates the mechanism for tracking events related to a single user action

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -83,7 +83,7 @@ jobs:
         run: yarn build
 
       - name: E2E tests
-        uses: cypress-io/github-action@0df54ae94d937756289cf9540ecfe1c76167884a
+        uses: cypress-io/github-action@6fc22051da87f9a2e6cf3d289d28f262bff0249c
         env:
           DATABASE_HOST: database
           DATABASE_PORT: 5432
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -81,7 +81,7 @@ jobs:
         run: yarn build
 
       - name: E2E tests
-        uses: cypress-io/github-action@0df54ae94d937756289cf9540ecfe1c76167884a
+        uses: cypress-io/github-action@6fc22051da87f9a2e6cf3d289d28f262bff0249c
         env:
           DATABASE_HOST: database
           DATABASE_PORT: 5432
--- a/yarn.lock
+++ b/yarn.lock
@@ -14696,8 +14696,8 @@ __metadata:
   linkType: hard
 
 "sass@npm:^1.57.1":
-  version: 1.93.2
-  resolution: "sass@npm:1.93.2"
+  version: 1.93.3
+  resolution: "sass@npm:1.93.3"
   dependencies:
     "@parcel/watcher": "npm:^2.4.1"
     chokidar: "npm:^4.0.0"
@@ -14708,7 +14708,7 @@ __metadata:
       optional: true
   bin:
     sass: sass.js
-  checksum: 10c0/5a19f12dbe8c142e40c1e0473d1e624898242b1c21010301e169b528be8c580df6356329c798522b525eb11eda4b04b9b77422badc55c47889600f8477201d2b
+  checksum: 10c0/b9facc64de10c9d1514272c1dcbb48ca99d5f591a1cd43fd27d641275d9d95948f1299107ab5b309e2abb552667cca84a38a1a3df5116eb72eba4144bf850b6a
   languageName: node
   linkType: hard
 


packages/web-sdk/src/instrumentations/userActions/instrumentation.ts
```diff
@@ -19,7 +19,7 @@ export class UserActionInstrumentation extends BaseInstrumentation {
 
     this._userActionSub = userActionsMessageBus.subscribe(({ type, userAction }) => {
       if (type === 'user_action_start') {
-        proceessUserActionStarted(userAction);
+        processUserActionStarted(userAction);
```
kpelelis, line 22: As a non-native speaker I spent a good amount of time trying to figure out what the change was 😂

packages/web-sdk/src/instrumentations/userActions/userActionController.ts
```diff
@@ -0,0 +1,146 @@
...
+  private allMonitorsSub?: Subscription;
+  private stateSub?: Subscription;
+  private followUpTid?: number;
+  private haltTid?: number;
+
+  private isValid = false;
+  private isHalted = false;
+  private runningRequests = new Map<string, HttpRequestMessagePayload>();
+
+  constructor(private ua: UserActionInterface) {}
```
kpelelis: nit: we don't use `ua` as a short for user actions as it can be confused for user agent

packages/web-sdk/src/instrumentations/userActions/userActionController.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.clearTimer(this.haltTid);
+    this.haltTid = startTimeout(
+      this.haltTid,
+      () => {
+        // If still halted after timeout, end
+        if (this.ua.getState() === UserActionState.Halted) {
+          this.endAction();
+        }
+      },
+      defaultHaltTimeout
```
kpelelis, line 116: This used to be a configurable parameter. I would suggest modifying the API a little bit so that users can still make use of it.

packages/core/src/api/userActions/userAction.ts
```diff
@@ -21,20 +18,14 @@ export default class UserAction extends Observable implements UserActionInterfac
...
-  private _timeoutId?: number;
   private _itemBuffer: ItemBuffer<TransportItem>;
   private _transports: Transports;
-  private _haltTimeoutId: any;
-  private _isValid: boolean;
 
   constructor({
     name,
     parentId,
-    haltTimeout,
```
kpelelis, line 37: This comment applies for all `haltTimeout`, `extend` and `halt` methods. Although the change is needed, all these 3 are part of the public API and removing them without a deprecation notice, effectively breaks semantic versioning if we decide not to do a major release (which I guess we don't want to do yet).

kpelelis (changes requested):
This overall LGTM but we will have to address the breaking changes that this pull request introduces first, thus I will not approve

packages/core/src/api/userActions/types.ts
```diff
@@ -9,16 +9,12 @@ export enum UserActionState {
   Ended,
 }
 
-export type HaltPredicate = () => boolean;
-
 export interface UserActionInterface {
```
codecapitano, line 16: I wonder if any of the functions need to be exposed in teh public interface.
Do we have any concrete use cases for exposing addItem, end, cancel, getState to the users?
Personally i always tend towards starting super simple and expose more to users if there's a need for it.
Wdyt?

packages/core/src/api/userActions/userAction.ts
codecapitano: I think in this case it's fine because the change is for another iteration on Faro v2.beta which already contains breaking user action api changes. Though we should update the docs to describe breaking change between beta.x and beta.y versions.

Also the feature was only available as preview stating that breaking changes might happen.

packages/web-sdk/src/instrumentations/userActions/processUserActionEventHandler.test.ts
```diff
@@ -3,18 +3,13 @@ import { jest } from '@jest/globals';
 import { Observable, UserActionState } from '@grafana/faro-core';
```
codecapitano, line 3: "@grafana/faro-core"' has no exported member named 'UserActionState'. Did you mean 'UserAction'?

martinkuba: It is exported here
https://github.com/grafana/faro-web-sdk/blob/main/packages/core/src/index.ts#L6

Since getState() is part of UserActionInterface, then the enum must be exported as well.

packages/web-sdk/src/instrumentations/userActions/userActionController.ts
martinkuba: The UserAction class accepts this as a parameter, but it's not currently being passed when user action is instantiated [here](https://github.com/grafana/faro-web-sdk/blob/main/packages/core/src/api/userActions/initialize.ts#L42-L49), and it's also not part of the public configuration. I would suggest that we add it later if we find there is a need for it.

packages/core/src/api/userActions/userAction.ts
martinkuba: I am marking this as resolved, but @kpelelis please feel free to reopen if you still disagree.

packages/core/src/api/userActions/types.ts
martinkuba: I have moved all of these functions to a separate interface.

PR #1680, opened 2025-10-22
extract event tracking to a separate module

martinkuba:
## Why

The mechanism for tracking events that happen within a given time window can be reused for different instrumentations.

## What

This extracts the mechanism to a module that is separate from the User Action instrumentation.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


martinkuba:
Implemented as part of https://github.com/grafana/faro-web-sdk/pull/1691

PR #1691, opened 2025-10-25
feat: navigation instrumentation

martinkuba:
## Why

Capture events that represent SPA soft navigations.

## What

This adds a new instrumentation for capturing navigation events. It includes the following changes:

- added a new observable for URL changes
- added a new observable for interactions (click, keydown etc.)
- changed all observable monitors to be singletons, so that they can be used by multiple instrumentations
- moved all observable monitors to `instrumentations/_internal` folder to signal that these are shared functions but not intended for external use (I am fine renaming the folder or moving elsewhere)
- added `activityWindowTracker` class that is responsible for tracking related events within a given time window. This could be reused by the user actions instrumentation.
- added `navigation` instrumentation that uses the `activityWindowTracker` and observable monitors to detect soft navigations

## Links

Addresses https://github.com/grafana/faro-web-sdk/issues/485

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


packages/web-sdk/src/instrumentations/_internal/monitors/httpRequestMonitor.ts
eskirk, line 1: I like all the changes here, feels more resilient 👍🏼

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts
```diff
@@ -0,0 +1,145 @@
...
+          if (destination?.sameDocument && typeof destination.url === 'string') {
+            emit('navigate', destination.url);
+          }
+        } catch (_err) {
+          // Swallow to avoid impacting host app
+        }
+      };
+      (window as any).navigation.addEventListener('navigate', onNavigateHandler as any);
+
+      const NavigateEventCtor = (window as any).NavigateEvent;
```
eskirk: does Ctor mean "constructor"? if so lets just go for verbosity instead of brevity

eskirk (commented):
wish I had more comments but this is super clean. one comment about variable naming, but otherwise looks good to me.

demo/src/client/faro/initialize.ts
```diff
@@ -49,6 +49,7 @@ export function initializeFaro(): Faro {
     },
 
     trackUserActionsPreview: true,
+    trackNavigationPreview: true,
```
codecapitano, line 52: I know we use preview postfix for user actions.
Do you think it would make more sense to prefix as `Experimental` instead?

packages/web-sdk/src/instrumentations/navigation/instrumentation.ts
```diff
@@ -36,10 +36,9 @@ export class NavigationInstrumentation extends BaseInstrumentation {
           msg.events?.some((e: any) => e.type === 'dom-mutation')
         ) {
           faro.api.pushEvent('navigation', {
-            from: msg.events?.find((e: any) => e.type === 'url-change')?.from,
-            to: msg.events?.find((e: any) => e.type === 'url-change')?.to,
-            trigger: msg.events?.find((e: any) => e.type === 'url-change')?.trigger,
-            same_document: 'true',
+            fromUrl: msg.events?.find((e: any) => e.type === 'url-change')?.from,
+            toUrl: msg.events?.find((e: any) => e.type === 'url-change')?.to,
+            sameDocument: 'true',
```
codecapitano: NIT: you can destructure the result to a constant so you only need to call the function one time

```ts
const { from, to } = msg.events?.find((e: any) => e.type === 'url-change')
```

codecapitano (approved):
This is great. Love the refacorings and hardening the monitor observables.

PR #1772, opened 2025-11-11
chore(config): updated configuration for user actions instrumentation

martinkuba:
## Why

The user actions feature is no longer in preview in the upcoming v2 release. Users will not have to opt-in by setting the the `trackUserActionsPreview` config flag.

## What

This change removes the `trackUserActionsPreview` flag from configuration. It also consolidates other configuration related to user actions under `userActionsInstrumentation` section.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,19 +2,26 @@
 
 ## Next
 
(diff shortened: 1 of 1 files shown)


codecapitano (approved):
LGTM

Issue #1490, opened 2025-08-18
Add support for dynamic headers in Faro Web SDK transport options


juliendef:
## Description

Currently, the Faro Web SDK only accepts static headers for transport requests.
We would like to add support for dynamic headers, so that authentication tokens (e.g. JWT) can be injected at request time.

Example use case:
When sending data through a proxy, the proxy requires an Authorization header. Since tokens can expire and need refresh, we need a way to compute headers dynamically before each request.

## Proposed solution

Extend the OtlpTransportRequestOptions interface to allow headers to be a function (sync or async) in addition to a static object:

```ts
export interface OtlpTransportRequestOptions 
  extends Omit<RequestInit, 'body' | 'headers'> {
  headers?:
    | Record<string, string>
    | (() => Record<string, string> | Promise<Record<string, string>>);
}
```
The transport layer should resolve this function before making a request.

## Context

- Enables token-based authentication when using proxies.
- Allows rotation/refresh of auth tokens without restarting the SDK.
- Makes integration more secure and flexible.

juliendef:
Hi Marco, this should be treated as a feature request. 

I still lack sufficient context, hands‑on experience, and familiarity with the repository to draft a proper PR yet.
Thanks

juliendef:
Thank to all contributors.

@codecapitano 

Contrary to what the issue asks, the ability to set headers asynchronously for **ALL** transports (OtlpHttpTransport) has not been implemented.

To finally really consider this issue closed, I made an effort to make this possible. 
Can you take a look at it please https://github.com/grafana/faro-web-sdk/pull/1955 ? 🙏🏼 
Thanks

PR #1788, opened 2025-11-23
feat(transport): support dynamic headers in transports

rbaradari:
## Why

Currently, the Faro Web SDK only allows to configure static headers for transport requests.

When sending data through a proxy it might require a valid authentication token (e.g. JWT). Since tokens can expire and need to be refreshed, we need a way to lookup dynamic header values for each request.

## What

Extended the RequestOptions `headers` to also support callback functions that return a header value. This change is implemented for the Fetch transport and to stay aligned also for the OLTP HTTP transport.

I've not opted to also implement a Promise based asynchronous lookup. That would have significantly increased the change set. It is expected that any token refreshes are done by the application. When building requests by Faro we can directly lookup the currently available token.

## Links

Addresses feature request #1490.

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Next
 
+- Feature (`@grafana/faro-web-sdk`): Fetch transport now supports dynamic header values.
+  Each header can be a static string or a function returning a string, resolved at request
+  time (#1490).
+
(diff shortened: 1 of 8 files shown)


codecapitano (approved):
Great enhancement! Many thanks for your contribution 🥳 🙏

Issue #1604, opened 2025-09-24
Isolated Faro: Console and other global instrumentations are subject to a race condition based on intialization order


eskirk:
## Description

based on the way the code is written, when multiple Faro instances are living in the same browser window, the registerOnerror, registerOnunhandledrejection, and Console instrumentations will run multiple times. these instrumentations overwrite each other and will only apply to the latest Faro initialization. depending on how an application is built, this can cause undefined behavior. 

## Steps to reproduce

TBD

## Expected behavior

the Faro API needs to be called dynamically rather than set in the anonymous function that patches the original global APIs 

## Actual behavior

global API patches are overwritten by the latest initialization

martinkuba:
The onerror and unhandledrejection events actually work fine right now -

The onerror global function gets overridden with each instance, but it results in a chain of wrapped functions. This is not ideal, but it's not broken - all instances are notified of these errors.

The unhandledrejection event is observed by each instances calling `addEventListener('unhandledrejection')`, so there is no issue there.

rrmf:
@eskirk @martinkuba @codecapitano Can we close this issue? All referenced sub-issues are closed.

PR #1825, opened 2026-01-05
fix(ConsoleInstrumentation): make console instrumentation work with multiple SDK instances

### LINKED ISSUE #1829: Fix console instrumentation race condition with multiple instances

martinkuba:
## Why

Fixes #1829 - Console instrumentation overwrites global console methods each time a new isolated Faro instance is initialized, causing only the last instance to receive console events.

## What

Refactored console instrumentation to use the existing Observable/monitor pattern:

- added consoleMonitor.ts that patches console methods once and exposes an Observable
- multiple instances of the Console instrumentation now use this monitor, ensuring all receive console events
- added tests verifying isolated instances both receive console events

Also
- updated makeCoreConfig.test.ts to suppress console.error messages (does not change test behavior)
- fixed mocks in transport.test.ts, so that internal errors are not in test output

## Links

[Isolated Faro: Console and other global instrumentations are subject to a race condition based on intialization order · Issue #1604 · grafana/faro-web-sdk](https://github.com/grafana/faro-web-sdk/issues/1604)

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/console/instrumentation.test.ts
+++ b/packages/web-sdk/src/instrumentations/console/instrumentation.test.ts
@@ -7,10 +7,30 @@ import { __resetConsoleMonitorForTests } from '../_internal/monitors/consoleMoni
 
 import { ConsoleInstrumentation } from './instrumentation';
 
+// Mock globalObject for SDK's _faroInternal registration.
+// This enables testing non-isolated Faro instances by providing a clean object
+// for each test. This is separate from global.console below - they serve different SDK subsystems.
+let mockGlobalObject: Record<string, unknown> = {};
+
+// Mock globalObject module - all internal imports use the index path
+jest.mock('@grafana/faro-core/src/globalObject', () => ({
+  get globalObject() {
+    return mockGlobalObject;
+  },
+}));
+
+function resetMockGlobalObject() {
+  mockGlobalObject = {};
+}
+
 describe('ConsoleInstrumentation', () => {
   const originalConsole = console;
 
   beforeEach(() => {
+    resetMockGlobalObject();
+
+    // Window global mock for capturing console.log/error calls.
+    // The ConsoleInstrumentation patches these methods to intercept logs.
     global.console = {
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/console/instrumentation.test.ts
eskirk, line 1: can we add another test with parent1 & child1/2 (parent is not isolated, children are isolated) what is the expected behavior then?

this is a bit tricky because I think we would expect to see a console error in either child also show up in the parent(?) and then also show up in whichever child produced the console error - but I do not think we have a way to guarantee this behavior at this time. 

@codecapitano what do you think?

codecapitano: Very Good question @eskirk. Both views make sense to me.
Currently for ease of use my stance is to treat "isolated" as "as isolated as possible" so there'd be no need to share events with a main shell level Faro instance. 

I'd be pragmatic and wait if we get (enough) user feedback on that topic.

Wdyt @eskirk @martinkuba

codecapitano (approved):
Same as @eskirk said for the test case. Otherwise lgtm

packages/web-sdk/src/instrumentations/console/instrumentation.test.ts
martinkuba: Correct me if I am wrong, but the only difference with not isolated is that it 
- adds faro to the global window object
- can only be initialized once

Instrumentations are otherwise initialized the same way. So, I don't think that adding a test with not isolated instance will test anything that is already covered.

With this change, all faro instances (from the same shared code) that include the Console instrumentation will report console logs.

martinkuba: I have added a test that has one isolated and one non-isolated instance. It does not test anything new, but it does show/document that the behavior is the same with non-isolated instance.

PR #1839, opened 2026-01-12
test(errors): added tests for onerror and unhandledrejection events with multiple instances

martinkuba:
## Why

We had no tests for the onerror global handlers and the unhandledrejection event with multiple instances.

This is related to https://github.com/grafana/faro-web-sdk/issues/1830 and https://github.com/grafana/faro-web-sdk/issues/1831. 

## What

Added tests that verify that these two events work with multiple SDK instances.

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/errors/registerOnunhandledrejection.ts
+++ b/packages/web-sdk/src/instrumentations/errors/registerOnunhandledrejection.ts
@@ -5,8 +5,11 @@ import { primitiveUnhandledType, primitiveUnhandledValue } from './const';
 import { getErrorDetails } from './getErrorDetails';
 import type { ExtendedPromiseRejectionEvent } from './types';
 
+// Store handlers for cleanup in tests
+const registeredHandlers: Array<(evt: ExtendedPromiseRejectionEvent) => void> = [];
+
 export function registerOnunhandledrejection(api: API): void {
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/errors/registerOnunhandledrejection.ts
```diff
@@ -28,5 +31,16 @@ export function registerOnunhandledrejection(api: API): void {
     if (value) {
       api.pushError(new Error(value), { type, stackFrames });
     }
+  };
+
+  window.addEventListener('unhandledrejection', handler);
```
codecapitano, line 36: Just to confirm, storing handlers in `registeredHandlers` is just for the tests right?
When can we run into a situation where a single instance might register multiple handlers?

codecapitano (approved):
Just some questions otherwise lgtm

packages/web-sdk/src/instrumentations/errors/registerOnunhandledrejection.ts
martinkuba: Correct, it's only there for tests. It could also be used for cleanup. Some of our instrumentations implement `destroy()` and some do not - this makes it possible for the errors instrumentation to unsubscribe this event.

Issue #1845, opened 2026-01-13
TracingInstrumentation incompatible with multiple SDK instances (micro-frontends)


martinkuba:
## Description

The TracingInstrumentation has two interconnected problems that make it problematic in micro-frontend architectures:

  1. **spans are reported always only by the first-initialized micro frontend app**
This is because we initialize OTel instrumentations by registering a global tracer provider and then call registerInstrumentations(), which assigns the global provider to the instrumentations. Global provider can be registered only once, so tracer provider from the first instance will always be the only one used.

  2. **fetch/xhr instrumentations configuration of the last-initialized app is used**
This is the inverse of the tracer provider - the last-initialized instance configuration (ignoreUrls, propagateTraceHeaderCorsUrls, applyCustomAttributesOnSpan) is used for reported spans. The fetch/xhr instrumentations wrap global APIs, and they do this only once. This is correct because there really should only be one trace and therefore one span. But the mismatch between the global provider and the instrumentation configuration is problematic. 

## Expected behavior

When an HTTP calls is made
- ideally we automatically determine which micro frontend initiated the call and report span only by that instance
- if this is not technically feasible, then there are two options

Option 1: only one faro instance can handle tracing
- this could be either the global instance or the first-registered instance
- we enforce this in the SDK and document the behavior

Option 2: each instance must configure which calls it handles
- this is similar to the ignoreUrls configuration, but could be expanded to "exclude all by default + allowedUrls"

## Actual behavior

Assuming the page has two micro frontends A and B. Faro instance for A is initialized before B. When an HTTP call is made, then app A receives span and app B does not. In addition, configuration from app B is used instead of A, for example custom attributes or ignored URLs.

## Environment

- **SDK version:** 2.1.0
- **SDK instrumentations:** Tracing
- **Device type:** all
- **Device name:** all
- **OS:** all
- **Browser:** all

## Related issues

https://github.com/grafana/faro-web-sdk/issues/568
https://github.com/grafana/faro-web-sdk/issues/1818

pklam:
We need to decide on the current pattern for modeling traces for MFE and might be a small at 1 week for 1 eng. Dependent on the design.

rrmf:
@martinkuba can you please provide a few more details here? Thank you

martinkuba:
Correction on the second issue - it's not just a mismatch in configuration. Each MFE instance creates an additional wrapper around the fetch global function and generates its own span. Detailed explanation:

[...]

PR #1874, opened 2026-02-02
fix(web-tracing): decouple tracing events from global faro instance

### LINKED ISSUE #1847: Fix race condition in tracing instrumentation for faro.tracing.* events
This is specifically only for the faro.tracing.* events generated by the tracing instrumentation, not spans. The tracing instrumentation has other issues specifically when generating spans, but this issue is for the events only.

For more details, see https://github.com/grafana/faro-web-sdk/issues/1845

torresgol10:
## Description

This PR fixes a race condition where `faro.tracing.*` events were always sent via the global `faro` instance (typically the first one initialized) instead of the SDK instance associated with the `FaroTraceExporter`.

This behavior was problematic in micro-frontend architectures with multiple Faro instances (as highlighted in #1845), where events could be attributed to the wrong application or use incorrect instrumentation configurations.

## Changes

- **Refactored `sendFaroEvents`** in `packages/web-tracing/src/faroTraceExporter.utils.ts` to:
    - Eliminate the dependency on the global `faro` object import.
    - Require an explicit `API` instance as a mandatory parameter.
- **Updated `FaroTraceExporter`** in `packages/web-tracing/src/faroTraceExporter.ts` to pass its local `this.config.api` instance when calling `sendFaroEvents`.
- **Updated Tests** to verify that `sendFaroEvents` correctly uses the provided API instance.

## Related Global Issues

- Fixes #1847
- Part of #1845

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (if applicable)
- [x] Verified in local environment

### DIFF
--- a/packages/web-tracing/src/faroMetaAttributesSpanProcessor.test.ts
+++ b/packages/web-tracing/src/faroMetaAttributesSpanProcessor.test.ts
@@ -46,4 +46,65 @@ describe('faroMetaAttributesSpanProcessor', () => {
       'user.hash': 'hash',
     });
   });
+
+  it('support multiple instances having their own metas', () => {
+    const processor1 = new FaroMetaAttributesSpanProcessor(
+      {
+        onStart: jest.fn(),
+        onEnd: jest.fn(),
+        shutdown: jest.fn(),
+        forceFlush: jest.fn(),
+      },
+      {
+        value: {
+          session: {
+            id: 'session-1',
+          },
+        },
+        add: jest.fn(),
+        remove: jest.fn(),
+        addListener: jest.fn(),
+        removeListener: jest.fn(),
+      }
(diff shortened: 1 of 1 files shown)


packages/web-tracing/src/faroTraceExporter.utils.test.ts
```diff
@@ -14,15 +14,15 @@ describe('faroTraceExporter.utils', () => {
     const faro = initializeFaro(mockConfig());
 
     const mockPushEvent = jest.fn();
```
codecapitano, line 16: NIT: remove commented out code :)

eskirk (approved):
very simple solution - I would say we are good to go ahead and merge this. thank you @torresgol10

PR #1876, opened 2026-02-04
feat(instrumentation-replay): make recording conditional on session sampling

d0ugal:
Recording now only starts when the session is included in sampling. This prevents wasting CPU and memory resources recording sessions that will never be sent to the transport.

- Only record when session.attributes.isSampled === 'true'
- Listen for session changes to handle session rotation
- Stop recording when session becomes unsampled

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/experimental/instrumentation-replay/src/instrumentation.test.ts
+++ b/experimental/instrumentation-replay/src/instrumentation.test.ts
@@ -159,6 +159,15 @@ describe('ReplayInstrumentation', () => {
 
     it('should pass default recordAfter option to rrweb record', () => {
       instrumentation = new ReplayInstrumentation();
+
+      // Mock sampled session
+      mockGetSession.mockReturnValue({
+        id: 'test-session',
+        attributes: { isSampled: 'true' },
+      });
+      instrumentation['api'] = { getSession: mockGetSession } as any;
+      instrumentation['metas'] = { addListener: mockAddListener } as any;
+
       instrumentation.initialize();
 
       expect(mockRecord).toHaveBeenCalledWith(


experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -26,12 +26,37 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+    // Check if current session is sampled before starting recording
+    this.checkAndUpdateRecording();
+
+    // Listen for session changes
+    this.metas.addListener(() => {
+      this.checkAndUpdateRecording();
+    });
+  }
+
+  private checkAndUpdateRecording(): void {
```
codecapitano, line 38: 🤔 Thinking about another solution which is to export an observable from the session manager which emits an item including the sampling decision on every sessionID/sampling decision change. So you can just subscribe to session changes instead of listening to all meta updates.

We can add it to the same [place where call `onSessionChange()` ](https://github.com/grafana/faro-web-sdk/blob/c5503f0ee04cafffe1439be6573956decee8ef65/packages/web-sdk/src/instrumentations/session/sessionManager/sessionManagerUtils.ts#L91)

Wdyt @martinkuba

d0ugal: That sound good to me. Is this something I should do? or is there a better approach I can make in the meantime?

codecapitano: Hey @d0ugal it's definitely not a must though it's a clean way to synchronise sessions.
I'd leave it up to you. With the observable you have less calls to `checkAndUpdateRecording`
At the end both solutions are fine. Do you see any other places in session replay where you might want to sync session updates? 

Here's an [example](https://github.com/grafana/faro-web-sdk/blob/a541e09e9df6c7be440e3b11638b61f28241ea40/packages/web-sdk/src/instrumentations/userActions/instrumentation.ts#L20) where we subscribe to an observable form a different Faro package

codecapitano: While thinking a bit more about it. 
Let's stay with the metas listener solution.
The observable might be overkill at this point in time.

PR #1880, opened 2026-02-05
chore(deps): upgrade postgres image to 18.1

torresgol10:
## Why

Upgrade the Postgres database to version 18.1 to ensure the project is using the latest stable release, benefiting from recent performance improvements, security updates, and bug fixes.

## What

- Updated `postgres` service image to version `18.1` in `docker-compose.yaml`.

## Links

- Postgres Image: `postgres:18.1`

## Checklist

- [x] Tests added (Verified locally)
- [ ] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -76,13 +76,14 @@ services:
   database:
     env_file:
       - './.env'
-    image: 'postgres:18'
+    image: 'postgres:18.1@sha256:1090bc3a8ccfb0b55f78a494d76f8d603434f7e4553543d6e807bc7bd6bbd17f'
     volumes:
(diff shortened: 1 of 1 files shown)


eskirk (approved):

codecapitano:
> @codecapitano I think this might be caused by an older Postgres volume. If the data directory was initialized with 17.x, the 18.1 container can fail to start and never opens the port, which would explain the ECONNREFUSED. Please try removing the volume and recreating it.
> 
> On my local machine everything still works with those commands (docker compose up and yarn run start:demo).

Good hint @torresgol10 pruning the volume did the trick and everything works as expected.

PR #1887, opened 2026-02-10
feat: unique error tracking

codecapitano:
## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


packages/core/src/utils/webStorage.ts
```diff
@@ -0,0 +1,84 @@
+import { faro } from '../sdk';
+
```
codecapitano, line 2: This fiel isn't new, we just moved it from the web-sdk to the core package

PR #1896, opened 2026-02-13
feat(instrumentation-replay): add maskInputFn option for custom input masking

Blinkuu:
## Summary

- Add `maskInputFn` option to `ReplayInstrumentationOptions` to allow users to provide a custom function for transforming input values before they are recorded in session replays
- Export the new `MaskInputFn` type from the package
- Pass the option through to rrweb's `record` call
- Update tests and README documentation

## Test plan

- [x] Added test coverage for `maskInputFn` in default options verification
- [x] Added test coverage for custom `maskInputFn` in user-provided options
- [x] Added test verifying `maskInputFn` is passed correctly to rrweb record

### DIFF
--- a/experimental/instrumentation-replay/README.md
+++ b/experimental/instrumentation-replay/README.md
@@ -51,7 +51,10 @@ initializeFaro({
   - `range` - Range inputs
   - `textarea` - Textarea elements
   - `select` - Select dropdowns
-- **`maskInputFn`**: Custom function to transform input values before they are recorded
+- **`maskInputFn`**: Custom function to transform input values before they are recorded.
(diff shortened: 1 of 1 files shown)


experimental/instrumentation-replay/src/instrumentation.test.ts
```diff
@@ -61,13 +62,15 @@ describe('ReplayInstrumentation', () => {
 
     it('should use custom options when provided', () => {
       const beforeSendFn = jest.fn();
+      const maskInputFn: MaskInputFn = jest.fn((text, _element) => '*'.repeat(text.length));
```
codecapitano, line 65: TIL `String.repeat` 💡

PR #2005, opened 2026-04-22
fix(web-tracing): prevent undefined faro.api access error

### LINKED ISSUE #2004: can't access property "getActiveUserAction"
## Description

> TypeError: can't access property "getActiveUserAction", OA.api is undefined

## Environment

- **SDK version:** 2.3.1
- **SDK instrumentations:** Web SDK, Web Tracing, React
- **Device type:** desktop
- **Device name:** MacBook Pro 13"
- **OS:** macOS 26.3.1
- **Browser:** Chrome 147

## Demo

Image

## Context

[const currentAction = faro.api.getActiveUserAction();](https://github.com/grafana/faro-web-sdk/blob/868a5b7de1fa86b7d9e52507d4da5f95b2d10e1f/packages/web-tracing/src/faroXhrInstrumentation.ts#L37-L38)

`faro.api` – `api` is typed as any here and it seems that it can also be undefined. There should probably be a nullish check before accessing the `getActiveUserAction()` methdo from it.

Pascalmh:
## Why

`faro.api` is typed as `any` and can be nullish. This leads to a runtime error.

Fixes #2004

## What

Use optional chaining to safely handle cases where faro.api is undefined when accessing getActiveUserAction() method.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


martinkuba:
This should no longer be needed since https://github.com/grafana/faro-web-sdk/pull/2009 has been merged.

codecapitano:
Thanks a lot for taking the time to put this fix together, @Pascalmh — contributions like this are really appreciated!

As @martinkuba noted, the underlying issue is now covered by #2009, which provides a no-op `faro.api` default at module load so pre-initialization calls no longer throw. Since that change makes the optional-chaining guard here redundant, I'll go ahead and close this PR.

Issue #2004 should already be resolved on `main`. Thanks again for the contribution, and we hope to see more from you in the future!

PR #2014, opened 2026-04-28
feat(web-sdk): add `httpHost` to `faro.performance.resource` events

gouthamve-but-enterprise:
## Summary
- Adds an `httpHost` attribute to `faro.performance.resource` events so consumers can group/filter by destination host without re-parsing the URL.
- Mirrors the `http.host` attribute already emitted on `faro.tracing.fetch` events (sourced from OpenTelemetry's fetch instrumentation), giving the two event types a consistent dimension.
- `faro.performance.navigation` events inherit `httpHost` via `createFaroResourceTiming`.

The host is extracted using `new URL(name).host`, with an `unknown` fallback for unparseable URLs (e.g., `data:` / `blob:` / malformed entries).

## Test plan
- [x] `cd packages/web-sdk && npx jest --testPathPatterns=performance` — 30 passed
- [x] `cd packages/web-sdk && npx jest` — 242 passed, 1 skipped (pre-existing)
- [x] Prettier check on changed files passes
- [ ] Manual: load the demo app, confirm `faro.performance.resource` and `faro.tracing.fetch` events for the same fetch carry matching `http.host` values
- [ ] Manual: trigger a `data:`/`blob:` resource load and confirm the event has `http.host: 'unknown'` rather than throwing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
+++ b/packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
@@ -97,6 +97,7 @@ export function createFaroResourceTiming(resourceEntryRaw: PerformanceResourceTi
 
   return {
     name: name,
+    'http.host': getHostFromUrl(name),
     duration: toFaroPerformanceTimingString(duration),
     tcpHandshakeTime: toFaroPerformanceTimingString(connectEnd - connectStart),
     dnsLookupTime: toFaroPerformanceTimingString(domainLookupEnd - domainLookupStart),
@@ -181,6 +182,14 @@ function getDocumentParsingTime(): number | null {
   }
 
   return null;
+}
+
+function getHostFromUrl(url: string): string {
+  try {
+    return new URL(url).host || unknownString;
+  } catch {
+    return unknownString;
+  }
 }
 
 function toFaroPerformanceTimingString(v: unknown): string {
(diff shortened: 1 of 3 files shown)


kpelelis (dismissed):
LGTM 🚀 I would probably add the testcase the 🤖 mentioned but feel free to ignore

packages/web-sdk/src/instrumentations/performance/performanceUtils.ts
```diff
@@ -97,6 +97,7 @@ export function createFaroResourceTiming(resourceEntryRaw: PerformanceResourceTi
 
   return {
     name: name,
+    'http.host': getHostFromUrl(name),
```
martinkuba: I suggest we use `httpHost` to be consistent with the other attributes

codecapitano (changes requested):
As per comments from Martin and Kostas

gouthamve-but-enterprise:
Thanks for the review! Pushed `4c3da3cd`:

- Renamed `'http.host'` → `httpHost` to match the rest of `FaroResourceTiming` (per @martinkuba).
- Added `data:` and `blob:` URL cases to the `httpHost` test, asserting the `'unknown'` fallback (per Copilot / @kpelelis).

codecapitano (approved):

PR #2015, opened 2026-04-29
feat(replay): pause recording after user inactivity

d0ugal:
## Summary

- Adds an `inactivityThresholdMs` option to `ReplayInstrumentation` that pauses rrweb recording when no user interaction is detected for the configured duration (default: **60 seconds**).
- Recording resumes automatically with a fresh rrweb checkpoint (full DOM snapshot) when the user interacts again, so the player has complete state to resume from.
- Prevents long idle browser tabs from generating hours of empty DOM-mutation-only recordings that have no replay value and waste bandwidth/storage.

## Motivation

The existing rrweb `getInactivePeriods` algorithm only records gaps *between* user interactions — it never detects trailing inactivity. When a user stops interacting but leaves a tab open, the browser continues generating DOM mutations (IncrementalSnapshot source=0) indefinitely with zero replay value. This fix stops recording during those idle periods and resumes with a fresh checkpoint when the user returns.

## How it works

1. When recording starts, DOM event listeners are registered for `pointermove`, `pointerdown`, `scroll`, `keydown`, and `input` (capture phase, passive).
2. A `setTimeout` fires after `inactivityThresholdMs` of no interaction → calls rrweb `stop()`.
3. DOM listeners stay active. When the user interacts → calls rrweb `record()` again (fresh checkpoint), resets timer.
4. Emits `faro.session_recording.paused` / `faro.session_recording.resumed` events for observability.

## Usage

```ts
// Default: pauses after 60 seconds of no interaction
new ReplayInstrumentation()

// Custom threshold
new ReplayInstrumentation({
  inactivityThresholdMs: 120_000, // 2 minutes
})

// Disable inactivity pausing (record continuously)
new ReplayInstrumentation({
  inactivityThresholdMs: 0,
})
```

## Test plan

- [x] All 29 existing tests pass
- [ ] Manual test: verify recording pauses after threshold with no interaction
- [ ] Manual test: verify recording resumes with fresh checkpoint on interaction
- [ ] Verify paused/resumed events appear in Faro telemetry

## Related

- Companion backend PR: [feat(replay): detect trailing inactivity in manifest](https://github.com/grafana/app-o11y-kwl-endpoint/pull/1367)

### DIFF
--- a/experimental/instrumentation-replay/src/const.ts
+++ b/experimental/instrumentation-replay/src/const.ts
@@ -17,4 +17,5 @@ export const defaultReplayInstrumentationOptions: ReplayInstrumentationOptions =
   beforeSend: undefined,
   recordAfter: 'load',
   samplingRate: 1,
+  inactivityThresholdMs: 60_000,
 };
--- a/experimental/instrumentation-replay/src/instrumentation.ts
+++ b/experimental/instrumentation-replay/src/instrumentation.ts
@@ -8,6 +8,20 @@ import type { ReplayInstrumentationOptions } from './types';
 
 const faroSessionReplayEventName = 'faro.session_recording.event';
 const faroSessionReplayStartedEventName = 'faro.session_recording.started';
+const faroSessionReplayPausedEventName = 'faro.session_recording.paused';
+const faroSessionReplayResumedEventName = 'faro.session_recording.resumed';
+
+// DOM events that signal a human is present.  Aligned with rrweb's
+// IncrementalSource 1-5 (MouseMove, MouseInteraction, Scroll,
+// ViewportResize, Input).  We use pointer* instead of mouse*/touch*
+// because modern browsers fire PointerEvents for all input devices.
+const USER_INTERACTION_EVENTS: readonly string[] = [
+  'pointermove', // MouseMove / TouchMove
+  'pointerdown', // MouseInteraction (click, dblclick, etc.)
+  'scroll', // Scroll
+  'keydown', // Input
+  'input', // Input (covers typing without keydown, e.g. autofill)
+];
 
 export class ReplayInstrumentation extends BaseInstrumentation {
   readonly name = '@grafana/faro-instrumentation-replay';
@@ -15,7 +29,10 @@ export class ReplayInstrumentation extends BaseInstrumentation {
 
   private stopFn: { (): void } | null = null;
   private isRecording: boolean = false;
+  private isPaused: boolean = false;
   private options: ReplayInstrumentationOptions = defaultReplayInstrumentationOptions;
+  private inactivityTimer: ReturnType<typeof setTimeout> | null = null;
+  private boundOnUserInteraction: (() => void) | null = null;
 
   constructor(options: ReplayInstrumentationOptions = {}) {
     super();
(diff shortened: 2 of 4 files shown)


experimental/instrumentation-replay/src/const.ts
```diff
@@ -17,4 +17,5 @@ export const defaultReplayInstrumentationOptions: ReplayInstrumentationOptions =
   beforeSend: undefined,
   recordAfter: 'load',
   samplingRate: 1,
+  inactivityThresholdMs: 60_000,
```
codecapitano, line 20: Suggest lowering the default from 60s to **15–30s**. Rationale:

- Human users rarely sit completely idle for >30s while genuinely engaged with a page; the typical "reading" pause is 5–20s. 60s leaves a long window where animations, polling spinners, and DOM mutations from background tasks still produce zero-replay-value events.
- Industry defaults: PostHog uses 30s, Sentry's session-replay uses 15s of inactivity to end a segment, LogRocket uses ~30s. 60s is on the high end.
- Users who want longer windows can opt in; it's an explicit option.

If you want to keep 60s for safety on the first release, please leave a comment here recording the trade-off so it's easy to revisit.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -102,52 +119,151 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+
+      this.isPaused = false;
+      this.logDebug('Session replay resumed after user interaction');
+      this.api.pushEvent(faroSessionReplayResumedEventName, {});
+
+      this.resetInactivityTimer();
+    } catch (err) {
+      this.logWarn('Failed to resume session replay', err);
+    }
+  }
```
codecapitano, line 220: **No test coverage for the pause→resume cycle.** The logic is correct on inspection (`pauseRecording` keeps DOM listeners attached, the bound handler routes to `resumeRecording` when `isPaused`, `record()` is called again to take a fresh checkpoint) — but a future refactor could break this without any test failing.

Minimum coverage I'd expect:

```ts
it('pauses recording after inactivity threshold elapses', () => { /* fake timers + assert stopFn called + paused event pushed */ });
it('resumes recording with a fresh checkpoint when user interacts after pause', () => { /* dispatch pointerdown + assert record() called twice + resumed event pushed */ });
it('does not pause when inactivityThresholdMs is 0', () => { /* … */ });
it('does not pause when inactivityThresholdMs is undefined', () => { /* … */ });
it('removes DOM listeners on stopRecording', () => { /* assert listeners gone */ });
it('keeps DOM listeners attached across pauseRecording', () => { /* assert dispatch still triggers resume */ });
```

Using `jest.useFakeTimers()` and `document.dispatchEvent(new Event('pointerdown'))` (jsdom env) makes these straightforward — no rrweb internals needed since `record` is already mocked.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -102,52 +119,151 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+  private pauseRecording(): void {
+    if (!this.isRecording || this.isPaused) {
+      return;
+    }
+
+    this.stopRrweb();
+    this.isPaused = true;
+    this.logDebug('Session replay paused due to inactivity');
+    this.api.pushEvent(faroSessionReplayPausedEventName, {});
+  }
```
codecapitano, line 202: Minor: `pauseRecording` doesn't clear `inactivityTimer`. Today this is harmless because the only caller is the timer itself (which has already fired), but if anyone later adds a manual pause path (e.g., `pause()` API, `visibilitychange`-driven pause), the orphan timer will refire and call `pauseRecording` while already paused (the `isPaused` guard at line 189 makes it a no-op, so it's defensively safe — but cleaner to clear the timer here).

Suggested:

```ts
private pauseRecording(): void {
  if (!this.isRecording || this.isPaused) {
    return;
  }

  if (this.inactivityTimer !== null) {
    clearTimeout(this.inactivityTimer);
    this.inactivityTimer = null;
  }

  this.stopRrweb();
  this.isPaused = true;
  this.logDebug('Session replay paused due to inactivity');
  this.api.pushEvent(faroSessionReplayPausedEventName, {});
}
```

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -8,14 +8,31 @@ import type { ReplayInstrumentationOptions } from './types';
...
+// IncrementalSource 1-5 (MouseMove, MouseInteraction, Scroll,
+// ViewportResize, Input).  We use pointer* instead of mouse*/touch*
+// because modern browsers fire PointerEvents for all input devices.
+const USER_INTERACTION_EVENTS: readonly string[] = [
+  'pointermove', // MouseMove / TouchMove
+  'pointerdown', // MouseInteraction (click, dblclick, etc.)
+  'scroll', // Scroll
+  'keydown', // Input
+  'input', // Input (covers typing without keydown, e.g. autofill)
+];
```
codecapitano, line 24: Worth considering whether `visibilitychange` should also act as a wake/sleep signal. If the tab is hidden (background) we probably want to pause regardless of pointer activity, and on `visibilitychange → visible` we likely want to resume (with a fresh checkpoint, same as the pointer path). Out of scope for this PR if you'd rather ship the simpler version first, but worth filing a follow-up.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -102,52 +119,151 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+      if (this.isPaused) {
+        this.resumeRecording();
+      } else {
+        this.resetInactivityTimer();
+      }
+    };
+
+    for (const eventName of USER_INTERACTION_EVENTS) {
+      document.addEventListener(eventName, this.boundOnUserInteraction, { capture: true, passive: true });
+    }
```
codecapitano, line 238: Consider adding a brief comment explaining the listener-options choice:

```ts
// capture: true so we observe interaction even if a child handler calls stopPropagation()
// passive: true so this listener cannot block scrolling (required by Chrome for scroll/touch listeners)
```

experimental/instrumentation-replay/src/types.ts
```diff
@@ -80,6 +80,19 @@ export interface ReplayInstrumentationOptions {
...
+   * (mouse moves, clicks, scrolls, viewport resizes, or input).
+   *
+   * When the user interacts again, recording resumes automatically with a
+   * fresh rrweb checkpoint so the player has a complete DOM snapshot.
+   *
+   * Set to 0 or undefined to disable (record continuously).
+   *
+   * @default 60000 (1 minute)
+   */
+  inactivityThresholdMs?: number;
```
codecapitano, line 94: Doc nit: "Set to 0 or undefined to disable" — the code path treats negative values as disabled too (`!threshold || threshold <= 0`). Either reflect that in the docs or validate/clamp explicitly with a `logWarn` (matching how `samplingRate` is handled). I'd lean toward validation since `inactivityThresholdMs: -1` silently meaning "disabled" is a footgun.

codecapitano (commented):
Thanks for tackling trailing inactivity — the resume-with-checkpoint design is correct and will meaningfully reduce idle-tab payload. A few items to consider before merging.

**Higher-level questions**

1. **60s default feels too long.** With a 5-minute checkpoint cadence (`checkoutEveryNms: 300_000`) and rrweb's own `MouseMove` throttle, a fully idle tab still emits very few events per minute, but a *partially* idle tab (user reading, no pointer movement) generates DOM-mutation noise from animations/spinners/timers for the full 60s before pause kicks in. A more typical value in the industry (PostHog, LogRocket, Sentry) is 10–30s. I would suggest defaulting to **15–30s** and exposing the option as a clear knob; teams that want "forever" can set 0. See inline comment.

2. **Resume path looks correct, but is untested.** I traced it: `pauseRecording` deliberately does *not* tear down the DOM listeners (good), so on the next interaction the bound listener fires `resumeRecording()` which calls `record()` again — rrweb takes a fresh full-DOM checkpoint, as advertised. However there is no unit test covering the pause→resume cycle, so a future refactor could silently break it. See inline comment requesting a test.

**Other findings**

- Event listeners are correctly attached with `{ capture: true, passive: true }`. Worth a brief comment explaining the choice (capture phase to defeat third-party `stopPropagation`, passive to avoid scroll-jank warnings).
- `pointermove` fires very frequently — the handler is cheap (one `clearTimeout` + one `setTimeout`) so this is fine, but worth a comment noting the listener is intentionally hot.
- The PR body checklist still has unchecked manual-test items. Were those exercised? If yes, please tick. If not, the pause→resume unit test should at minimum be added before merge.

Overall direction is good; requesting changes only on the missing test and the default value, both small.

experimental/instrumentation-replay/src/types.ts
codecapitano: Retracting — this is a nit, not blocking. Resolving.

experimental/instrumentation-replay/src/instrumentation.ts
codecapitano: Retracting — this is a nit, not blocking. Resolving.

codecapitano (approved):

experimental/instrumentation-replay/src/const.ts
d0ugal: I think 60s is a reasonable starting default, though I agree it's still a guess. We should aim to optimise it over time

The key tradeoff is that each resume requires a full DOM checkpoint (rrweb re-calls record() to capture a complete snapshot), which is the single most expensive payload type. A lower threshold like 15-30s means a user who reads for
20s, scrolls, reads for 20s again would trigger two full checkpoints instead of zero. For pages with large DOMs each checkpoint can be 100KB+, so pausing too aggressively could actually increase total bandwidth.

I am going to work on some metrics to better inform us and we can optimise this.

PR #2028, opened 2026-05-05
feat(web-sdk): add gzip request compression to FetchTransport

d0ugal:
## Summary

- Adds `requestCompression` option to `FetchTransport` that gzip-compresses request bodies using the native `CompressionStream` API
- Defaults to `false` — opt-in to avoid breaking self-hosted collectors or custom endpoints that don't handle gzip request bodies
- Falls back to uncompressed with a warning when `CompressionStream` is unavailable (older browsers)
- Zero bundle size cost — uses native browser API, no new dependencies

## Compression results

Tested across 17 example sites spanning Angular, React, Vue, Svelte, Next.js, Nuxt, Remix, and Lit. Each site ran identical 60-second deterministic browsing sessions — once with `CompressionStream` disabled (baseline) and once with gzip enabled. A logging proxy captured every `/collect/` payload to compare sizes.

**78% median payload reduction (range: 55–86%)**

Most frameworks clustered around 75–82% savings. Sites with larger DOM trees (e.g. Ant Design) saw up to 86%. The lowest was 55% (Vuetify). Canvas-heavy sites with base64 rrweb snapshots saw 93%+ reduction but were excluded as outliers.

> **Note:** This testing was with session replay enabled, which produces the largest Faro payloads. Regular Faro payloads (logs, errors, measurements, traces) will also benefit from compression but are much smaller, so the absolute savings will be less significant.

## How to enable

The Grafana Cloud Faro collector already supports `Content-Encoding: gzip` on incoming requests, so cloud users can enable this immediately:

```ts
initializeFaro({
  url: '...',
  requestCompression: true,
});
```

Self-hosted collectors or custom endpoints should verify they handle gzip request bodies before enabling.

## How it works

After `JSON.stringify()` produces the body, if compression is enabled the body is piped through `CompressionStream('gzip')` and sent as a `Blob` with `Content-Encoding: gzip`. The `Blob.size` is used for the keepalive threshold check (60KB), meaning compressed payloads are more likely to qualify for keepalive — a side benefit for page-unload reliability.

### DIFF
--- a/packages/web-sdk/src/transports/fetch/transport.ts
+++ b/packages/web-sdk/src/transports/fetch/transport.ts
@@ -31,7 +31,7 @@ export class FetchTransport extends BaseTransport {
     this.rateLimitBackoffMs = options.defaultRateLimitBackoffMs ?? DEFAULT_RATE_LIMIT_BACKOFF_MS;
     this.getNow = options.getNow ?? (() => Date.now());
 
-    const requestCompression = options.requestCompression ?? true;
+    const requestCompression = options.requestCompression ?? false;
 
     if (requestCompression && typeof CompressionStream === 'undefined') {
       this.compressionEnabled = false;
--- a/packages/web-sdk/src/transports/fetch/transport.ts
+++ b/packages/web-sdk/src/transports/fetch/transport.ts
@@ -35,7 +35,9 @@ export class FetchTransport extends BaseTransport {
 
     if (requestCompression && typeof CompressionStream === 'undefined') {
       this.compressionEnabled = false;
-      this.logWarn('requestCompression is enabled but CompressionStream is not available. Falling back to uncompressed.');
+      this.logWarn(
+        'requestCompression is enabled but CompressionStream is not available. Falling back to uncompressed.'
+      );
     } else {
       this.compressionEnabled = requestCompression;
     }
--- a/packages/web-sdk/src/transports/fetch/transport.ts
+++ b/packages/web-sdk/src/transports/fetch/transport.ts
@@ -31,7 +31,7 @@ export class FetchTransport extends BaseTransport {
     this.rateLimitBackoffMs = options.defaultRateLimitBackoffMs ?? DEFAULT_RATE_LIMIT_BACKOFF_MS;
     this.getNow = options.getNow ?? (() => Date.now());
 
-    const requestCompression = options.requestCompression ?? false;
+    const requestCompression = options.requestCompression ?? true;
 
     if (requestCompression && typeof CompressionStream === 'undefined') {
       this.compressionEnabled = false;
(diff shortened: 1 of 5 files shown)


eskirk (commented):
I think this is an awesome change - my one thought is making sure this does not break OSS users if they upgrade and now their ingestion is broken since the receiver of their choice does not support compression.

once we verify that the faro components that we have in otel contrib would also be fine with this change, I say we go through with it ¯\\\_(ツ)\_/¯

d0ugal:
Support has been added to alloy and I have made compression optional. So I think this is ready for review.

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -130,6 +153,26 @@ export class FetchTransport extends BaseTransport {
     return new Date(now + this.rateLimitBackoffMs);
   }
 
+  private async compress(body: string): Promise<Blob> {
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode(body));
+        controller.close();
+      },
+    }).pipeThrough(new CompressionStream('gzip'));
```
kpelelis, line 162: Note that the collector also supports `brotli` and `deflate`. If you want to use a default, I suggest [Brotli](https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream/CompressionStream#format), or giving the option to the user to select compression algorithm

kpelelis (approved):
Overall LGTM. You could take a look for other compression algorithms that the cloud collector supports

packages/web-sdk/src/transports/fetch/transport.ts
d0ugal: I've made the algorithm configurable. `requestCompression` now accepts `true` (defaults to `gzip`), or a string: `'gzip'`, `'deflate'`, or `'brotli'`.

Regarding algorithm choice and browser support for `CompressionStream`:

| Algorithm | Chrome | Firefox | Safari | `Content-Encoding` |
|-----------|--------|---------|--------|--------------------|
| `gzip` | 80+ | 113+ | 16.4+ | `gzip` |
| `deflate` | 80+ | 113+ | 16.4+ | `deflate` |
| `brotli` | **Not supported** | 147+ | 18.4+ | `br` |

`brotli` typically compresses ~15-20% better than gzip, but Chrome/Chromium still doesn't support it in `CompressionStream` (as of mid-2026, covering ~65% of browser traffic). So `gzip` remains the safe default — `true` maps to `gzip`. Users on Firefox/Safari-heavy audiences can opt into `'brotli'`, and it'll just work when Chrome eventually adds support.

The existing `CompressionStream` availability fallback still applies: if the browser doesn't support `CompressionStream` at all (or doesn't support the chosen algorithm), it falls back to uncompressed with a warning.

PR #2039, opened 2026-05-08
feat(core): include git commit hash in app meta

eskirk:
## Summary

- Adds `getGitHash(appName)` to `core/src/utils/sourceMaps.ts` — reads `window.__faroGitHash_<appName>` from the global object, mirrors the existing `getBundleId` pattern exactly
- Adds `gitHash?: string` to `MetaApp` in `core/src/metas/types.ts`
- `registerInitialMetas` now calls `getGitHash` and sets `app.gitHash`, so every telemetry payload carries the commit hash

The git hash itself is injected at build time by the bundler plugins. Companion PR: https://github.com/grafana/faro-javascript-bundler-plugins/pull/533

## Test plan

- [ ] Initialize Faro in an app built with the companion bundler plugin PR — confirm `app.gitHash` is present in the meta and appears in telemetry payloads
- [ ] Initialize Faro in an app without the bundler plugin (no `window.__faroGitHash_*`) — confirm `app.gitHash` is `undefined` and no error is thrown

### DIFF
--- a/packages/core/src/metas/initialize.test.ts
+++ b/packages/core/src/metas/initialize.test.ts
@@ -21,7 +21,9 @@ describe('metas', () => {
   it('preserves config.app.gitHash when global __faroGitHash is absent', () => {
     delete (global as any).__faroGitHash_test;
 
-    const { metas } = initializeFaro(mockConfig({ app: { name: 'test', version: '1.0.0', gitHash: 'deadbeefdeadbeefdeadbeefdeadbeefdeadbeef' } }));
+    const { metas } = initializeFaro(
+      mockConfig({ app: { name: 'test', version: '1.0.0', gitHash: 'deadbeefdeadbeefdeadbeefdeadbeefdeadbeef' } })
+    );
     expect(metas.value.app?.gitHash).toEqual('deadbeefdeadbeefdeadbeefdeadbeefdeadbeef');
   });
 
--- a/packages/core/src/metas/initialize.test.ts
+++ b/packages/core/src/metas/initialize.test.ts
@@ -2,6 +2,22 @@ import { initializeFaro } from '../initialize';
 import { mockConfig } from '../testUtils';
 
 describe('metas', () => {
+  it('sets app.gitHash from global object on initialization', () => {
+    (global as any).__faroGitHash_test = 'abc123def456abc123def456abc123def456abc1';
+
+    const { metas } = initializeFaro(mockConfig());
+    expect(metas.value.app?.gitHash).toEqual('abc123def456abc123def456abc123def456abc1');
+
+    delete (global as any).__faroGitHash_test;
+  });
+
+  it('leaves app.gitHash undefined when global is not set', () => {
+    delete (global as any).__faroGitHash_test;
+
+    const { metas } = initializeFaro(mockConfig());
+    expect(metas.value.app?.gitHash).toBeUndefined();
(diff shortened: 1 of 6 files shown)


packages/core/src/metas/registerInitial.ts
```diff
@@ -1,17 +1,21 @@
...
   const initial: Meta = {
     sdk: {
       version: VERSION,
       name: 'faro',
     },
     app: {
-      bundleId: faro.config.app.name && getBundleId(faro.config.app.name),
+      bundleId: appName && getBundleId(appName),
+      ...(gitHash !== undefined ? { gitHash } : {}),
     },
```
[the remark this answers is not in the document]
kpelelis: I think this is a fair point

packages/core/src/metas/registerInitial.ts
```diff
@@ -1,17 +1,21 @@
...
+
   const initial: Meta = {
     sdk: {
       version: VERSION,
       name: 'faro',
     },
     app: {
-      bundleId: faro.config.app.name && getBundleId(faro.config.app.name),
+      bundleId: appName && getBundleId(appName),
+      ...(gitHash !== undefined ? { gitHash } : {}),
```
codecapitano, line 18: For the spec record: the conditional spread for `gitHash` is actually a behavioral fix vs. the bundleId pattern, not a divergence. Line 28 spreads `initial.app` last, so the always-set `bundleId: appName && getBundleId(appName)` (which evaluates to `undefined` when the global is absent) clobbers a user-supplied `config.app.bundleId`. The conditional spread for `gitHash` here dodges the clobber.

`design.md` Decision 5 ("direct copy of the bundleId pattern with no new abstractions") is therefore stale. Two options:
1. Update Decision 5 to acknowledge the asymmetry is intentional.
2. Fix `bundleId` to use the same conditional-spread shape (out of scope here — follow-up).

No change requested in this PR; flagging so the spec gets corrected.

packages/core/src/metas/initialize.test.ts
```diff
@@ -2,6 +2,22 @@ import { initializeFaro } from '../initialize';
...
+
+    delete (global as any).__faroGitHash_test;
+  });
+
+  it('leaves app.gitHash undefined when global is not set', () => {
+    delete (global as any).__faroGitHash_test;
+
+    const { metas } = initializeFaro(mockConfig());
+    expect(metas.value.app?.gitHash).toBeUndefined();
+  });
```
codecapitano, line 19: Worth adding one test asserting that user-supplied `config.app.gitHash` is preserved when the global `__faroGitHash_<app>` is absent — that's the whole point of the conditional-spread fix in this PR and currently isn't asserted anywhere.

Sketch (adjust to your `mockConfig` signature):

```ts
it('preserves config.app.gitHash when global is absent', () => {
  delete (global as any).__faroGitHash_test;
  const { metas } = initializeFaro(mockConfig({ app: { name: 'test', gitHash: 'user-set' } }));
  expect(metas.value.app?.gitHash).toEqual('user-set');
});
```

codecapitano (commented):
Conditional spread for `gitHash` is correct — it's a behavioral fix vs. the bundleId pattern, not a divergence. Inline note on what this means for `design.md` Decision 5, plus one suggested test.

codecapitano (approved):

Issue #2072, opened 2026-05-11
Investigate lower-priority transport scheduling to reduce main thread impact


d0ugal:
## Summary

The Faro transport layer currently runs all work — batching, serialization, and `fetch()` calls — on the main thread with a simple `setInterval`-based batch timer. This can impact website performance, particularly when **session replay (rrweb) is enabled**, which generates high-frequency, heavy payloads that share the same transport pipeline as regular telemetry.

The goal is to minimize the performance impact of Faro instrumentation on the host website.

## Why Replay Makes This Urgent

Session replay generates significantly more transport load than standard telemetry:

- **High frequency**: every mouse move, DOM mutation, scroll, click, and input change produces an rrweb event
- **Heavy payloads**: full DOM snapshots (every 5 min or on resume from inactivity), plus incremental mutation records
- **Double serialization on main thread**: each rrweb event is `JSON.stringify()`-ed individually in `handleEvent()`, then the entire batch is `JSON.stringify()`-ed again in `FetchTransport.send()` — this is the primary CPU cost
- **Shared pipeline**: replay events flow through the exact same `BatchExecutor → FetchTransport` path as logs, errors, and measurements — no separate channel, no different priority
- **Same batch settings**: replay uses the same 250ms/50-item batch config as lightweight telemetry, which may not be optimal for replay's burst patterns

## Current Architecture

```
rrweb record() emit callback
    → JSON.stringify(event)                          ← serialization #1 (main thread)
    → api.pushEvent('faro.session_recording.event')
    → BatchExecutor (250ms / 50 items, setInterval)
    → FetchTransport.send()
        → JSON.stringify(getTransportBody(items))    ← serialization #2 (main thread)
        → fetch() with keepalive
```

Key files:
- `experimental/instrumentation-replay/src/instrumentation.ts` — rrweb integration, event handling, first serialization
- `packages/core/src/transports/batchExecutor.ts` — item buffering and time/size-based flush
- `packages/web-sdk/src/transports/fetch/transport.ts` — HTTP transport via fetch
- `packages/core/src/utils/promiseBuffer.ts` — concurrent request management

### 1. Web Worker Transport (Recommended Focus)

Move serialization and `fetch()` calls to a dedicated Web Worker. This is the strongest option for replay because it **eliminates** main-thread CPU cost rather than just deferring it.

```
Main thread                              Worker thread
rrweb emit →  postMessage(event)  →      buffer events
                                         JSON.stringify (both passes)
[...]

PR #2073, opened 2026-05-11
feat(web-sdk): offload transport to Web Worker

d0ugal:
## Summary

Moves JSON serialization and `fetch()` calls off the main thread by running them inside an inline Web Worker. This reduces main-thread jank from telemetry transport, especially when **session replay** is enabled (high-frequency rrweb events with heavy payloads that are double-serialized).

- **Opt-in**: Set `enableWorker: true` in `FetchTransportOptions` to activate the worker path. Default behavior is unchanged.
- **Worker path**: `BatchExecutor` flushes items → main thread resolves dynamic headers and posts items to Worker via `postMessage` (structured clone) → Worker runs `getTransportBody()`, `JSON.stringify()`, and `fetch()`
- **Automatic fallback**: If `Worker` isn't available (SSR, CSP `worker-src` restrictions, old browsers), or the Worker crashes at runtime, all calls route through `sendDirect()` — the exact same code path as before this change
- **Unload safety**: When `document.visibilityState === 'hidden'` (visibilitychange flush), bypasses the Worker and sends directly with `keepalive: true` so data survives page death

### Architecture

```
Main thread                                Worker thread
────────────────────────────────           ─────────────────────
BatchExecutor flush                        
  → resolve async headers                  
  → postMessage({items, url, ...})    →    getTransportBody(items)
                                           JSON.stringify(body)
                                           fetch(url, {body, headers})
                                           handle 429 / rate limiting
  ← postMessage({type, status})       ←    
```

### Key decisions

- Worker is **opt-in** via `enableWorker: true` — existing users are not affected until they explicitly enable it
- Worker script is **inlined as a Blob URL** — no separate file to load, no build pipeline changes
- `getTransportBody()` and `mergeResourceSpans()` are **duplicated** inside the worker (they're ~50 lines of pure logic with no deps)
- `AbortSignal` is stripped from `requestOptions` before `postMessage` (not structured-cloneable) — acceptable since abort signals for telemetry are extremely uncommon
- `PromiseBuffer` stays on the main thread for backpressure; each worker message gets a unique ID to match responses
- Relative collector URLs (e.g. `/collect/...`) are resolved to absolute via `new URL(url, document.baseURI)` before posting to the Worker, since blob: workers have no page origin to resolve against

### CSP considerations

The inline Worker is created from a `blob:` URL. Sites with a Content-Security-Policy need `worker-src 'self' blob:;` (or equivalent) for the worker path to activate. When CSP blocks the worker, the fallback to `sendDirect()` activates silently — no errors, no data loss.

### Backwards compatibility

Zero breaking changes:
- Worker is opt-in (`enableWorker: true`), so default behavior is identical to before
- Fallback to identical behavior when Workers unavailable
- `visibilitychange` flush path unchanged (direct fetch + keepalive)
- Worker crash → automatic recovery to main-thread path

## Test plan

- [ ] Verify existing FetchTransport tests pass
- [ ] TypeScript compilation passes (`tsc --build` for both CJS and ESM)
- [ ] Manual testing in browser: verify telemetry requests originate from Worker thread (DevTools Network → Initiator column shows "worker" or empty `Referer` header)
- [ ] Manual testing: verify fallback works when CSP blocks blob: workers
- [ ] Manual testing with replay enabled: verify reduced main-thread long tasks during active recording
- [ ] Test page unload scenario: verify data is delivered via direct path when tab is closed

Relates to #2072


d0ugal:
## Benchmark Validation

Tested worker vs direct transport using a stress-test benchmark site (100×12 table with nested elements, ~50% cell updates at 60fps, continuous API calls to a Go backend, rrweb replay enabled). 5 runs per mode with 20× CPU throttle in Chrome DevTools.

**App API latency (averages across 5 runs):**

| Metric | Worker | Direct | Delta |
|--------|:------:|:------:|:-----:|
| p50 | 749 ms | 912 ms | **-18%** |
| p95 | 1,128 ms | 1,702 ms | **-34%** |
| Mean | 739 ms | 942 ms | **-22%** |
| Max | 1,235 ms | 1,902 ms | **-35%** |

Long task count and total blocking time were roughly the same (~86 vs ~88 long tasks) — the DOM mutation work dominates regardless of transport mode. The worker's benefit is preventing `JSON.stringify` + `fetch` from competing with the app's own network calls on the main thread, which shows up clearly in reduced tail latency spikes.

On unthrottled fast hardware (M5 Mac) there was no measurable difference — the CPU is fast enough that serialization completes instantly either way. The improvement targets slower/mobile devices where rrweb payload serialization can block the main thread.

packages/web-sdk/src/transports/fetch/transportWorker.ts
```diff
@@ -0,0 +1,217 @@
+// This file runs inside a Web Worker (blob URL).
+// It MUST be self-contained — no imports from the main module graph.
```
domasx2, line 2: why not? Rollup will compile it to a single file anyway. Seems like avoiding code copy/paste is a good idea

packages/web-sdk/src/transports/fetch/types.ts
```diff
@@ -30,6 +30,8 @@ export interface FetchTransportOptions {
   // compress request bodies with gzip using the native CompressionStream API.
   // falls back to uncompressed if CompressionStream is not available.
   requestCompression?: boolean;
+  // offload JSON serialization and fetch to a Web Worker
+  enableWorker?: boolean;
```
domasx2, line 34: I would suggest to create a separate `WorkerFetchTransport` using composition or inheritance instead of baking it into `FetchTransport`, wdyt

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -46,82 +64,238 @@ export class FetchTransport extends BaseTransport {
...
+      // Use worker when available and page is visible.
+      // When the page is hidden (visibilitychange flush), send directly
+      // so keepalive/sendBeacon can deliver before the page dies.
+      // Also bypass when an AbortSignal is present since it can't be cloned to a worker.
+      const hasSignal = !!(this.options.requestOptions as RequestInit | undefined)?.signal;
+      if (this.worker && document.visibilityState !== 'hidden' && !hasSignal) {
+        try {
+          return await this.sendViaWorker(items);
+        } catch {
+          // Worker path failed — fall through to direct send
```
domasx2, line 163: hm, possible double-send in this case, don't know for sure if error prevented sending

domasx2 (commented):
Hey! I think ths is a great feature to explore, a few concerns though

kpelelis:
Hey @d0ugal, is this something that you still want to pursue?

PR #2094, opened 2026-05-26
chore(e2e): add minimal smoke harness

codecapitano:
## Summary

Adds a minimal vite + React harness at `e2e/smoke/` and three Cypress specs at `cypress/e2e/smoke/` that exercise the Faro Web SDK against an intercepted `/collect` endpoint:

- `init.cy.ts` — asserts the SDK populates `meta.browser` and `meta.app` on transport POSTs.
- `transport.cy.ts` — clicks the "push log" button, asserts the intercepted payload contains the expected log entry at `info`.
- `error-boundary.cy.ts` — clicks the "throw error" button inside a `FaroErrorBoundary`, asserts the intercepted payload contains the captured exception.

Wires a new `e2e-smoke` CI job that runs the harness on the existing `cypress/browsers:26.0.0` image **without** the Postgres service container the demo-driven E2E currently requires.

Sets up a follow-up PR that removes the in-repo demo app and points Faro users at `grafana/quickpizza` (already deployed at `quickpizza.grafana-dev.com`) as the canonical Frontend Observability demo.

**Purely additive in this PR**: the existing `cypress.config.ts` + demo-driven `e2e` job stay in place. Nothing is removed.

## Scope decisions

- **Conservative coverage.** Smoke covers init/log/error-boundary. The existing demo-driven E2E covers more (events, userActions, measurements, tracing, multiple error shapes — see `cypress/e2e/demo/*.cy.ts`). That breadth is genuinely useful and will be revisited as the smoke matures; for v1, the goal is foundational regression signal that runs in seconds without Postgres.
- **No `<StrictMode>`** around the harness app — React's dev double-render would double-fire the thrown error in `Thrower`, making assertions ambiguous.
- **`cy.on('uncaught:exception', () => false)`** is set in `error-boundary.cy.ts` — React in dev mode propagates boundary-caught errors to `window.onerror`, mirroring the same defensive pattern in the existing `cypress/e2e/demo/errors.cy.ts`.
- **`data-cy="btn-*"`** prefix in `main.tsx` matches the existing demo specs' convention.
- **Cypress preprocessor**: the new `cypress.smoke.config.ts` uses Cypress's default `esbuild` bundler — no dependency on `@cypress/webpack-preprocessor`. This unblocks dropping the webpack preprocessor in the follow-up PR.

## Side adjustment to `eslint.config.mjs`

Added an override for `cypress/**/*.{js,ts}` declaring Cypress globals (`cy`, `Cypress`, `context`, `it`, `expect`, etc.). Without this, the pre-commit lint-staged hook fails on any new Cypress spec — the existing demo specs predate the hook so they never tripped it. `yarn quality:lint` (root) doesn't lint `cypress/` at all (it only lints per-workspace), so this gap was invisible until now.

## Test plan

- [x] `yarn install --immutable` — lockfile accepted (peer-dep warnings are pre-existing)
- [x] `yarn build` — 7 workspaces green
- [x] `yarn start:smoke` — vite serves the harness on `:5174`
- [x] `yarn quality:e2e:smoke:ci` — 3 specs / 3 tests pass in ~3s
- [x] `yarn quality:e2e:ci` (existing demo-driven E2E) — unchanged, still works (not run locally; runs in `e2e` CI job — purely additive change)
- [x] `yarn quality:lint:fix` — no unwanted changes
- [x] `yarn quality:lint` — 7 workspaces green
- [x] `yarn quality:circular-deps` — no cycles, 449 files
- [x] `yarn quality:test` — 537 unit tests pass, 1 pre-existing skip
- [x] ESLint clean on the new specs **and** the existing `cypress/e2e/demo/*.cy.ts` files (the new override fixes both)

## Follow-up

Once this is merged + green on `main`, a separate PR will:

- Delete `demo/` and `cypress/e2e/demo/` + `cypress.config.ts`.
- Remove the `e2e` job + Postgres service from `pull-request.yml` and `release-please.yml`.
- Drop `@cypress/webpack-preprocessor`, `esbuild-loader`, `webpack` from root devDeps.
- Update README + tutorial docs to point at `grafana/quickpizza`.
- Add a `CONTRIBUTING.md` section documenting local-dev paths (knowledge-workbench harness, Grafana Cloud dev, in-memory stub).


domasx2 (dismissed):
makes sense.
I have an oranisational proposal tho: maybe move the e2e mini app to a separate package rather than making it part of root project, make it more self contained?
Would remove exra deps and surface area from rot project

codecapitano:
> makes sense. I have an oranisational proposal tho: maybe move the e2e mini app to a separate package rather than making it part of root project, make it more self contained? Would remove exra deps and surface area from rot project

Makes a lot of sense, will update

domasx2 (approved):
one more thing tho... playwright? :D

PR #2108, opened 2026-05-29
fix(core): use crypto.getRandomValues in genShortID

martinkuba:
## Summary

Replaces `Math.random()` in `genShortID` with `globalThis.crypto.getRandomValues`, falling back to `Math.random()` in environments without Web Crypto.

## Why

Static analysis tools (CodeQL, and many third-party SAST scanners customers run over the Faro bundle) flag `Math.random()` used to generate identifiers as a potential weakness. The alert is technically a false positive here — Faro session IDs aren't authentication tokens and predicting one doesn't enable an attack — but switching to the secure primitive avoids the recurring scanner noise both internally and for downstream consumers.

The replacement also removes the small non-uniformity in the generated IDs: the old `Math.floor(Math.random() * 59)` slightly favored the lower end of the alphabet. With `Uint32Array` the bias is ~1 part in 72 million.

## Test plan

- [x] `yarn quality` passes in `packages/core`
- [x] Existing session-manager tests pass (`packages/web-sdk/src/instrumentations/session` — 61 tests)
- [x] Spot-check that `genShortID()` still returns a 10-char string from the expected alphabet in a browser

### DIFF
--- a/packages/core/src/utils/shortId.ts
+++ b/packages/core/src/utils/shortId.ts
@@ -1,7 +1,20 @@
 const alphabet = 'abcdefghijkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ0123456789';
 
 export function genShortID(length = 10): string {
-  return Array.from(Array(length))
-    .map(() => alphabet[Math.floor(Math.random() * alphabet.length)]!)
-    .join('');
+  const values = new Uint32Array(length);
+  const cryptoObj = typeof globalThis !== 'undefined' ? globalThis.crypto : undefined;
+
+  if (cryptoObj?.getRandomValues) {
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/shortId.ts
```diff
@@ -1,7 +1,20 @@
 const alphabet = 'abcdefghijkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ0123456789';
 
 export function genShortID(length = 10): string {
```
domasx2, line 3: maybe add a test for it? some non trivial logic here now

domasx2 (dismissed):

eskirk (approved):

PR #2110, opened 2026-05-30
refactor(ci): split release-please workflow by privilege level

martinkuba:
## Summary

Splits the release-please workflow into three jobs separated by privilege level. The App token is no longer in scope while code from the release-please branch executes.

## Why

Today the release-please job mints an App token (broader permissions than the default `GITHUB_TOKEN`), checks out the release-please branch, and runs `yarn install --immutable` + `yarn quality:lint:fix` on the checked-out content — all in the same job, with the App token in env the whole time.

The release-please branch is writable by anyone with repo write access. An attacker-controlled commit on that branch could run a malicious `postinstall` during `yarn install`, exfiltrate the App token, and push tampered commits under the bot's identity. The same threat applies to any compromised transitive dev dep on main: its postinstall would fire next to the App token. Static analysis (CodeQL `actions/untrusted-checkout/medium`) and GitHub's own security docs flag this pattern.

This PR removes that adjacency entirely. The lockfile refresh and changelog formatting still happen automatically — just in an unprivileged context.

## Architecture

```
release-please       [PRIVILEGED — App token]
  └─ mint token, run release-please-action
  └─ resolve PR head SHA via API
  └─ STOPS — no checkout of release-please branch

refresh-release-pr   [UNPRIVILEGED — contents: read only]
  └─ checkout release-please's pinned SHA
  └─ yarn install --immutable --mode skip-build (no lifecycle scripts)
  └─ yarn quality:lint:fix
  └─ git diff → patch artifact

apply-refresh        [PRIVILEGED — App token, narrow scope]
  └─ checkout pinned SHA pristine (no patch content yet)
  └─ download patch artifact
  └─ VALIDATE patch paths against allow-list (yarn.lock + **/CHANGELOG.md)
  └─ git apply (data only) + git push
```

Key invariants:
- The App token never coexists with code execution on checked-out release-please content.
- The patch is validated against a strict path allow-list before apply. Verified against the last 5 release PRs (#2098, #2070, #2068, #2064, #2052) — the refresh has always touched only `yarn.lock` and `CHANGELOG.md`.
- `--mode skip-build` skips all install/postinstall scripts. Verified locally that `yarn quality:lint:fix` works without lifecycle scripts firing.
- Push is plain (no `--force`): if the release-please branch moved unexpectedly, the push fails non-fast-forward rather than silently overriding.

The `test` and `publish` jobs are unchanged — they run in the merge-trigger flow, not the create-PR flow, and operate on main directly.

## Test plan

Cannot fully verify until a release-please PR is created. The behavior I'd want to confirm in CI:

- [ ] On a push that should not create a release: only `release-please` job runs, all downstream jobs are skipped.
- [ ] On a push that creates a release PR: `release-please` → `refresh-release-pr` → `apply-refresh` runs. The release PR gets the refresh commit under the `app-o11y-kwl-ci[bot]` identity.
- [ ] On a push that is the merge of a release PR: `release-please` → `test` → `publish` runs (with environment approval gate).
- [ ] `apply-refresh` correctly rejects a patch touching files outside the allow-list (we can simulate this in a future test).

## Not included

- Concurrency directive (`cancel-in-progress`). Decided to leave separate so the security refactor stands on its own.
- Hardening the `publish` job's `yarn install --immutable` (which still runs lifecycle scripts with the npm OIDC context available). Out of scope here; tracked separately.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -10,6 +10,12 @@ permissions:
   id-token: write
 
 jobs:
+  # Create the release PR (or, on a merged release PR, tag the release). This
+  # job is privileged: it mints a GitHub App token via Vault OIDC and uses it
+  # to push commits/tags/PR. It MUST NOT execute untrusted code from the
+  # release-please branch — that happens in the unprivileged
+  # `refresh-release-pr` job, and `apply-refresh` only does pure-data
+  # operations (git apply + git push) against a path-allow-listed patch.
   release-please:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
@@ -23,7 +29,10 @@ jobs:
       id-token: write
     outputs:
       releases_created: ${{ steps.release.outputs.releases_created }}
+      prs_created: ${{ steps.release.outputs.prs_created }}
       tag_name: ${{ steps.release.outputs.tag_name }}
+      pr_head_sha: ${{ steps.resolve-pr.outputs.sha }}
+      pr_head_ref: ${{ steps.resolve-pr.outputs.ref }}
 
     steps:
       - name: Mint GitHub App token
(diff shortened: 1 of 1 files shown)


domasx2:
been a while in draft and largely superseded by later improvements, notably disabling postinstall scripts. Closing for now

Issue #1803, opened 2025-12-03
Support Initialization in Web Worker


zicklag:
## Description

We need to create a faro SDK instance in a web worker, where our app does some processing, but because `window` doesn't exist in the web worker scope, it fails while initializing the default meatas and checking for a `k6` parameter on the `window` object.

## Proposed solution

I think this can probably be fixed by using `globalThis` instead of `window`. `globalThis` is equivalent to `window` in a browser session, but also exists in all other JS contexts.

## Context

I tried using the config to disable exposing faro on the global object, but that doesn't fix it because that's not actually where it's trying to access the `window` object in this case.

Would you be open to a PR to fix this?

PR #2122, opened 2026-06-03
fix(core): allow BatchExecutor to run in worker scopes

bereziukadam:
## Why
`BatchExecutor`'s constructor touches `document.addEventListener` and `window.setInterval` unconditionally, so it throws `ReferenceError: window is not defined` (and `document is not defined`) when `initializeFaro` runs inside a web worker or shared worker - preventing batching from being used in any worker-side Faro integration.

## What
- Guard the `visibilitychange` listener behind `typeof document !== 'undefined'` - worker scopes have no `document` and no equivalent event.
- Replace `window.setInterval` with the bare `setInterval` reference so the timer is sourced from whichever global is available (`Window` on the main thread, `WorkerGlobalScope` in workers).
- Add a `node`-environment regression test that constructs `BatchExecutor` with neither `document` nor `window` defined and verifies the worker-safe interval flushes pending items.

Main-thread behaviour is unchanged: the document-hidden flush still fires when present, all existing tests still pass.

## Checklist
- [x] Tests added
- [ ] Changelog updated _(handled by release-please from the conventional commit title)_
- [x] Documentation updated _(no public API change)_


bereziukadam:
Related issue: https://github.com/grafana/faro-web-sdk/issues/1803

packages/core/src/transports/batchExecutor.ts
```diff
@@ -12,26 +12,27 @@ export class BatchExecutor {
...
     }
 
     // Send batched/buffered data when user navigates to new page, switches or closes the tab, minimizes or closes the browser.
     // If on mobile, it also sends data if user switches from the browser to a different app.
-    document.addEventListener('visibilitychange', () => {
-      if (document.visibilityState === 'hidden') {
-        this.flush();
-      }
-    });
+    if (typeof document !== 'undefined') {
```
bereziukadam, line 29: Guarded so the constructor doesn't throw in worker scopes - `document` is undefined there. Worker scopes have no `visibilitychange` event analog either, so we just skip this hook entirely.

packages/core/src/transports/batchExecutor.ts
```diff
@@ -49,7 +50,7 @@ export class BatchExecutor {
   start(): void {
     this.paused = false;
     if (this.sendTimeout > 0) {
-      this.flushInterval = window.setInterval(() => this.flush(), this.sendTimeout);
+      this.flushInterval = setInterval(() => this.flush(), this.sendTimeout);
```
bereziukadam, line 53: setInterval` is defined on both `Window` and `WorkerGlobalScope`. Using the bare reference instead of `window.setInterval` lets `BatchExecutor` run unchanged in dedicated and shared workers without any feature detection at the call site.

packages/core/src/transports/batchExecutor.ts
```diff
@@ -12,26 +12,27 @@ export class BatchExecutor {
   private signalBuffer: TransportItem[] = [];
   private sendFn: SendFn;
   private paused: boolean;
-  private flushInterval: number;
+  private flushInterval?: ReturnType<typeof setInterval>;
```
bereziukadam, line 15: Type is platform-agnostic: `ReturnType<typeof setInterval>` resolves to `number` under DOM-only lib and to `NodeJS.Timeout` when `@types/node` is in `tsconfig.types` (as it is here for the jest test env). Made it optional and dropped the `-1` sentinel - `clearInterval(undefined)` is a no-op in both browser and Node runtimes, so `pause()` still works correctly before `start()` has run.

codecapitano (approved):
Good change, many thanks.
Note: Faro has more uses of the window object but they are mostly related to specific instruments.

codecapitano (approved):

PR #2126, opened 2026-06-11
fix(replay): use fixed-length mask for input values

d0ugal:
## Summary

- Replace the default `maskInputFn` (which fell through to rrweb's length-preserving `'*'.repeat(text.length)`) with a constant-length mask (`'******'`) that does not leak input length
- Return empty string for empty inputs to preserve correct visual state in replays
- Guard against `maskInputFn: undefined` silently reverting to rrweb's insecure default
- User-supplied `maskInputFn` still takes precedence when provided
- Update README to document the new default

Addresses finding **F-6 (P2)** from AppSec review ASE26012.

Closes https://github.com/grafana/session-replay/issues/93

## Why a fixed-length mask also prevents keystroke-counting

rrweb's `initInputObserver` listens for `input` and `change` events on every keystroke, but it deduplicates consecutive events via `cbWithDedup` — if the new masked `text` is identical to the last recorded value, the event is dropped silently. Because every non-empty input now produces the same `'******'` string, only the first keystroke is recorded; subsequent keystrokes are deduped away. An attacker cannot count `******` events to infer input length.

Concretely:
1. User types `"a"` → `maskInputFn("a")` → `"******"` → **recorded**
2. User types `"ab"` → `maskInputFn("ab")` → `"******"` → **deduped, dropped**
3. User types `"abc"` → `maskInputFn("abc")` → `"******"` → **deduped, dropped**
4. User clears field → `maskInputFn("")` → `""` → **recorded**

The replay sees only two events: `"******"` then `""`.

## Test plan

- [x] Default mask produces `'******'` for inputs of any length (4-char and 20-char produce identical output)
- [x] Empty input returns empty string (no false `******` in empty fields)
- [x] Custom `maskInputFn` overrides the default when supplied
- [x] Explicit `maskInputFn: undefined` falls back to the secure default (regression test)
- [x] Default `maskInputFn` is passed through to rrweb's `record()` options
- [x] All existing tests continue to pass (41 total)


domasx2 (approved):
makes sense 👍

kpelelis:
Pinging @d0ugal if you still want to get this merged

PR #2141, opened 2026-06-23
feat(hackathon): remote sampling config

eskirk:
## Why

Operators can't change an instrumented app's session sampling rate without the customer redeploying their own app. This adds opt-in **remote configuration** so Faro fetches the session sampling rate from the collector's `GET /config/{key}` on init and applies it — without blocking page load.

## What

- Opt-in `remoteConfig?: { url?, timeoutMs?, maxBufferBytes? }`; **`initializeFaro` stays synchronous** and the default (no `remoteConfig`) path is unchanged.
- Core transports gain a bounded `hold`/`flushHeld`/`dropHeld` pre-decision buffer primitive (default 64KB).
- New `web-sdk/remoteConfig` module: schema-versioned `localStorage` cache, `ETag` conditional fetch with a ~1.5s timeout, **warm-cache** fast path and **cold-cache defer-and-buffer**, finalize-once flush-or-drop, never throws (always falls back to bundled/cached default).
- Cold path forces keep-all before init and reconciles the session via `markSessionNotSampled` at finalize, so the **remote rate is the single source of truth** for the current session; a locally-defined custom `sampler` still wins.

> Hackathon limitation (documented, non-blocking): a cold + not-sampled session can leak its `session_start` event pre-hold; production follow-up is to engage the hold before instrumentation init.

## Links

- Endpoint slice: grafana/app-o11y-kwl-endpoint#1612
- Part of the cross-repo "remote sampling config" hackathon (endpoint + this SDK + plugin + terraform-provider).

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated (CHANGELOG only for the hackathon)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+### Features
+
+- **web-sdk:** add opt-in `remoteConfig` option for a remotely-configured session sampling rate. When provided, the SDK initializes instrumentation immediately, buffers outgoing telemetry in a bounded in-memory queue, resolves the per-app sampling rate (cache-first, with a hard timeout and ETag-based background revalidation), finalizes the session sampling decision exactly once, then flushes-or-drops the buffer. `initializeFaro()` remains synchronous; when `remoteConfig` is absent, behavior is unchanged.
+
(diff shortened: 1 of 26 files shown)


codecapitano:
@eskirk do you want to keep that or can we we close the PR?

eskirk:
@codecapitano nah it was a good experiment, but it will need more discussions

PR #2143, opened 2026-06-23
chore(renovate): pin react-router to v7

domasx2:
Constrains `react-router`/`react-router-dom` to `<8.0.0` so Renovate stops proposing the v8 major - Grafana does not support that yet afaik.


codecapitano:
@domasx2 what are the reasons to pin < v8 versions?

domasx2:
oops, i may have fired this off against the wrong repo :fearful:

PR #2162, opened 2026-06-27
fix: emit navigation changes after commit

### LINKED ISSUE #1794: Navigation instrumentation should use the currententrychange event
## Description

The navigation instrumentation is listening to the `navigate` event to detect navigations. This event is fired when a navigation is initialized, and it allows for the navigation to be cancelled. This could lead to reporting false positives. We should use the [currententrychange](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/currententrychange_event) event instead, which signifies a committed/successful navigation.

Saithej2k:
Summary

- Listen for Navigation API currententrychange events instead of navigate events.
- Emit URL changes from navigation.currentEntry.url after navigation commits.
- Remove the navigate intercept wrapper and update URL monitor tests.

Testing

- git diff --check
- yarn eslint packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.test.ts
- yarn prettier --check packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.test.ts
- Focused Jest run for packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.test.ts

Closes #1794

### DIFF
--- a/packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.test.ts
+++ b/packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.test.ts
@@ -70,14 +70,13 @@ describe('monitorUrlChanges', () => {
 
   describe('with Navigation API', () => {
     let originalNavigation: any;
-    let originalNavigateEvent: any;
 
     beforeEach(() => {
       originalNavigation = (window as any).navigation;
-      originalNavigateEvent = (window as any).NavigateEvent;
 
-      const listeners: Record<string, Function[]> = { navigate: [] };
+      const listeners: Record<string, Function[]> = { currententrychange: [], navigate: [] };
       (window as any).navigation = {
+        currentEntry: { url: window.location.href },
         addEventListener: (type: string, cb: Function) => listeners[type]?.push(cb),
         removeEventListener: (type: string, cb: Function) => {
           const arr = listeners[type];
@@ -91,41 +90,27 @@ describe('monitorUrlChanges', () => {
         },
         _dispatch: (type: string, ev: any) => listeners[type]?.forEach((cb) => cb(ev)),
       };
-
-      function FakeNavigateEvent(this: any, _type: string, init: any) {
-        this.destination = init?.destination;
-      }
-      (FakeNavigateEvent as any).prototype = {
-        intercept: jest.fn(),
-      };
-
-      (window as any).NavigateEvent = FakeNavigateEvent as any;
     });
 
     afterEach(() => {
       (window as any).navigation = originalNavigation;
-      (window as any).NavigateEvent = originalNavigateEvent;
     });
 
-    it('emits on same-document navigate events and does not patch history', () => {
+    it('emits on currententrychange events and does not patch history', () => {
       const initialHref = window.location.href;
       const observable = monitorUrlChanges();
       const subscriber = jest.fn();
       observable.subscribe(subscriber);
 
-      (window as any).navigation._dispatch(
-        'navigate',
-        new (window as any).NavigateEvent('navigate', {
-          destination: { url: initialHref + '#nav', sameDocument: true },
-        })
-      );
+      (window as any).navigation.currentEntry = { url: initialHref + '#nav' };
+      (window as any).navigation._dispatch('currententrychange', {});
 
       expect(subscriber).toHaveBeenCalledTimes(1);
       expect(subscriber).toHaveBeenCalledWith({
         type: MESSAGE_TYPE_URL_CHANGE,
         from: initialHref,
         to: initialHref + '#nav',
-        trigger: 'navigate',
+        trigger: 'currententrychange',
       });
 
       // ensure history methods were not wrapped in this mode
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts
```diff
@@ -34,51 +33,25 @@ export function monitorUrlChanges(): Observable<UrlChangeMessage> {
...
+    const navigation = (window as any).navigation;
+    const hasNavigation =
+      typeof navigation?.addEventListener === 'function' &&
+      typeof navigation?.removeEventListener === 'function' &&
+      'currentEntry' in navigation;
 
     if (hasNavigation) {
       // Prefer Navigation API when supported: do not patch history or add popstate/hashchange listeners
-      onNavigateHandler = (e: any) => {
+      onCurrentEntryChangeHandler = () => {
```
kpelelis: The handler discards its event argument. I suggest using `event.from.url` for from instead of the global lastHref. That global is exactly the state the old bug corrupted.

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts
```diff
@@ -6,7 +6,7 @@ export type UrlChangeMessage = {
   type: typeof MESSAGE_TYPE_URL_CHANGE;
   from: string;
   to: string;
-  trigger: 'pushState' | 'replaceState' | 'popstate' | 'hashchange' | 'navigate' | 'navigate-intercept';
+  trigger: 'pushState' | 'replaceState' | 'popstate' | 'hashchange' | 'currententrychange';
```
kpelelis: The Navigation branch emits one trigger where the fallback emits four, so a push is indistinguishable from a back button traversal. Map event.navigationType, or drop trigger entirely since nothing consumes it.

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts
```diff
@@ -34,51 +33,25 @@ export function monitorUrlChanges(): Observable<UrlChangeMessage> {
...
       // Prefer Navigation API when supported: do not patch history or add popstate/hashchange listeners
-      onNavigateHandler = (e: any) => {
+      onCurrentEntryChangeHandler = () => {
         try {
-          const destination = e?.destination as { url?: string; sameDocument?: boolean } | undefined;
-          if (destination?.sameDocument && typeof destination.url === 'string') {
-            emit('navigate', destination.url);
+          const currentUrl = navigation.currentEntry?.url;
+          if (typeof currentUrl === 'string') {
+            emit('currententrychange', currentUrl);
```
kpelelis: Emitting at commit rather than at `intercept()` means `intercept({ commit: 'after-transition' })` can land outside ActivityWindowTracker's 100 ms window and drop the event. Is this what you intended for or should it be handled differently?

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts
```diff
@@ -34,51 +33,25 @@ export function monitorUrlChanges(): Observable<UrlChangeMessage> {
   }
 
   if (!isInstrumented) {
-    const hasNavigation = 'navigation' in window && 'NavigateEvent' in (window as any);
+    const navigation = (window as any).navigation;
```
kpelelis: This shadows the global navigation object. Rename to navigationApi.

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.test.ts
```diff
@@ -135,26 +120,34 @@ describe('monitorUrlChanges', () => {
       expect(replaceDesc?.value).toBeDefined();
     });
 
-    it('emits on intercept for cross-document navigations (soft navigation)', () => {
+    it('does not emit on navigate events before navigation commits', () => {
```
kpelelis: This doesn't reproduce the underlying issue. No navigate listener exists to fire, so it's near tautological. Add the case with real impact: dispatch navigate, never commit, then assert the next real `currententrychange` still reports the correct from.

kpelelis (changes requested):
Looks good. Take a look at inline comments

domasx2:
Hi @Saithej2k can you take a look at the comments?

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts
```diff
@@ -16,69 +16,59 @@ let originalPushState: typeof window.history.pushState | undefined;
...
     lastHref = location.href;
   }
 
-  function emit(trigger: UrlChangeMessage['trigger'], toOverride?: string) {
+  function emit(trigger: UrlChangeMessage['trigger'], toOverride?: string, fromOverride?: string) {
     const next = toOverride ?? location.href;
-    if (next !== lastHref) {
-      urlChangeObservable!.notify({ type: MESSAGE_TYPE_URL_CHANGE, from: lastHref!, to: next, trigger });
+    const previous = fromOverride ?? lastHref;
+    if (previous && next !== previous) {
```
martinkuba, line 44: Because of this check, reloads will never be captured. I would actually not include reload rather than make it work.

Reloads are worth capturing, but they belong on a cross-document path this monitor doesn't currently handle at all (and is out of scope for this change).

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.test.ts
```diff
@@ -135,26 +123,49 @@ describe('monitorUrlChanges', () => {
       expect(replaceDesc?.value).toBeDefined();
     });
 
-    it('emits on intercept for cross-document navigations (soft navigation)', () => {
+    it('does not let uncommitted navigate events corrupt the next committed from URL', () => {
```
martinkuba, line 126: This test is incomplete because the from URL is taken from the from parameter of the currententrychange event argument, so even if the previous navigate event polluted the lastHref variable, this test would not catch it. However, this scenario is not realistic anyway because the currententrychange event will always have from populated.

I think the better test would be simply - emit navigate event, assert that the monitor did not emit, then emit currententrychange and check it did emit. The missing piece is the assertion in the middle.

martinkuba (commented):
@Saithej2k Thanks for the last round of changes. I left comment on a couple of things that I think should still be addressed.

packages/web-sdk/src/instrumentations/_internal/monitors/urlChangeMonitor.ts
martinkuba: I think the option in question is `intercept({ precommitHandler })`. `after-transition` is a value for `focusReset` and `scroll` ([MDN link](https://developer.mozilla.org/en-US/docs/Web/API/NavigateEvent/intercept#parameters)).

A precommit handler that makes a fetch call will keep the tracker active. It would break when the precommit work isn't an observed HTTP request (like cached promise, a timer). I think that's narrow and the remedy is in the tracker rather than here. I don't think this should block this PR.

PR #2167, opened 2026-06-29
fix(session): stop a background tab from emitting an expired session id

domasx2:
## Problem

With persistent sessions (shared across tabs via `localStorage`), when one tab
rotates its session — e.g. when the lifetime cap is reached — other tabs that
still hold the previous session id in memory never pick up the change. They keep
emitting the **expired** id on every signal, so a background tab can stay pinned
to a dead session indefinitely.

## Fix

When a tab finds a newer valid session in shared storage, it now adopts that
session into its in-memory metas instead of only refreshing `lastActivity`, so it
stops emitting the stale id. Adoption is silent — no duplicate
`session_start`/`session_extend`, since the rotating tab already emitted it — and
is gated by an instance-scoped flag so concurrent faro instances don't share it.

## Test

Adds a Playwright smoke spec covering the cross-tab case: after one tab rotates a
shared session, a background tab converges to the rotated id rather than
continuing to emit the expired one.

_PR description authored by Claude on Domas's behalf._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/web-sdk/src/instrumentations/session/sessionManager/VolatileSessionManager.ts
+++ b/packages/web-sdk/src/instrumentations/session/sessionManager/VolatileSessionManager.ts
@@ -1,4 +1,5 @@
 import { faro, stringifyExternalJson } from '@grafana/faro-core';
+import type { MetaSession } from '@grafana/faro-core';
 
 import { throttle } from '../../../utils';
 import { getItem, removeItem, setItem, webStorageType } from '../../../utils/webStorage';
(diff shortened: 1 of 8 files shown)


packages/web-sdk/src/instrumentations/session/sessionManager/VolatileSessionManager.ts
```diff
@@ -11,10 +12,26 @@ export class VolatileSessionsManager {
...
+  isAdopting = (): boolean => this.adopting;
+
+  private adoptSession = (sessionMeta: MetaSession): void => {
+    this.adopting = true;
+    try {
+      faro.api?.setSession(sessionMeta);
+    } finally {
+      this.adopting = false;
+    }
+  };
```
codecapitano: sessionStorage is tab-local, so this adoption block can never fire — `sessionFromStorage.sessionId` will always match the in-memory id. Consider removing it (and not passing `adoptSession` here) rather than keeping it for parity so we don't need to maintain the dead code.

domasx2: good point, updated to stub `isAdopting` for volatile session manager

codecapitano (approved):

PR #2173, opened 2026-07-07
feat(instrumentation-replay): add ssn / creditCard / usAddress keys to maskInputOptions

justin-thurston:
## Summary

Adds three value-pattern matchers to `maskInputOptions` in `@grafana/faro-instrumentation-replay`, addressing the SSN / credit-card / US-address portion of #2169:

- `ssn` — masks any input value matching `XXX-XX-XXXX` or 9 contiguous digits (with SSA-disallowed ranges rejected: area `000` / `666` / `9XX`, group `00`, serial `0000`).
- `creditCard` — masks 13–19 digit PANs with optional dash/space separators that pass a Luhn checksum, so 16-digit order IDs are not swept up.
- `usAddress` — masks values matching `Number Street, City, ST, ZIP` (with optional `+4` ZIP extension) on a single line.

## Why this needs a wrapper rather than a straight passthrough

rrweb's `maskInputFn` only fires for inputs it has already decided to mask (via `maskAllInputs` or `maskInputOptions`). Value-pattern matching only helps if the wrapper sees *every* input.

When any of the three new keys is enabled the instrumentation:

1. Forces `maskAllInputs: true` to rrweb so the wrapper is consulted for every input.
2. Wraps `maskInputFn` to:
   - Mask the value if it matches any enabled pattern.
   - Otherwise, re-implement rrweb's original mask decision using the user's `maskAllInputs` + `maskInputOptions` + element type, so inputs that would not have been masked before are still passed through unchanged.
3. Strips `ssn` / `creditCard` / `usAddress` from `maskInputOptions` before forwarding to rrweb, which only recognizes HTML input-type keys.

When none of the three keys is enabled the behavior is unchanged.

## Files changed

- `experimental/instrumentation-replay/src/types.ts` — add `ssn` / `creditCard` / `usAddress` to `MaskInputOptions`; export `PATTERN_MASK_KEYS` / `PatternMaskKey`.
- `experimental/instrumentation-replay/src/patterns.ts` (new) — `isUsSsn` / `isCreditCard` / `isUsAddress` detectors, Luhn checksum, `valueMatchesEnabledPattern`, `hasAnyPatternEnabled`.
- `experimental/instrumentation-replay/src/instrumentation.ts` — build a pattern-aware `maskInputFn` wrapper, force `maskAllInputs` when patterns are enabled, sanitize the options object before handing it to rrweb.
- `experimental/instrumentation-replay/src/patterns.test.ts` (new) — 42 tests covering formatted/unformatted SSN, Luhn-valid PANs across issuer lengths (13/14/15/16/19), address shapes, and negative cases (SSA-disallowed SSN ranges, 16-digit non-Luhn order IDs, malformed addresses).
- `experimental/instrumentation-replay/src/instrumentation.test.ts` — 7 new tests exercising the wrapper: key sanitization, forced `maskAllInputs`, no wrapper when patterns are disabled, positive/negative masking, respect for existing input-type opt-ins, delegation to a user-supplied `maskInputFn`.
- `experimental/instrumentation-replay/README.md` — document the three new keys with a note that they are value-pattern matchers rather than input-type keys.

## Test plan

- [x] `patterns.test.ts`: 42/42 passing locally.
- [ ] `instrumentation.test.ts`: existing suite fails to run on a fresh clone locally due to a pre-existing `@grafana/faro-core` module-resolution issue in ts-jest that also affects `main` — CI's workspace-linker setup should pick up the new tests.
- [x] `eslint`, `prettier -c`, and `markdownlint` are all clean on changed files.

## Refs

- Issue: #2169
- Scope: this PR implements the `maskInputOptions`-keyed subset of that issue. A broader `maskTextPatterns` surface for masking pattern matches in rendered text content (outside inputs) is left as future work.


experimental/instrumentation-replay/src/types.ts
```diff
@@ -132,6 +132,30 @@ export type MaskInputOptions = Partial<{
...
+   */
+  usAddress: boolean;
 }>;
 
+/**
+ * Subset of `MaskInputOptions` whose keys are value-pattern matchers
+ * rather than HTML input types. These keys are consumed inside the
+ * instrumentation and stripped before forwarding to rrweb.
+ */
+export const PATTERN_MASK_KEYS = ['ssn', 'creditCard', 'usAddress'] as const;
```
domasx2: nitpick: not strictly a "type"

experimental/instrumentation-replay/src/patterns.ts
```diff
@@ -0,0 +1,91 @@
...
+// Conservative US address shape on a single line:
+//   "<number> <street...>, <city>, <ST>, <ZIP>"
+// ZIP may be 5 or 9 digits; state is two letters.
+const US_ADDRESS_REGEX = /^\s*\d+\s+[A-Za-z0-9.'\- ]+,\s*[A-Za-z.'\- ]+,\s*[A-Za-z]{2},?\s*\d{5}(?:-\d{4})?\s*$/;
+
+export function isUsSsn(value: string): boolean {
+  return SSN_REGEX.test(value);
+}
+
+export function isCreditCard(value: string): boolean {
```
domasx2, line 36: AI review flagged that it will mask the the full number, but won't mask partial number while user is typing 🤔 I haven't tested and not sure how much of a concern it is though

domasx2 (dismissed):
makes sense, few nits

kpelelis:
Hey @justin-thurston, I am doing an open pull request grooming. Is this something that you are still looking into getting merged?

experimental/instrumentation-replay/src/types.ts
justin-thurston: Good catch — moved `PATTERN_MASK_KEYS` and `PatternMaskKey` into `patterns.ts` (where they're actually used) so `types.ts` stays types-only. See a957a98b.

experimental/instrumentation-replay/src/patterns.ts
justin-thurston: Good flag. Addressed in a957a98b by adding `elementIndicatesSensitiveField()` — the wrapper now inspects the input's [`autocomplete`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill) attribute in addition to the value shape, so partial values mask from the first keystroke.

Coverage:
- `creditCard: true` → mask any input with a `cc-*` autocomplete token (`cc-number`, `cc-csc`, `cc-name`, `cc-exp*`, `cc-type`). Handles space-separated multi-token values like `"section-billing cc-number"`.
- `usAddress: true` → mask any input with `autocomplete=street-address` / `address-line1..3`. Intentionally excludes `address-level*` (city/state) and `postal-code` — those are less individually sensitive and out of scope for the single-line-address matcher.
- `ssn` — no standard `autocomplete` value exists in the WHATWG spec, so it stays value-only. Non-autocompleted PAN/address inputs still fall through to the completed-value Luhn / single-line-address detectors.

This is the same approach Sentry Replay and Datadog RUM take for CC field detection. 16 new tests cover the detector directly (`patterns.test.ts`) plus mid-typing masking through the wrapper (`instrumentation.test.ts`).

justin-thurston:
@domasx2 — updated per your review:

- **Nit #1** (`types.ts:158`) resolved by moving `PATTERN_MASK_KEYS` / `PatternMaskKey` into `patterns.ts` so `types.ts` stays types-only.
- **Nit #2** (`patterns.ts` partial PAN mid-typing) resolved by adding `elementIndicatesSensitiveField()` — the wrapper now mask-triggers on the input's [`autocomplete`](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill) attribute (`cc-*`, `street-address`, `address-line1..3`), so partial values mask from the first keystroke. Same approach Sentry Replay / Datadog RUM take. SSN has no standard `autocomplete` value so it stays value-only.

Also rebased onto `main` — #2126's fixed-length `defaultMaskInputFn` is now honored end-to-end, and the TS4111 CI failures should clear (main's new `tsconfig.spec.json` disables that check for the spec project).

Local verification: **114/114** jest tests passing, `eslint` + `prettier` + `markdownlint` + circular-deps all clean.

Detailed replies on both inline threads. Ready for another pass when you have a minute — thanks!

experimental/instrumentation-replay/src/patterns.ts
domasx2: It seems to only work on credit cards, not ssn, and only with autcomplete enabled on the site. Is that inentional? Site ommiting autocomplete won't be masked afaik :thinking:

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -131,15 +137,25 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
 
   private buildRecordOptions(): recordOptions<eventWithTime> {
+    const patternsEnabled = hasAnyPatternEnabled(this.options.maskInputOptions);
+    const sanitizedMaskInputOptions = sanitizeMaskInputOptions(this.options.maskInputOptions);
+
+    // rrweb only invokes `maskInputFn` for inputs it has already decided to
+    // mask. When any value-pattern key is enabled we need our wrapper to see
+    // every input, so we force `maskAllInputs: true` to rrweb and re-implement
+    // the user's mask decision inside the wrapper (see #2169).
+    const effectiveMaskAllInputs = patternsEnabled ? true : this.options.maskAllInputs;
```
domasx2, line 147: this seems to have a side effect - if true, rrweb will also omit which selection option was enabled. I wonder if there is a better way to implement this

domasx2 (commented):
few more nits. Also asking @d0ugal and @Blinkuu for reviews

Issue #1818, opened 2025-12-10
Instrumenting with TracingInstrumentation causing noisy neighbour failure


Jayclifford345:
## Description

We are currently trying to instrument our application with `new TracingInstrumentation()` it appears that when we do this, it causes other plugins that are also instrumented with traces to error out due to the following error

<img width="1512" height="866" alt="Image" src="https://github.com/user-attachments/assets/e4f2bd83-529c-4ea1-90b7-c2b9acaf509d" />

Looking at the console errors in the screenshot, the key issues are:

**OpenTelemetry duplicate registration errors:**

```
@opentelemetry/api: Attempted duplicate registration of API: trace
  at t.setGlobalTracerProvider (module.js:4:148379)
  at si.register (module.js:4:101737)
  at Js.initialize (module.js:4:136125)
```

```
@opentelemetry/api: Attempted duplicate registration of API: propagation
  at t.setGlobalPropagator (module.js:4:386974)
  at si.register (module.js:4:101809)
  at Js.initialize (module.js:4:136125)
```

**Root cause:** Grafana's core already has its own Faro instance with `TracingInstrumentation` that registers the global OpenTelemetry tracer and propagator. When your plugin's `TracingInstrumentation` tries to register them again, OTEL rejects it as a duplicate.

The `isolate: true` option isolates Faro's internal state, but **it does NOT prevent the OTEL global API conflict** - there's only one global `window` and OTEL's tracer/propagator can only be registered once.

## Steps to reproduce

https://github.com/user-attachments/assets/3310d8da-4a19-4117-842b-4f355529beda

## Environment

```
    "@grafana/faro-react": "^2.0.2",
    "@grafana/faro-web-tracing": "^2.0.2",
```

Issue #568, opened 2024-04-25
TracingInstrumentation for micro frontends


pietrarenner:
I have a micro frontend application that I have been monitoring using Faro. I have initialized one Faro instance per micro frontend and set the isolate variable to true in all of them, except for the one that has my global scope. I have also initialized the TracingInstrumentation in all of them. The problem is that traces that come from different micro frontends are considered to come from the same mfe, at least that is what is show to me in Grafana Cloud. For example, I have an mfe called A that sends a request to http://localhost:8000/ and a micro frontend called B that sends a request to http://localhost:8000/b. A never sends a request to http://localhost:8000/b, but, when the request is sent to that url, what appears in Grafana Cloud is as if the request had been sent by A. 

Is there a way for me to solve this? I am not quite sure what I have done wrong.

codecapitano:
Hi @pietrarenner 
I don't have a quick fix yet but I think I know what we need to change in the web-tracing package to better support microfrontends.

pietrarenner:
Thank you for the response, @codecapitano! I will be waiting for those changes!

ValeraKorovelkov:
Hi @pietrarenner. Did manual instrumentation for each MFE resolve your issue?

ValeraKorovelkov:
@codecapitano Hey! I have the same issue and tried to instrument my MFEs manually using the guide you provided, but I'm encountering a few errors
```
@opentelemetry/api: Attempted duplicate registration of API: trace
@opentelemetry/api: Attempted duplicate registration of API: context
@opentelemetry/api: Attempted duplicate registration of API: propagation
```

himanshuReact:
I have something silmilar issue where we have multiple microfrontend  in react(A,B,C) and they are hosted in the root microfrontend app Root-Shell.  We want to automtaically tag the api without any manual code so that call requests and error coming from microfrontend A should get logged with Microfrontend A only and not with the Root-Shell app. I am getting the issue that  microfrontend A api is getting tagged with A and Root Shell both in grafana dashboard Explore. I have used multiple initialisation with multiple Faro collector for this. Before I have tried with single initialisation with single collector with window object approach and multiple initialisation with single collector and isolate but it does not give the expected result. Please suggest the approach and any fix or alternative. Thanks in advance

himanshuReact:
Any luck regarding the solution @pietrarenner

codecapitano:
Hi all,

there's new option option in Faro to monitor MFEs.
[...]

PR #2178, opened 2026-07-19
fix(web-tracing): enforce a single tracing owner across Faro instances

### LINKED ISSUE #1845: TracingInstrumentation incompatible with multiple SDK instances (micro-frontends)

martinkuba:
## Why

When multiple Faro instances run on the same page (micro-frontends), `TracingInstrumentation` behaves incorrectly for every app on the page: each request produces duplicate spans in the trace, and the resulting spans carry a mix of every instance's configuration.

OpenTelemetry allows only one global tracer provider per browsing context, and the fetch/XHR instrumentations patch the global `fetch`/`XMLHttpRequest`. Today every Faro instance runs both `provider.register()` and `registerInstrumentations()` on init, but the two have opposite outcomes:

- `provider.register()` is deduplicated by OTel — the **first** instance wins, so all spans are exported through the first app (and secondary apps never receive their own spans).
- `registerInstrumentations()` is not deduplicated — the re-patch guard (`_isFetchPatched`) is per-instance, so each instance stacks another wrapper on the global `fetch`/`XHR`. Every wrapper executes, so each request produces one CLIENT span per instance (nested), each shaped by that instance's config. The second `provider.register()` also logs OTel "duplicate registration" errors.

The result is duplicate spans in the trace, config from every instance mixed together, and secondary apps blind to their own network calls. See #1845 (and the duplicate-registration crash in #1818).

## What

Make tracing single-owner: the first Faro instance to initialize a `TracingInstrumentation` claims ownership and registers the provider + instruments fetch/XHR as before. Any later instance detects the existing owner, logs a warning, and skips registration — so there is exactly one provider, one wrapper, and one span per request, with consistent config.

- New `tracingOwner.ts` — first-to-claim owner slot stored on the global object (shared across independently bundled micro-frontends; a module-level flag would not be). Independent of the `isolate` option, since provider registration and fetch/XHR patching are inherently global.
- `instrumentation.ts` — guard at the top of `initialize()`. Non-owners still call `api.initOTEL(trace, context)` so manual spans keep working against the shared provider, then return.
- Tests for the owner helper and the single-owner behavior.
- Docs: Tracing section in the architecture doc + an MFE note in the web-tracing README.

This is the first step (single tracing owner) toward the broader cross-instance MFE tracing work; per-instance HTTP event visibility is a follow-up.

## Links

Closes #1845
Related: #1818, #568

## Checklist

- [x] Tests added
- [x] Changelog updated 
- [x] Documentation updated

### DIFF
--- a/packages/web-tracing/README.md
+++ b/packages/web-tracing/README.md
@@ -23,6 +23,11 @@ export spans directly; they can still create spans against the shared provider v
 and attributed to it. This applies regardless of the `isolate` config option, since tracer-provider
 registration and fetch/XHR patching are inherently global.
 
+Cross-origin trace-header propagation is preserved across instances: every instance's
+`propagateTraceHeaderCorsUrls` is merged into the owner, so the owning instance still injects W3C
+trace headers for URLs configured by any instance. Per-instance custom span attributes
+(`applyCustomAttributesOnSpan`) are not merged — only the owner's are applied.
+
 [faro-web-sdk-package]: https://github.com/grafana/faro-web-sdk/tree/main/packages/web-sdk
 [opentelemetry-js]: https://opentelemetry.io/docs/instrumentation/js/
 [quick-start]: https://github.com/grafana/faro-web-sdk/blob/main/docs/sources/tutorials/quick-start-browser.md
--- a/packages/web-tracing/README.md
+++ b/packages/web-tracing/README.md
@@ -5,6 +5,21 @@ This package provides tools for integrating [OpenTelemetry][opentelemetry-js] ba
 
 See [quick start document][quick-start] for instructions how to set up and use.
 
+## Multiple Faro instances on one page (micro-frontends)
+
+OpenTelemetry allows only a single global tracer provider, propagator and context manager per
+browsing context, and the fetch/XHR instrumentations patch the global `fetch`/`XMLHttpRequest` only
+once. Because of this, when several Faro instances run on the same page (for example in a
+micro-frontend architecture), **only the first instance to initialize tracing owns it**: it registers
+the tracer provider and instruments fetch/XHR. Any Faro instance that adds a `TracingInstrumentation`
+afterwards detects the existing owner, logs a warning, and does not register a second provider or
+re-patch fetch/XHR (this also avoids OpenTelemetry "duplicate registration" errors).
+
+As a result, all spans are emitted by, and shaped by the configuration of, the owning instance. Later
+instances can still create manual spans against the shared provider via `faro.api.getOTEL()`, but do
+not emit their own automatic HTTP spans. This applies regardless of the `isolate` config option, since
+tracer-provider registration and fetch/XHR patching are inherently global.
+
 [faro-web-sdk-package]: https://github.com/grafana/faro-web-sdk/tree/main/packages/web-sdk
 [opentelemetry-js]: https://opentelemetry.io/docs/instrumentation/js/
 [quick-start]: https://github.com/grafana/faro-web-sdk/blob/main/docs/sources/tutorials/quick-start-browser.md
--- a/packages/web-tracing/README.md
+++ b/packages/web-tracing/README.md
@@ -15,10 +15,13 @@ the tracer provider and instruments fetch/XHR. Any Faro instance that adds a `Tr
 afterwards detects the existing owner, logs a warning, and does not register a second provider or
 re-patch fetch/XHR (this also avoids OpenTelemetry "duplicate registration" errors).
 
-As a result, all spans are emitted by, and shaped by the configuration of, the owning instance. Later
-instances can still create manual spans against the shared provider via `faro.api.getOTEL()`, but do
-not emit their own automatic HTTP spans. This applies regardless of the `isolate` config option, since
-tracer-provider registration and fetch/XHR patching are inherently global.
+As a result, **all spans are exported by the owning instance** — using its exporter, its resource
+(`service.name`), and its collector — no matter which instance created them, and only the owner
+produces automatic fetch/XHR spans. Later instances register no exporter of their own, so they never
+export spans directly; they can still create spans against the shared provider via
+`faro.api.getOTEL()` (for example, manual spans), but those spans are exported by the owning instance
+and attributed to it. This applies regardless of the `isolate` config option, since tracer-provider
+registration and fetch/XHR patching are inherently global.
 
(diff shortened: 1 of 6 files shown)


packages/web-tracing/README.md
```diff
@@ -5,6 +5,21 @@ This package provides tools for integrating [OpenTelemetry][opentelemetry-js] ba
...
+browsing context, and the fetch/XHR instrumentations patch the global `fetch`/`XMLHttpRequest` only
+once. Because of this, when several Faro instances run on the same page (for example in a
+micro-frontend architecture), **only the first instance to initialize tracing owns it**: it registers
+the tracer provider and instruments fetch/XHR. Any Faro instance that adds a `TracingInstrumentation`
+afterwards detects the existing owner, logs a warning, and does not register a second provider or
+re-patch fetch/XHR (this also avoids OpenTelemetry "duplicate registration" errors).
+
+As a result, all spans are emitted by, and shaped by the configuration of, the owning instance. Later
+instances can still create manual spans against the shared provider via `faro.api.getOTEL()`, but do
+not emit their own automatic HTTP spans. This applies regardless of the `isolate` config option, since
```
domasx2: later instances also do not exprot spans, is that correct?

domasx2 (dismissed):

packages/web-tracing/README.md
martinkuba: They do not, but they can add custom spans for the owning instance to export. I have clarified this in the doc.

packages/web-tracing/src/tracingOwner.ts
```diff
@@ -0,0 +1,59 @@
...
+    return false;
+  }
+
+  Object.defineProperty(getGlobal(), tracingOwnerKey, {
+    // configurable so tests can reset the claim; writable:false still prevents accidental reassignment
+    configurable: true,
+    enumerable: false,
+    writable: false,
+    value: { appName },
+  });
```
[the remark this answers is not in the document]
martinkuba: This is an internal object and unlikely to be tampered with.

packages/web-tracing/src/instrumentation.ts
eskirk, line 81: my only concern is that this change might unexpectedly break existing "working" implementations. 

say a child app implements a `propagateTraceHeaderCorsUrls` that is needed by the parent. even though the behavior was technically broken before, now the same traces may see CORS issues because the child config is no longer getting applied. I dont think this should prevent this from moving forward, but at least wanted to share that.

eskirk (dismissed):

packages/web-tracing/src/instrumentation.ts
martinkuba: Good catch, I have updated this to combine the `propagateTraceHeaderCorsUrls` config if the TracingInstrumentation is initialized multiple times. This will preserve the existing behavior.

domasx2 (approved):

martinkuba:
After some more discussion with @eskirk, I'm temporarily moving this to draft to agree on how we treat the behavior changes for secondary MFE apps before merging.

I think we should ship this as a fix (not a major bump), and document the multi-instance behavior changes clearly. Single-instance apps are completely unaffected, and multi-instance tracing was already broken (duplicate, misattributed spans), so this is fundamentally fixing an undocumented defect. But I want to be clear that it does change behavior for MFE setups in some edge cases.

The behavior changes for secondary apps:

- propagateTraceHeaderCorsUrls
  - already handled - it's unioned across all instances, so cross-origin propagation still works for URLs any app configures. this is non-breaking.
  - the union is clean (no collisions), and it's arguably the correct long-term behavior anyway.

- custom instrumentations
  - previously secondary apps could register custom instrumentations that produced spans; now they're skipped entirely (not just fetch/XHR)
  - this is breaking but a narrow edge case (probably non-existent)

- custom attributes
  - secondary apps' (duplicate) spans are no longer created, so no span carries these attributes - the trace is correct, but querying for those attributes directly won't match anything
  - we could merge all apps' attributes onto the single span, but that isn't fully non-breaking either (same key with different values loses one), and it produces a semantically muddy span. And if we accept owner-only for ignoreUrls below, the merge isn't worth the complexity.

- ignoreUrls
  - also not just noise removal: if the primary ignores a URL and a secondary does not, today the secondary still produces a span; now the primary's ignore wins and there's no span at all
  - there's no clean alternative - union suppresses even more, and intersection brings back collector-endpoint noise (each app only ignores its own collector)

So my proposal is to treat this as a bug fix that changes MFE behavior in these edge cases, document it in the changelog and docs, and restore per-app visibility as events in a follow-up. If we'd rather be strict and call this breaking, the alternative is a major version bump.

Issue #2105, opened 2026-05-29
OtlpHttpTransport drops all but first value from measurement payload.values (web vitals attribution data lost)


Ankit3794:
## Description

Since PR #595 (web vitals with attribution), the webVitalsWithAttribution instrumentation pushes multiple values per measurement (e.g., INP sends inp, delta, input_delay, processing_duration, presentation_delay, interaction_time, next_paint_time).

However, toMeasurementLogRecord in the OTLP transport only extracts the first key-value pair from payload.values, silently discarding all attribution sub-values.

## Steps to reproduce

- Initialize Faro with both FetchTransport and OtlpHttpTransport
- Enable webVitalsWithAttribution instrumentation
- Trigger an INP event (click any element)
- Compare what arrives at each backend:

FetchTransport → Alloy → Loki (correct — all values present):
```
inp=24 delta=24 input_delay=5.300000 processing_duration=0.300000 presentation_delay=18.400000
```

OtlpHttpTransport → OTel Collector → ClickHouse (broken — only first value):

```
measurement.name = "inp"
measurement.value = "24"
// input_delay, processing_duration, presentation_delay → MISSING
```

## Expected behavior

When webVitalsWithAttribution instrumentation pushes a measurement with multiple values, all values should be serialized as OTLP log record attributes.

## Actual behavior

The OTLP transport only extracts the first key-value pair from payload.values. All remaining values are silently dropped:

## Environment

- **SDK version:** - 2.4.0
- **SDK instrumentations:** - WebInstrumentation, TracingInstrumentation

PR #1386, opened 2025-07-04
feat: update web-vitals to v5

bfmatei:
## Why

Update `web-vitals` to v5.

## What

Update `web-vitals` to v5.

## Links

https://github.com/grafana/faro-web-sdk/issues/1336

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated

PR #595, opened 2024-05-18
Add web vitals with attribution instrumentation

vitebo:
## Why

This is a feature that will be useful as discussed in this issue https://github.com/grafana/faro-web-sdk/issues/593

## What

- I created a web vitals instrumentation with attribution
- I created a `captureWebVitalsAttribution` option in the `getWebInstrumentations` function to decide which vitals instrumentation to load

## Links

https://github.com/grafana/faro-web-sdk/issues/593

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

Issue #593, opened 2024-05-15
Enhancing WebVitalInstrumentation with Additional Information


vitebo:
## Description

Currently, the WebVitalInstrumentation only sends the metric value. However, in many cases, it would be highly beneficial to include additional data to aid in debugging.

## Proposed Solution

I propose that the SDK should offer two versions of WebVitalInstrumentation. The enhanced version would provide more detailed WebVitals information, utilizing the ['web-vitals/attribution'](https://github.com/GoogleChrome/web-vitals?tab=readme-ov-file#send-attribution-data) feature.

This enhanced instrumentation was initially developed for an internal company project due to the absence of such a feature. Considering its broad applicability, I believe incorporating it into this SDK would be beneficial. I am eager to contribute this solution if there is community interest.

codecapitano:
Thanks so much @vitebo 
I have thought about this for some time as well and consider attribution data this as very valuable. Didn't tackle it due to conflicting priorities. So a contribution would be amazing.

Great initiative, looking forward to your PR 🙏 

PS: 
I am on vacation next week so don't wonder if it takes a bit till I answer.
But @eskirk and @cedricziel can help if you've any questions in the meantime.

PPS:
Out of curiosity, did you build any custom visualizations for web-vitals + attribution data?

Cheers,
Marco

vitebo:
Thank you, Marco!

I think I can work on it this weekend.

> Out of curiosity, did you build any custom visualizations for web-vitals + attribution data?

We haven't built anything very elaborate yet. Creating the visualizations is scheduled for our next sprint. Currently, we only have a basic dashboard that identifies the page elements causing poor INP, which has been our main issue.

<details>
  <summary>INP visualization preview</summary>
  
  https://github.com/grafana/faro-web-sdk/assets/19560693/b33c818f-6222-49dd-a45d-92313c970857

</details>

There's still a lot of room for improvement in the visualization shown in the video above. Once we create better visualizations for web vitals + attribution, I would love to come back and share our results here.

codecapitano:
Hey André thanks for sharing love it.

> I think I can work on it this weekend.

Take all the time you need. It may happen that the review process may take a bit longer as usual due to vacation season.

May I ask another question since your team is working on tackling performance things 🎉 
Du you also use our navigation / resource timings instrumentation?

vitebo:
Not yet. We adopted Grafana Faro recently (less than a month ago), so we are gradually connecting the instrumentations.
[...]

PR #2187, opened 2026-07-24
feat(web-sdk): forward LCP resource details and LoAF-based INP attribution in web vitals

### LINKED ISSUE #2186: Forward LCP resource details and LoAF-based INP attribution in Web Vitals measurements
## Description

The Web Vitals instrumentation forwards a useful subset of the `web-vitals` attribution data, but two groups of fields that make RUM data _actionable_ are currently discarded in `packages/web-sdk/src/instrumentations/webVitals/webVitalsWithAttribution.ts`:

**1. LCP: which resource was the LCP element**

`measureLCP()` does not forward `attribution.url` or anything from `attribution.lcpResourceEntry`. When the LCP element is an image, knowing _which_ resource it was is essential to act on the data (decide what to preload, mark with `fetchpriority`, or fix cache behavior). Today the only hint is the `element` CSS selector, which is unstable across builds when CSS-in-JS is used. Notably, `web-vitals` has provided `url` since v3.0.0 (2022-08, GoogleChrome/web-vitals#244). The data predates Faro's attribution support, but has never been forwarded.

**2. INP: which script caused the slow interaction**

`measureINP()` forwards the phase breakdown (`input_delay`, `processing_duration`, `presentation_delay`) but none of the Long Animation Frame (LoAF) data that `web-vitals` v5 exposes: `longestScript`, `totalScriptDuration`, `totalStyleAndLayoutDuration`, `totalPaintDuration`, `totalUnattributedDuration` (added in GoogleChrome/web-vitals#592). Without it, RUM tells us _that_ an interaction was slow, but not _which script_ caused it, so every regression requires local reproduction with a profiler.

## Proposed solution

**LCP** — forward the resource details, guarded by a new opt-in config flag (default `false`), following the precedent of `trackResources`, which also gates URL-bearing data behind an explicit opt-in:

```ts
webVitalsInstrumentation?: {
  reportAllChanges?: boolean;
  trackLcpAttributionResource?: boolean; // default: false
};
```

- context `resource_url` = `attribution.url`
- context `resource_delivery_type` = `attribution.lcpResourceEntry?.deliveryType` (cache visibility)
- context `resource_initiator_type` = `attribution.lcpResourceEntry?.initiatorType`

**INP** — forward the LoAF summary (no flag, matching the always-on attribution introduced in #1771):

- values: `total_script_duration`, `total_style_and_layout_duration`, `total_paint_duration`, `total_unattributed_duration`, `longest_script_intersecting_duration`
- context: `longest_script_subpart`, `longest_script_invoker`, `longest_script_invoker_type`, `longest_script_source_url`, `longest_script_source_function_name`, `longest_script_source_char_position`

`longAnimationFrameEntries` / `processedEventEntries` (arrays of entries) are intentionally _not_ forwarded since `MeasurementContext` is a flat `Record<string, string>` — the summary above covers the useful part.

Open question: `longest_script_source_url` contains a script URL. These are typically first-party static asset URLs (and Faro already sends script URLs by default in error stack traces), so I propose sending them unguarded — but I'm happy to gate the `longest_script_source_*` fields behind an opt-in flag as well if you prefer symmetry with the LCP side. Users who need stricter policies (e.g. only same-origin URLs, query masking) can post-process or drop fields in `beforeSend`.

I have a working implementation (incl. tests) verified end-to-end in a production-scale app; PR follows.

Example of a real INP measurement with the proposed fields (CPU-throttled to trigger a slow interaction):

```json
{
  "type": "web-vitals",
  "values": {
    "inp": 472,
    "input_delay": 10.7,
    "processing_duration": 109.4,
    "presentation_delay": 351.9,
    "total_script_duration": 61.7,
    "total_style_and_layout_duration": 38,
    "total_paint_duration": 346.9,
    "total_unattributed_duration": 25.4,
    "longest_script_intersecting_duration": 94.4
  },
  "context": {
    "rating": "needs-improvement",
    "longest_script_subpart": "processing-duration",
    "longest_script_invoker": "DIV#root.onclick",
    "longest_script_invoker_type": "event-listener",
    "longest_script_source_url": "https://example.com/static/js/index.js",
    "longest_script_source_function_name": "Yz",
    "longest_script_source_char_position": "36457"
  }
}
```

## Context

- Attribution support was introduced in #595 (requested in #593) and made always-on in #1771; this proposal only extends the set of forwarded fields.
- Data availability in `web-vitals`: LCP `url` / `lcpResourceEntry` since v3.0.0 (2022-08, GoogleChrome/web-vitals#244); `longAnimationFrameEntries` since v4.0.0 (2024-05, GoogleChrome/web-vitals#442); the `longestScript` summary and duration buckets since v5.0.0 (2025-05, GoogleChrome/web-vitals#592). All predate Faro's v5 adoption in #1386, but were not picked up then.
- Related: #2105 (OTLP transport drops all but the first value from `payload.values`). This proposal adds several values, so it may be worth considering that issue for OTLP users; the Faro receiver/Collector path is unaffected.

nus3:
## Why

Closes #2186

The Web Vitals attribution data currently forwarded shows that a metric was slow, but often not what to fix:

- For **LCP**, `attribution.url` and `attribution.lcpResourceEntry` are discarded, so when the LCP element is an image there is no way to tell which resource to preload or mark with `fetchpriority`. The `element` CSS selector is the only hint, and it is unstable across builds when CSS-in-JS is used.
- For **INP**, the LoAF-based attribution that `web-vitals` v5 provides (`longestScript`, duration buckets — GoogleChrome/web-vitals#592) is discarded, so there is no way to tell which script caused a slow interaction without reproducing it locally with a profiler.

## What

**LCP** (`measureLCP()`): forward the LCP resource details behind a new opt-in flag `webVitalsInstrumentation.trackAttributionSources` (default `false` — resource URLs may be sensitive, so this follows the same opt-in approach as `trackResources`):

- context `resource_url` = `attribution.url`
- context `resource_delivery_type` = `attribution.lcpResourceEntry?.deliveryType`
- context `resource_initiator_type` = `attribution.lcpResourceEntry?.initiatorType`

**INP** (`measureINP()`): forward the LoAF summary:

- values (always-on, matching the always-on attribution established in #1771): `total_script_duration`, `total_style_and_layout_duration`, `total_paint_duration`, `total_unattributed_duration`, `longest_script_intersecting_duration`
- context (always-on): `longest_script_subpart`, `longest_script_invoker_type`
- context (gated behind `trackAttributionSources`, since these may contain URLs or DOM identifiers): `longest_script_invoker`, `longest_script_source_url`, `longest_script_source_function_name`, `longest_script_source_char_position`

Notes:

- `longAnimationFrameEntries` / `processedEventEntries` (entry arrays) are intentionally not forwarded: `MeasurementContext` is a flat `Record<string, string>`, and the summary fields are sufficient to act on.
- `longest_script_source_char_position` is stringified explicitly with an `!== undefined` check so that a valid position of `0` survives the truthy check in `addIfPresent`.
- `deliveryType` is not yet in the TS DOM lib, so it is read via a local `PerformanceResourceTiming & { deliveryType?: string }` type.
- All fields are only added when present (`addIfPresent`), so payloads for browsers without LoAF support (or non-resource LCPs) are unchanged.
- No breaking changes; without the new flag, the LCP payload is identical to before.
- Verified end-to-end in a production-scale app (Grafana Faro → Alloy → Loki): real payloads confirmed for both the LCP resource fields and the INP LoAF fields.
- Per review feedback, the INP `longest_script_invoker` / `longest_script_source_*` fields are gated behind the same opt-in flag as the LCP resource fields; the flag was renamed from `trackLcpAttributionResource` to `trackAttributionSources` to reflect that it now covers both.

## Links

- Issue: #2186
- Attribution history in this repo: #593 (request), #595 (introduction), #1771 (always-on), #1386 (web-vitals v5 adoption)
- Data availability in `web-vitals`: `url` / `lcpResourceEntry` since v3.0.0 (GoogleChrome/web-vitals#244); `longAnimationFrameEntries` since v4.0.0 (GoogleChrome/web-vitals#442); `longestScript` + buckets since v5.0.0 (GoogleChrome/web-vitals#592)
- Related: #2105 (OTLP transport drops all but the first value from `payload.values` — relevant for OTLP users since this PR adds several values; the Faro receiver path is unaffected)

## Checklist

- [x] Tests added (opt-in on/off for LCP, LoAF present/absent for INP, empty `deliveryType` handling)
- [x] Changelog updated (via release-please from the PR title, no manual edit)
- [x] Documentation updated (JSDoc on the new config option in `packages/core/src/config/types.ts`, same style as `reportAllChanges`)


packages/web-sdk/src/instrumentations/webVitals/webVitalsWithAttribution.ts
```diff
@@ -86,11 +95,28 @@ export class WebVitalsWithAttribution {
         this.addIfPresent(values, 'input_delay', inputDelay);
         this.addIfPresent(values, 'processing_duration', processingDuration);
         this.addIfPresent(values, 'next_paint_time', nextPaintTime);
+        this.addIfPresent(values, 'total_script_duration', totalScriptDuration);
+        this.addIfPresent(values, 'total_style_and_layout_duration', totalStyleAndLayoutDuration);
+        this.addIfPresent(values, 'total_paint_duration', totalPaintDuration);
+        this.addIfPresent(values, 'total_unattributed_duration', totalUnattributedDuration);
+        this.addIfPresent(values, 'longest_script_intersecting_duration', longestScript?.intersectingDuration);
```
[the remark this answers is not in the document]
nus3: Fixed in fbda7b1.

I intentionally left the pre-existing fields (e.g. `input_delay`, `resource_load_delay`) on the old truthy behavior: they have shipped with these semantics for a while, and silently starting to emit `0` values could surprise downstream dashboards. If the maintainers would like that fixed globally, I think it deserves a separate issue/PR.

packages/web-sdk/src/instrumentations/webVitals/webVitalsWithAttribution.ts
```diff
@@ -86,11 +95,28 @@ export class WebVitalsWithAttribution {
...
+        this.addNumberIfDefined(values, 'total_paint_duration', totalPaintDuration);
+        this.addNumberIfDefined(values, 'total_unattributed_duration', totalUnattributedDuration);
+        this.addNumberIfDefined(values, 'longest_script_intersecting_duration', longestScript?.intersectingDuration);
 
         const context = this.buildInitialContext(metric);
         this.addIfPresent(context, loadStateKey, loadState);
         this.addIfPresent(context, 'interaction_target', interactionTarget);
         this.addIfPresent(context, 'interaction_type', interactionType);
+        this.addIfPresent(context, 'longest_script_subpart', longestScript?.subpart);
+        this.addIfPresent(context, 'longest_script_invoker', longestScript?.entry.invoker);
```
kpelelis: For both LCP symmetry as well as privacy concerns, I think all of `longest_script_invoker`, `longest_script_source_url`, `longest_script_source_function_name` and `longest_script_source_char_position.` should be gated behind the feature flag

kpelelis (changes requested):
Overall good work. Requesting changes just for the gated attributions to be addressed

packages/web-sdk/src/instrumentations/webVitals/webVitalsWithAttribution.ts
nus3: Makes sense, thanks — done in 0a14f14.

 All four fields (`longest_script_invoker`, `longest_script_source_url`, `longest_script_source_function_name`, `longest_script_source_char_position`) are now only forwarded when the opt-in flag is enabled, mirroring the LCP side. `longest_script_subpart` and `longest_script_invoker_type` stay always-on since they are enum-like values (`processing-duration`, `classic-script`, …) with no URLs or DOM identifiers — happy to gate those too if you'd prefer.

 Since the flag now gates both the LCP resource fields and the INP script source fields, I renamed it from `trackLcpAttributionResource` to `trackAttributionSources` and updated its JSDoc to list exactly which fields it controls. If you'd rather keep the original name (or prefer a different one), happy to change it.

kpelelis (approved):

PR #2195, opened 2026-07-29
fix(web-sdk): initialize without throwing when there is no DOM

### LINKED ISSUE #1643: initializeFaro crashes when there is no window
## Description

In a test-context for a Chrome extension I'm working on, there's no `window` object. `initializeFaro` crashes because it's looking at `(window as any).k6` to see if K6 is running.

## Steps to reproduce

1. call `initializeFaro` without a `window`
 
## Expected behavior

`isK6BrowserSession` is `false`

## Actual behavior

Faro throws an exception

## Environment

- **SDK version:** 1.19.0
- **SDK instrumentations:** Web SDK, Web Tracing
- **Device type:** desktop
- **Device name:** MacBook Pro 13"
- **OS:** macOS Sequoia
- **Browser:** Chrome 139

kpelelis:
## Why

`initializeFaro` throws a `ReferenceError: window is not defined` in environments that do not declare `window` — SSR, workers, and test contexts such as the Chrome extension setup in #1643.

The issue points at the k6 session detection, and that is the first thing to crash, but it is not the only one. With just that line fixed, initialization dies next in `performanceObserverSupported`, and after that in `browserMeta`. There are 42 bare `window` references across 13 files in `web-sdk`, so guarding each access individually is not a sensible fix — the web instrumentations and the browser metas genuinely require a DOM.

## What

- `makeCoreConfig` reads the k6 properties off core's `globalObject` instead of `window`, so detection no longer throws where `window` is undeclared.
- When no DOM is present, the DOM-dependent defaults are left out rather than made defensive one call at a time:
  - the `browser` and `page` metas are skipped (`os` and `sdk` remain, `getUAResult` already guards `navigator`);
  - instrumentations named `@grafana/faro-web-sdk:instrumentation-*` are filtered out and a warning is logged once. Instrumentations supplied by the caller pass through untouched, since they may be environment-agnostic.
- New `isBrowserEnvironment()` helper in `packages/web-sdk/src/utils`. It uses `typeof window`/`typeof document` rather than reading off `globalObject` — `typeof` is safe for undeclared identifiers, and going through `globalObject` breaks any test that mocks that module.

Faro now initializes without a DOM and keeps serving everything reported through its API (`pushEvent`, `pushLog`, `pushError`, metas).

## Links

Fixes #1643

## Checklist

- [x] Tests added — `makeCoreConfig.node.test.ts`, a new suite running under `@jest-environment node` so there is a real no-DOM environment; covers k6 detection, instrumentation filtering, pass-through of caller instrumentations, meta selection, and `initializeFaro` end to end

## Verification

- `yarn quality:test` green across all 7 projects.
- `yarn quality:lint:packages` green.
- Playwright smoke suite green in real Chromium (`e2e/smoke`, 8/8). This is the meaningful regression guard here: had the DOM detection been wrong in a browser, every instrumentation would have been silently dropped, and `init.spec.ts` asserts browser meta on `/collect` payloads.
- Reproduced the original crash against the built CJS output in plain node before the fix, and confirmed init, `metas.value`, and the push APIs all work after it.

### DIFF
--- a/packages/web-sdk/src/config/makeCoreConfig.node.test.ts
+++ b/packages/web-sdk/src/config/makeCoreConfig.node.test.ts
@@ -1,0 +1,67 @@
+/**
+ * @jest-environment node
+ */
+import { isFunction } from '@grafana/faro-core';
+
+import { initializeFaro } from '../initialize';
+
+import { makeCoreConfig } from './makeCoreConfig';
+
+const browserConfig = {
+  url: 'http://example.com/my-collector',
+  app: { name: 'no-dom-app' },
+};
+
+describe('makeCoreConfig in an environment without a DOM', () => {
+  it('does not throw while detecting a k6 session', () => {
+    expect(() => makeCoreConfig(browserConfig)).not.toThrow();
+  });
+
+  it('skips the web instrumentations because they depend on DOM APIs', () => {
+    const config = makeCoreConfig(browserConfig);
+
+    expect(config?.instrumentations).toEqual([]);
+  });
+
+  it('keeps instrumentations which are not provided by the web SDK', () => {
+    const customInstrumentation = {
+      name: 'custom-instrumentation',
+      version: '1.0.0',
+      initialize: () => {},
+    };
+
+    const config = makeCoreConfig({
+      ...browserConfig,
+      instrumentations: [customInstrumentation as any],
+    });
+
+    expect(config?.instrumentations).toEqual([customInstrumentation]);
+  });
+
+  it('omits the browser and page metas and keeps the DOM independent ones', () => {
+    const config = makeCoreConfig(browserConfig);
+    const metas = config?.metas.map((item) => (isFunction(item) ? item() : item));
+
+    // the os meta resolves to an empty object because there is no user agent to parse
+    expect(metas).toEqual([{}, { sdk: expect.objectContaining({ name: 'faro-web' }) }]);
+  });
+});
+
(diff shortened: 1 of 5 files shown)


eskirk (changes requested):
@kpelelis one issue that I see is that by filtering out all @grafana/faro-web-sdk:instrumentation- instrumentations you actually lose the session instrumentation, which is needed to set session IDs client side. we may want to filter out another way?

codecapitano:
@kpelelis @eskirk this one needs more discussion.
The PR links an issue which is about running faro inside Chrome extensions which isn't a supported scenario in Faro.
It still keeps adding otel instruments which is no issue in regards to the default http instruments we add bc they don't use window. 

That being said, maybe a better way would be to document that instrumenting extensions isn't supported ootb but that users have option to make it work by leaving out those instruments which use the window object. Sone of our defaults instruments like the perf. instruments or navigation instrument don't even make sense in  a browser env. 

Wdyt?

eskirk:
@codecapitano generally agree. running faro without a DOM really only happens for browser extensions and I am not sure we need to change our core code in order to accommodate for this... I think this should be a documentation change.

codecapitano:
@kpelelis @eskirk 
Here are the doc PRs
- https://github.com/grafana/website/pull/33226
- https://github.com/grafana/faro-web-sdk/pull/2227

kpelelis:
Closing off in favour of the documentation added by @codecapitano

PR #2196, opened 2026-07-29
fix(core): keep config.paused in sync with pause and unpause

### LINKED ISSUE #205: `window.faro.pause()` does not `update faro.config.paused`
When pausing / unpausing a faro instance using the global object, despite of using the default `window.faro`, or passing a `globalObjectKey`, does not update the value found in `config.paused` on the global object.

Example: 

If I run: `window.faro.pause()`
then inspect the global instance: `window.faro.config.paused`, the value is still the default: `false`

## Steps to reproduce

1. Initialize faro: 

import {
  initializeFaro,
  getWebInstrumentations,
  ReactIntegration,
  ReactRouterVersion,
  Faro,
} from '@grafana/faro-react';
import { TracingInstrumentation } from '@grafana/faro-web-tracing';
import { Route, useHistory } from 'react-router-dom';

// .....

const faro = initializeFaro({
    url: 'some-url',
    apiKey: 'some-api-key',
    instrumentations: [
      ...getWebInstrumentations({
        captureConsole: true,
      }),
      new TracingInstrumentation({
        resourceAttributes: {
          'service.name': 'some-service-name',
          'team.name': 'some-team-name',
        },
      }),
      new ReactIntegration({
        router: {
          version: ReactRouterVersion.V5,
          dependencies: {
            history: useHistory,
            Route,
          },
        },
      }),
    ],
    session: (window as any).__PRELOADED_STATE__?.faro?.session,
    app: {
      name: 'some-app-name',
      version: 'some-app-version',
      environment: undefined,
    },
    user: {
      username: 'some-username',
    },
  });

2. Pause the instance: `window.faro.pause()``

3. Check if the value changes in the global instance: `window.faro.config.paused`

## Expected behavior

`window.faro.config.paused` should be updated upon pausing / unpausing the instance.

## Actual behavior

I cannot infer if an instance is paused. 

## Environment
SDK version: 1.0.5
SDK instrumentations: React
Device type: desktop
Device name: MacBook Air M1 13"
**OS:**MacOS
**Browser:**Chrome

kpelelis:
## Why

`faro.pause()` and `faro.unpause()` did not update `faro.config.paused`, so anything inspecting the instance — including `window.faro.config.paused`, which is how #205 was reported — kept reading the value the SDK was initialized with. Ingestion did stop correctly; only the observable state was stale.

`initializeTransports` tracked the paused state in a local `let paused`, a second copy of `config.paused` that nothing ever wrote back.

## What

Use `config.paused` as the single source of truth in `initializeTransports` and drop the local mirror. `config` is the same object reference that `registerFaro` exposes as `faro.config`, so `faro.config.paused`, `faro.transports.isPaused()`, and the global instance now agree.

No behavioural change to ingestion — `execute` reads the same flag it always did, just from one place.

## Links

Fixes #205

## Checklist

- [x] Tests added — unit tests in `packages/core/src/faro.test.ts` covering both transitions and the initialized-paused case, plus a Playwright spec asserting the reported symptom in a real browser

## Verification

- `yarn quality:test` green across all 7 projects; `yarn quality:lint:packages` green.
- Playwright smoke suite green in real Chromium, 9/9 including the new `pause.spec.ts`.
- The new e2e spec was checked to be non-vacuous: reverting `transports/initialize.ts` to `main` and rebuilding makes it fail on `Expected: true, Received: false`, and it passes with the fix in place.

### DIFF
--- a/e2e/smoke/tests/pause.spec.ts
+++ b/e2e/smoke/tests/pause.spec.ts
@@ -1,0 +1,32 @@
+import type { Faro } from '@grafana/faro-core';
+
+import { expect, test } from './fixtures';
+
+declare global {
+  interface Window {
+    faro: Faro;
+  }
+}
+
+test.describe('Smoke / pause', () => {
+  test('exposes the paused state on the global faro instance', async ({ page, collector }) => {
+    await page.goto('/');
+
+    const readConfigPaused = () => page.evaluate(() => window.faro.config.paused);
+
+    expect(await readConfigPaused()).toBe(false);
+
+    await page.evaluate(() => window.faro.pause());
+    expect(await readConfigPaused()).toBe(true);
+
+    await page.evaluate(() => window.faro.unpause());
+    expect(await readConfigPaused()).toBe(false);
+
+    // ingestion resumes once unpaused
+    await page.locator('[data-cy="btn-push-log"]').click();
+
+    const log = await collector.waitForMatch((b) => b.logs?.find((l) => l.message === 'smoke harness log'));
+
+    expect(log.level).toBe('info');
+  });
+});
--- a/packages/core/src/faro.test.ts
+++ b/packages/core/src/faro.test.ts
@@ -75,4 +75,36 @@ describe('faro', () => {
     expect(items[0]?.payload.message).toEqual('test1');
     expect(items[1]?.payload.message).toEqual('test3');
   });
+
+  it('reflects the paused state in config.paused', () => {
+    const faro = initializeFaro(
+      mockConfig({
+        isolate: true,
+        transports: [new MockTransport()],
+      })
+    );
+
+    expect(faro.config.paused).toBe(false);
+
+    faro.pause();
+    expect(faro.config.paused).toBe(true);
+    expect(faro.transports.isPaused()).toBe(true);
+
+    faro.unpause();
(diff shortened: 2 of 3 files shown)


domasx2:
I'm doubtful about the rationale. Imho it's reasonable to expect the config object to be immutable and reperesent initial configuration. Setting runtime state on it sets an odd precendent. Maybe instead rename `config.paused` to `config.pausedOnStart` or `initiallyPaused` or smth, wdyt? also replied on original issue

domasx2:
or `faro.config` -> `faro.initialConfig`

codecapitano:
> I'm doubtful about the rationale. Imho it's reasonable to expect the config object to be immutable and reperesent initial configuration. Setting runtime state on it sets an odd precendent. Maybe instead rename `config.paused` to `config.pausedOnStart` or `initiallyPaused` or smth, wdyt? also replied on original issue

`config` is read-only for consumers and `pause()` / `unpause()` are the supported mutating APIs. 
Keeping `config.paused` synchronized makes it the effective state observable.

codecapitano:
> or `faro.config` -> `faro.initialConfig`

This is a breaking(-ish) change

codecapitano (approved):

PR #2198, opened 2026-07-29
fix(web-tracing): stop injecting traceparent while Faro is paused

### LINKED ISSUE #710: Traceparent header is still added to requests when the faro instance is paused
## Steps to reproduce

1. Init faro
2. Pause faro `.pause()`
3. Make any request
4. Notice the tracing instrumentation from OpenTelemetry is not paused -> `traceparent` header is still added to those requests

## Expected behavior

When the faro instance is in the paused state, tracing instrumentation should also be paused (don't add `traceparent` header on following requests).

## Actual behavior

When the faro instance is in the paused state, tracing instrumentation is NOT paused (`traceparent` header is still added to the following requests).

## Environment

- **SDK version:** 1.10.0
- **SDK instrumentations:** ReactIntegration, TracingInstrumentation, ViewInstrumentation, SessionInstrumentation, WebVitalsInstrumentation, ErrorsInstrumentation, PerformanceInstrumentation
- **Device type:** laptop
- **Device name:** Macbook Pro 16-inch, 2023
- **OS:** Sonoma 14.6.1
- **Browser:** Chrome Version 128.0.6613.138 (Official Build) (arm64)

## Context

Having some customers that don't want any kind of tracking enabled (switched dinamically) and seeing the `traceparent` header on requests might get them suspicious.

kpelelis:
## Why

`faro.pause()` stops signals from being sent, but the OTel propagator is registered globally and kept injecting `traceparent` into every outgoing request (#710). A paused SDK therefore still looked like it was tracking.

That matters beyond tidiness — the reporter pauses Faro specifically for customers who do not want tracking enabled, and a `traceparent` header on their requests makes it look like tracking continued.

Note that turning off sampling would not fix this: a non-recording span still has a valid span context, so the W3C propagator writes a `traceparent` with the sampled flag cleared. The header has to be suppressed at the propagator.

## What

Wrap the propagator in `FaroPausablePropagator`, which skips `inject` while `transports.isPaused()` and otherwise delegates.

- A caller-supplied `options.propagator` is wrapped too, so pausing behaves the same with a custom propagator.
- The paused state is read per call, not captured at construction, so pause/unpause take effect immediately.
- `extract` stays enabled while paused: reading an incoming trace context sends nothing, and dropping it would detach a resumed session from the trace it belongs to.
- `fields()` delegates, so CORS header allowlisting is unaffected.

`FaroPausablePropagator` is exported, matching how the other Faro span processors and exporters in this package are exposed.

## Links

Fixes #710

## Checklist

- [x] Tests added — unit tests for the wrapper (including a real `W3CTraceContextPropagator` with a live span context, asserting the exact `traceparent` is written when active and nothing at all when paused), plus a Playwright spec asserting the reported symptom in a real browser

## Verification

- New `e2e/smoke/tests/tracing-pause.spec.ts` drives a real fetch in Chromium and inspects the outgoing request headers: `traceparent` present, `pause()`, header absent, `unpause()`, header present again. It needed a `Traced fetch` button in the harness.
- Verified non-vacuous: with the propagator wiring reverted to `main` the spec fails with a `traceparent` of `00-28d58bf6c9064146d5478949a223f32a-a1ef2034909927ff-01` on the paused request.
- `yarn quality:test` green across all 7 projects; `yarn quality:lint:packages` green; full Playwright suite 9/9.

### DIFF
--- a/e2e/smoke/src/main.tsx
+++ b/e2e/smoke/src/main.tsx
@@ -70,6 +70,9 @@ function App() {
       <button data-cy="btn-push-error" onClick={() => faro.api.pushError(new Error('smoke harness pushError'))}>
         Push error via API
       </button>
+      <button data-cy="btn-traced-fetch" onClick={() => void fetch('/api/ping').catch(() => undefined)}>
+        Traced fetch
+      </button>
       <button
         data-cy="btn-emit-span"
         onClick={() => {
--- a/e2e/smoke/tests/tracing-pause.spec.ts
+++ b/e2e/smoke/tests/tracing-pause.spec.ts
@@ -1,0 +1,39 @@
+import type { Faro } from '@grafana/faro-core';
+
+import { expect, test } from './fixtures';
+
+declare global {
+  interface Window {
+    faro: Faro;
+  }
+}
+
+test.describe('Smoke / tracing while paused', () => {
+  test('stops adding traceparent to requests while paused', async ({ page }) => {
+    const traceparents: Array<string | undefined> = [];
+
+    await page.route('**/api/ping', async (route) => {
+      traceparents.push(route.request().headers()['traceparent']);
+      await route.fulfill({ status: 200, body: '{}' });
+    });
+
+    await page.goto('/');
+
+    const fetchOnce = async () => {
+      const request = page.waitForRequest('**/api/ping');
+      await page.locator('[data-cy="btn-traced-fetch"]').click();
+      await request;
+    };
+
+    await fetchOnce();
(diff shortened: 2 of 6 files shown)


kpelelis:
Heads-up on an overlap: #2178 (draft) reworks the same `provider.register()` call in `packages/web-tracing/src/instrumentation.ts` that this PR wraps, so whichever of the two lands second will need a rebase.

The conflict is only textual — the pausable propagator is orthogonal to the single-tracing-owner model. Under #2178 the wrapper would just move to wherever the owning instance registers the provider, and `isPaused()` should then read from that owning instance's transports.

Happy to rebase this onto #2178 instead if @martinkuba would rather land that first.

domasx2 (approved):
lgtm

martinkuba (approved):

Issue #1898, opened 2026-02-14
fetch keepalive can fail silently when cumulative body size exceeds browser limit


martinkuba:
## Description

The [FetchTransport](https://github.com/grafana/faro-web-sdk/blob/main/packages/web-sdk/src/transports/fetch/transport.ts) currently enables keepalive based on [individual request body size](https://github.com/grafana/faro-web-sdk/blob/main/packages/web-sdk/src/transports/fetch/transport.ts#L74).

Browsers, however, enforce a cumulative 64KiB limit across all pending keepalive requests from the page. When multiple requests are in flight concurrently, their combined body sizes can exceed this limit even though each individual request is under 60KB.

NOTE: This limit is shared with any other keepalive requests on the page (including navigator.sendBeacon calls from other libraries).

## Expected behavior

While we cannot track the cumulative body size across all requests, we can track the cumulative size of requests made by Faro. This decreases the odds of this issue occurring.

In addition, we could consider retrying failed requests. This would serve as a safety net for cases where the cumulative limit is exceeded by other code on the page that Faro cannot track

For reference, the OpenTelemetry JS SDK recently implemented this approach: https://github.com/open-telemetry/opentelemetry-js/pull/6391

kpelelis:
Triaging this: both asks in the description already landed in #2151 (merged 2026-06-26) — `FetchTransport.reserveKeepalive` now tracks cumulative pending body size and request count before enabling keepalive, and a failed keepalive request is retried once with `keepalive: false` as the safety net for budget consumed by other code on the page.

One gap remained in that implementation, which #2199 fixes: the reservation measured `String.length` (UTF-16 code units) against what is a **byte** limit in the browser, so payloads of CJK, emoji or accented text were undercounted by up to 3x — a 20,000 code unit CJK body is really 60,000 bytes.

Suggest closing this once #2199 lands.

PR #2199, opened 2026-07-29
fix(web-sdk): measure the keepalive budget in bytes

kpelelis:
## Why

While picking up #1898 I found that **both things the issue asks for already landed** in #2151 (merged 2026-06-26, five months after the issue was filed): `reserveKeepalive` tracks cumulative pending body size and request count, and a failed keepalive request is retried once with `keepalive: false`. The issue was simply never closed. So this PR is not that work.

What is still wrong is the unit of measurement. The browser keepalive budget is a **byte** limit, but the reservation used `String.length`, which counts UTF-16 code units:

```
payload            codeUnits   utf8Bytes   ratio
ascii error             1014        1014    1.00
accented (de/fr)        1014        2014    1.99
emoji message           1014        2014    1.99
CJK message             1014        3014    2.97
```

So a payload of CJK, emoji or accented text — an error message, a user attribute, a log line — was undercounted by up to 3x. A 20,000 code unit CJK body is really 60,000 bytes, passes the individual check, and three of them reserve 60KB of budget while actually sending 180KB. That is over the shared 64KiB limit and produces exactly the silent keepalive failure the budget was added to prevent.

## What

Measure the encoded byte length in `getBodyByteSize`, falling back to `String.length` where `TextEncoder` is unavailable.

The compressed path was already correct — it reserves `Blob.size`, which is in bytes.

## Links

Refs #1898

**#1898 can be closed** once this lands: the cumulative tracking and the retry safety net it asks for are on `main` already via #2151, and this PR fixes the remaining accounting gap in them.

## Checklist

- [x] Tests added — a non-ASCII payload that is under 60,000 code units but over 60,000 bytes must not get keepalive. Verified non-vacuous: the test reports `Expected: false, Received: true` against `main`.

## Verification

- `yarn quality:test` green across all 7 projects; `yarn quality:lint:packages` green.
- The existing keepalive tests (individual limit, cumulative budget, retry fallback, compression threshold) are unchanged and still pass.

### DIFF
--- a/packages/web-sdk/src/transports/fetch/transport.ts
+++ b/packages/web-sdk/src/transports/fetch/transport.ts
@@ -23,6 +23,20 @@ interface KeepaliveReservation {
   release: () => void;
 }
 
+/**
+ * The browser keepalive budget is measured in bytes, while `String.length` counts UTF-16 code
+ * units. A payload of non-ASCII text is up to three times larger than its length suggests, so
+ * reserving by length lets Faro send far more than it accounted for and reintroduces the silent
+ * keepalive failures this budget exists to prevent.
+ */
+function getBodyByteSize(body: string): number {
+  if (typeof TextEncoder === 'undefined') {
+    return body.length;
+  }
+
+  return new TextEncoder().encode(body).byteLength;
+}
+
 export class FetchTransport extends BaseTransport {
   readonly name = '@grafana/faro-web-sdk:transport-fetch';
   readonly version = VERSION;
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -85,7 +99,7 @@ export class FetchTransport extends BaseTransport {
         }
 
         let body: string | Blob = jsonBody;
-        let bodySize = jsonBody.length;
+        let bodySize = getBodyByteSize(jsonBody);
```
domasx2, line 102: nit: this is non trivial calculation that  just gets overrwritten on line 107 if compression is enabled. could easily refactor with if/else

domasx2 (commented):
seems correct, but one nit

domasx2 (approved):

PR #2200, opened 2026-07-29
fix(web-tracing): type the full set of fetch and xhr instrumentation options

### LINKED ISSUE #1464: web-tracing package: Add missing properties to types for fetch and xhr instrumentation options
## Why

The fetch and xhr instrumentations allow much more config options as exposed by the respective Faro types.
Faro is already forwarding all necessary config options to the underlying instruments and only the typings are missing.

## What
Update the [config options types](https://github.com/grafana/faro-web-sdk/blob/d6a64712b5c7c181917468d58d6e7d3b62e3e235/packages/web-tracing/src/types.ts#L28) to include the missing properties.

kpelelis:
## Why

`fetchInstrumentationOptions` and `xhrInstrumentationOptions` were typed as hand-written pairs exposing only `applyCustomAttributesOnSpan` and `ignoreNetworkEvents`, while `getDefaultOTELInstrumentations` spreads every key onto the underlying OTel instrumentation. So `clearTimingResources`, `measureRequestSize`, `semconvStabilityOptIn`, `propagateTraceHeaderCorsUrls` and `enabled` all worked at runtime but did not compile (#1464).

## What

Derive both option types from the OTel config types — `FetchInstrumentationConfig` and `XMLHttpRequestInstrumentationConfig` — so they cannot drift from what is actually forwarded.

`ignoreUrls` is omitted from both. Faro derives it from the configured transports, and because the caller options are spread *after* Faro's, a caller-supplied `ignoreUrls` would silently replace it and let Faro trace its own collector endpoint. The existing top-level `ignoreUrls` is the supported way in, and it is merged with Faro's own.

### One real bug, not just a typing gap

`requestHook` was not merely untyped — it was broken. `createFetchInstrumentationOptions` assigns Faro's own `requestHook` *after* spreading the caller options:

```js
return {
  ...sharedOptions,
  ignoreNetworkEvents: true,
  ...fetchInstrumentationOptions,   // caller's requestHook lands here
  applyCustomAttributesOnSpan: ...,
  requestHook: (span, _) => { ... }, // ...and is then overwritten
};
```

So a caller-supplied `requestHook` was silently dropped. Simply typing it would have advertised an option that does nothing, so this PR composes the two — Faro's user-action attributes are applied, then the caller's hook runs — matching how `applyCustomAttributesOnSpan` is already handled via `fetchCustomAttributeFunctionWithDefaults`.

## Links

Fixes #1464

## Checklist

- [x] Tests added — one asserting the newly typed options reach the instrumentations, one asserting a caller `requestHook` is invoked with the span and request

## Verification

- Both tests verified non-vacuous. Against `main` the options test fails to compile with `TS2353: 'clearTimingResources' does not exist in type '{ applyCustomAttributesOnSpan?: ...; ignoreNetworkEvents?: boolean }'` for fetch and xhr, and `TS2353` for `requestHook`; the hook test fails with `Number of calls: 0`.
- `yarn build` clean, `yarn quality:test` green across all 7 projects, `yarn quality:lint:packages` green.
- Deliberately avoids `instrumentation.ts`, so it does not collide with #2178.


packages/web-tracing/src/getDefaultOTELInstrumentations.test.ts
```diff
@@ -63,4 +63,55 @@ describe('getDefaultOTELInstrumentations', () => {
...
+
+  // The option types used to expose only applyCustomAttributesOnSpan and ignoreNetworkEvents even
+  // though every key is spread onto the OTel instrumentation. See issue #1464.
+  it('forwards the full set of OTel instrumentation options', () => {
+    getDefaultOTELInstrumentations({
+      fetchInstrumentationOptions: {
+        clearTimingResources: true,
+        measureRequestSize: true,
+        semconvStabilityOptIn: 'http',
+        enabled: false,
```
domasx2: claude flagged a gotcha: this flag means "has it been enabled" rather than "it should be enabled". So if you provide `true`, it _**wont be enabled**_. Maybe best to omit it entirely?

domasx2 (changes requested):
have a concern about misleading `enabled` falg, see comment

packages/web-tracing/src/getDefaultOTELInstrumentations.test.ts
codecapitano: Agreed, and it is worse than a naming problem. I have omitted `enabled` from both option types in e103250e.

One correction to the mechanics, because the flag fails in the opposite direction to what the description suggests. Passing `true` does enable the instrumentation. It is passing `false` that does not disable it.

In `@opentelemetry/instrumentation@0.221.0`:

- `platform/browser/instrumentation.js`: the constructor runs `if (this._config.enabled) this.enable()`, and `setConfig` defaults `enabled` to `true`.
- `autoLoaderUtils.js`: `enableInstrumentations` runs `if (!instrumentation.getConfig().enabled) instrumentation.enable()`. The comment there explains why, which is to avoid double enabling something the constructor already enabled.

Faro always constructs and then registers (`getDefaultOTELInstrumentations.ts` line 19, then `instrumentation.ts` line 140), so:

- `enabled: true`: enabled by the constructor, which is also the default.
- `enabled: false`: skipped by the constructor, then enabled by `registerInstrumentations`.

Either value ends up enabled. The flag cannot change behaviour in Faro's setup at all, so advertising it in a public type would be misleading in both directions.

The supported way to turn fetch or XHR tracing off is to pass your own `instrumentations` array to the tracing instrumentation. That is now documented on the option itself.

codecapitano:
I pushed three commits to this branch, plus a merge of `main`. Summary below so nothing is a surprise. Happy to drop any of it if you disagree.

## The branch no longer compiled against `main`

This is the one nobody raised in review, and it blocked the merge rather than just the review.

The new test passed `semconvStabilityOptIn: 'http'` for both fetch and XHR. That key was part of `FetchInstrumentationConfig` at the commit this branch started from, which locked OpenTelemetry `0.219.0`. It was removed in `0.221.0`, which `main` locks today after #2154. Since the option types are now derived with `Omit<FetchInstrumentationConfig, ...>`, the object literal in the test hits an excess property check, and Jest type checks through `ts-jest` and `tsconfig.spec.json`. The green checks on this PR were running against the old lockfile, so merging would have broken `main`.

I merged `main` into the branch (no rebase, so your commit is untouched) and removed the key from the test. Nothing shrank for users. The type is derived from the OpenTelemetry config, so if that key comes back, it comes back automatically. That is exactly the drift protection this PR set out to add.

## `enabled` omitted

See the reply on Domas's thread for the details. Short version: the flag cannot disable the instrumentation in Faro's setup, so it is now omitted from both option types and the reason is documented on the option.

## The `requestHook` test only proved half of its name

`runs a caller supplied fetch requestHook alongside the Faro one` supplied no active user action and asserted only the caller hook. Replacing the composed hook with the caller hook alone would still have passed, which would silently drop the Faro user action attributes from every span whenever a caller configures `requestHook`. That is the exact regression the composition exists to prevent.

The test now starts a real user action and asserts both halves. Verified by removing the two `setAttribute` calls, which fails the test.

## Naming the composition

`applyCustomAttributesOnSpan` composes through the named `fetchCustomAttributeFunctionWithDefaults`, while `requestHook` composed inline in an anonymous closure. That left two mechanisms for the same rule side by side, each with its own explanatory comment. I extracted `fetchRequestHookWithDefaults` so both keys read the same way. It stays in `getDefaultOTELInstrumentations.ts` rather than moving to `instrumentationUtils.ts`, which holds pure OpenTelemetry span helpers and would otherwise gain a dependency on `faro`.

Smaller cleanups in the same commit: the test uses a real user action from `faro.api.startUserAction` instead of a partial object behind a double cast, reuses one options constant instead of repeating the same literal four times, and reads the mock argument without a tuple cast.

## One documentation fix

The new comment told readers to use "the `ignoreUrls` above". That field is unreachable for anyone configuring through `TracingInstrumentation`, because `instrumentationOptions` omits it. The real entry point is the `ignoreUrls` option in the Faro config, which `FetchTransport.getIgnoreUrls` merges with the collector URL. The comment now points there.

## Possible follow up, not done here

`Omit` is erased at runtime, so a caller using plain JavaScript, or whose config arrives as `any`, can still pass `ignoreUrls` and replace the transport derived list. Stripping both keys in the builders would make the constraint hold regardless of the caller's type checking. That changes runtime behaviour, so I left it out of this PR. Worth a separate issue if you think it is worth doing.

domasx2 (approved):

PR #2239, opened 2026-08-18
fix(user-actions): trigger user actions from clicks on nested child elements

digitalcrab:
## Why

data-faro-user-action-name only matched event.target directly. Grafana UI's Button renders
its label inside a span, so a click there never matched the button's attribute and the
action was silently dropped - the report is exactly this case: `<button data-faro-user-action-name="save"><span>Save</span></button>`.

## What

processUserEvent now walks up with closest() to find the attribute on an ancestor, not just
the exact click target. That surfaced two more bugs on the way, so I fixed all three:

- closest() returns null when nothing matches - the old code read element.dataset directly,
  so every untracked click threw.
- the closest() selector was built from the raw config value instead of the normalized
  data- attribute, so a camelCase dataAttributeName config never matched anything.
- getUserActionNameFromElement now accepts a null element and returns undefined instead of
  throwing.

Added tests for the nested-element case, the no-ancestor case, and null handling.

## Checklist

- [x] Tests added
- [x] Changelog updated
- [ ] Documentation updated

### DIFF
--- a/packages/web-sdk/src/instrumentations/userActions/processUserActionEventHandler.ts
+++ b/packages/web-sdk/src/instrumentations/userActions/processUserActionEventHandler.ts
@@ -6,6 +6,7 @@ import { UserActionController } from './userActionController';
 import {
   convertDataAttributeName,
   deriveUserActionTimeoutDataAttribute,
+  normalizeDataAttributeName,
   normalizeInitialActivityTimeout,
   type TimeoutWarning,
 } from './util';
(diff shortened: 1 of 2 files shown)


packages/web-sdk/src/instrumentations/userActions/processUserActionEventHandler.ts
```diff
@@ -25,12 +26,18 @@ export function getUserEventHandler(
...
   function processUserEvent(event: PointerEvent | KeyboardEvent): void {
-    const element = event.target as HTMLElement;
     const dataAttributeName = config.userActionsInstrumentation?.dataAttributeName ?? userActionDataAttribute;
+
+    const target = event.target;
+    if (!(target instanceof Element)) {
+      return;
+    }
+
+    const element = target.closest(`[${normalizeDataAttributeName(dataAttributeName)}]`) as HTMLElement | null;
```
kpelelis: Since the input of target.closest is not CSS escaped, there is a case where this could throw and unhandled exception from a missconfigured SDK. 

For example
```lang=javascript
initializeFaro({
  url: 'http://localhost:1234/collect',
  app: { name: 'faulty-ua' },
  userActionsInstrumentation: { dataAttributeName: 'data-foo]' },
});
```

kpelelis (changes requested):
LGTM. Nice finding!! There is only a data sanitisation issue, but other than that, happy to merge

kpelelis (approved):

Issue #2241, opened 2026-08-20
Epic: Improve session replay delivery reliability


Blinkuu:
## Summary

Session replay is an ordered stream of checkpoints and incremental changes. Losing ordinary telemetry is usually tolerable; losing replay events can leave a recording incomplete or impossible to play.

Improve the browser delivery path so it handles transient collector failures, rate limiting, page lifecycle changes, and unrecoverable gaps according to the stronger reliability requirements of session replay.

## Goals

- Retry transient delivery failures with bounded, response-aware policies.
- Retain replay batches during temporary backoff instead of silently dropping them.
- Give replay events stable delivery identity across retries.
- Recover from exhausted delivery by starting a fresh replay checkpoint.
- Offer opt-in durable delivery through the browser Service Worker API.
- Keep ordinary Faro telemetry on a lightweight best-effort path.

## Non-goals

- Promise delivery when the browser cannot execute SDK or Service Worker code.
- Make all Faro signals durable by default.
- Add unbounded retries or unbounded browser storage.
- Change rrweb playback semantics.

## Success criteria

- A short network or retryable HTTP failure does not lose the affected replay batch.
- Rate limiting does not silently discard replay events produced during backoff.
- Retried replay events retain stable identity.
- Exhausted delivery produces a bounded recording gap followed by a fresh playable checkpoint.
- Opt-in Service Worker delivery survives page termination or reload within documented storage and browser limits.
- Retry, exhaustion, recovery, and durable-outbox behavior are covered by deterministic and browser-level tests.

PR #2247, opened 2026-08-20
fix(web-sdk): retry transient Faro delivery failures

### LINKED ISSUE #2243: Retry transient Faro delivery failures
## What to build

Make Fetch transport delivery resilient to short-lived network and collector failures. A rejected batch remains the unit of retry, uses the same serialized body on every attempt, and either reaches an accepted response or ends with an explicit permanent or exhausted outcome.

The policy should remain conservative for ambiguous browser network failures until end-to-end replay delivery is idempotent.

## Acceptance criteria

- [ ] Successful HTTP responses complete delivery without retrying.
- [ ] Network failures, SDK-created request timeouts, HTTP 408/425/429, and retryable 5xx responses retry with bounded exponential backoff and jitter.
- [ ] HTTP 429 and 503 honor a valid `Retry-After` response while retaining the rejected batch.
- [ ] Permanent 4xx responses stop immediately, and malformed `Retry-After` values use the bounded fallback policy.
- [ ] Ambiguous network failures perform no more than the currently supported single retry; response-known failures follow bounded attempt and elapsed-time budgets.
- [ ] Large session replay checkpoints that do not use keepalive retain their complete body across a retry.
- [ ] Keepalive budgeting measures transmitted bytes for non-ASCII and compressed bodies.
- [ ] Deterministic tests cover accepted, retryable, permanent, timeout, exhausted, and `Retry-After` outcomes.

## Blocked by

None — can start immediately.

Blinkuu:
## Why

Short-lived network, collector, and rate-limit failures currently lose the affected Faro batch. This is especially damaging for session replay checkpoints and incremental events, where one lost batch can leave a recording incomplete.

Fetch v2 adds opt-in reliable delivery while leaving the existing Fetch transport unchanged as the default.

## What

- add a separate internal Fetch v2 implementation selected only by `experimental.fetchTransportV2`
- retry HTTP 408, 425, 429, 500, 502, 503, and 504 responses plus retryable network failures
- apply one bounded maximum-attempt budget to every retryable failure, counting every raw Fetch request, including the non-keepalive compatibility fallback, as an attempt
- use internal defaults for exponential backoff, jitter, maximum backoff, admission capacity, and concurrency
- share a 10-second timeout budget across a keepalive request and its compatibility fallback when `AbortController` is available; otherwise warn and use browser-managed Fetch timing
- prepare the body, resolved headers, session identity, and `Idempotency-Key` once per batch and preserve them across every attempt
- budget keepalive requests using transmitted bytes, including non-ASCII and compressed bodies
- honor `Retry-After` on HTTP 429 and 503 using the RFC 9110 delta-seconds and HTTP-date formats; drop the batch when the stated delay or numeric overflow exceeds the internal ceiling
- retain each accepted batch for its complete delivery lifecycle while declining new work when admission capacity is exhausted
- schedule waiting batches by their individual retry deadlines and space redeliveries when their Fetch attempt actually starts, including behind saturated concurrency
- flush waiting batches once on page hide, avoid further unload retries, and resume normal retry behavior after bfcache restoration
- keep admission, backoff, throttling, and scheduling independent of Fetch so the delivery queue can be reused by another transport
- emit payload-free terminal diagnostics while keeping intermediate retry activity at debug level
- keep Fetch v2 and its retry policy internal; enabling the experimental flag uses the internal defaults

## Verification

- Fetch v2 transport, RFC 9110 HTTP-date parser, and configuration factory tests: 122 passed
- page lifecycle tests with `PageTransitionEvent` unavailable: 3 passed
- Core and Web SDK TypeScript checks: passed
- repository-wide lint and diff checks: passed

## Links

- Related to #2241
- Closes https://github.com/grafana/faro-web-sdk/issues/2243

## Checklist

- [x] Tests added
- [ ] Changelog updated — release automation owns `CHANGELOG.md`
- [x] Experimental flag documented in the configuration type


packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -138,25 +205,275 @@ export class FetchTransport extends BaseTransport {
...
+      return undefined;
+    }
+
+    if (/^\d+$/.test(retryAfterHeader)) {
+      const delayMs = Number(retryAfterHeader) * 1000;
+      return Number.isFinite(delayMs) ? delayMs : undefined;
+    }
+
+    let date: number;
+    const asctimeMatch = ASCTIME_DATE_PATTERN.exec(retryAfterHeader);
```
codecapitano: _AI-assisted review (Claude Code + Codex cross-model review)._

**High:** The asctime day group here accepts any two digits (00-99), not just valid calendar days. A header like `Sun Dec 99 00:00:00 2026` passes this check, and `Date.UTC` then normalizes day 99 into a real date about three months later. A malformed or malicious `Retry-After` value can push `disabledUntil` far into the future and drop every batch until then, instead of falling back to the bounded default backoff.

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -121,12 +186,14 @@ export class FetchTransport extends BaseTransport {
           ...(requestOptionsWithoutKeepalive ?? {}),
         };
 
-        return this.fetchWithKeepaliveRetry(url, requestInit, bodySize, configuredKeepalive).catch((err) => {
-          this.logError('Failed sending payload to the receiver\n', JSON.parse(jsonBody), err);
-        });
+        return this.fetchWithRetry(url, requestInit, bodySize, configuredKeepalive, callerSignal);
```
codecapitano: _AI-assisted review (Claude Code + Codex cross-model review)._

**High:** Each buffered task now stays occupied for the entire retry lifecycle, up to `maxElapsedTimeMs` (120 seconds by default), instead of a single fetch call. With the default concurrency of 5 and buffer size of 30, a sustained collector outage or a burst of 503/`Retry-After` responses can fill the buffer within seconds. After that, every new batch is dropped with `Task buffer full` instead of being retried, which is a bigger loss than before this change, for the exact failure mode the retry policy is meant to protect against.

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -242,8 +561,8 @@ export class FetchTransport extends BaseTransport {
     }
 
     if (response.status === TOO_MANY_REQUESTS) {
-      this.disabledUntil = this.getRetryAfterDate(response);
-      this.logWarn(`Too many requests, backing off until ${this.disabledUntil}`);
+      this.disabledUntil = new Date(this.getNow() + (this.getRetryAfterDelayMs(response) ?? this.rateLimitBackoffMs));
```
codecapitano: _AI-assisted review (Claude Code + Codex cross-model review)._

**Medium:** `disabledUntil` is shared, transport-wide state, but it is only checked at the start of a new `send()` call. A batch's own retry loop schedules its next attempt purely from its own last `Retry-After` value and never re-checks `disabledUntil` before firing. So if one batch is told to wait 1 second and another concurrently gets told to wait 60 seconds, the first batch's retry still goes out at the 1 second mark, even though the transport has otherwise decided to back off from the collector for a minute.

packages/web-sdk/src/transports/fetch/transport.ts
codecapitano: Fixed in a54d72b8: `Retry-After` dates are now round-tripped through `Date.UTC` and rejected if the parsed components don't match the input, so an out-of-range day like 99 falls back to the bounded default backoff instead of a bogus far-future date. Covered by the new malformed-date test cases.

packages/web-sdk/src/transports/fetch/transport.ts
codecapitano: Fixed in a54d72b8: the buffer slot is now acquired per logical attempt instead of for the whole retry lifecycle, and a new `allowOverflow` option on `promiseBuffer.add` lets an already-admitted retry continue even if new sends fill the buffer in the meantime. Covered by the new tests "releases request capacity while a batch waits to retry" and "retains a retry when new requests fill the buffer".

packages/web-sdk/src/transports/fetch/transport.ts
codecapitano: Fixed in a54d72b8: a `rateLimitGeneration` counter lets a batch's retry loop detect a newer, longer transport-wide backoff set by a different batch and wait it out before its next attempt, instead of firing on its own now-stale `Retry-After`. Covered by the new test "honors a longer transport-wide Retry-After before an earlier batch retries".

packages/web-sdk/src/config/makeCoreConfig.ts
```diff
@@ -33,11 +34,15 @@ export function makeCoreConfig(browserConfig: BrowserConfig): Config {
...
+    if (browserConfig.experimental?.fetchTransportV2) {
+      internalLogger.error('experimental.fetchTransportV2 cannot take effect when explicit "transports" are defined');
+    }
 
     transports.push(...browserConfig.transports);
   } else if (browserConfig.url) {
+    const Transport = browserConfig.experimental?.fetchTransportV2 ? FetchTransportV2 : FetchTransport;
     transports.push(
-      new FetchTransport({
+      new Transport({
```
codecapitano, line 45: Nit: Transport is very generic. Suggestion: call it `FetchTransport` and import the legacy `FetchTransport` as a custom import e. g. `FetchTransportLegacy` or so?

packages/web-sdk/src/transports/fetch-v2/deliveryQueue.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+
+  private runAttempt<T>(perform: () => Promise<T>): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+      const run = () => {
+        this.inProgress++;
+        perform()
+          .then(resolve, reject)
+          .finally(() => {
```
codecapitano: This code uses `.finally()`. Safari 10.1, Safari 11, and iOS Safari 10.3-11.2 do not support `Promise.prototype.finally`. These browsers are part of this package's browserslist target, and this repo has no polyfill for it. On these browsers, the call throws before `this.inProgress--` runs. This leaks one of the 5 concurrency slots. After 5 such failures, the queue stops making progress.

Suggestion: rewrite the cleanup without `.finally()`, for example with `.then(onSettled, onSettled)`. Add a test that removes `Promise.prototype.finally` to check this case.

_Posted by an AI reviewer._

packages/web-sdk/src/transports/fetch-v2/transport.ts
```diff
@@ -0,0 +1,339 @@
...
+      },
+      getNow: this.getNow,
+      getRandom: options.getRandom ?? Math.random,
+      onRetry: (delayMs, nextAttempt) => {
+        this.logDebug(`Retrying failed request after ${delayMs}ms. Attempt ${nextAttempt}.`);
+      },
+    });
+
+    if (typeof window !== 'undefined') {
+      window.addEventListener('pagehide', () => this.deliveryQueue.flush());
```
codecapitano, line 80: This listener does not check `event.persisted`. `deliveryQueue.flush()` (deliveryQueue.ts:161) sets `unloading = true` and never resets it. The `pagehide` event also fires on normal back/forward-cache navigation, not only on page close. After one back navigation, retries stay disabled for the rest of the page's life.

Suggestion: check `event.persisted` here, or add a `pageshow` listener that resets `unloading = false` when `event.persisted` is true.

_Posted by an AI reviewer._

packages/web-sdk/src/transports/fetch-v2/deliveryQueue.ts
```diff
@@ -0,0 +1,228 @@
...
+            };
+          }
+        }
+      },
+      release,
+    };
+  }
+
+  flush(): void {
+    this.unloading = true;
```
codecapitano, line 163: Note on the back/forward cache (bfcache): `pagehide` is the right event to use here. Unlike `unload` or `beforeunload`, it does not block bfcache. The only gap is the missing `event.persisted` check next to it (see the other comment on transport.ts:79). While a page is frozen in bfcache, this queue's state (timers, waiting list) is paused, not lost, and resumes correctly after `pageshow`. Only this one flag stays stuck. A page can also go through `pagehide`/`pageshow` more than once, so this is not a one-time edge case.

_Posted by an AI reviewer._

packages/web-sdk/src/transports/fetch-v2/transport.ts
```diff
@@ -0,0 +1,339 @@
...
+      }
+      throw error;
+    } finally {
+      reservation.release();
+    }
+  }
+
+  private async fetchWithTimeout(requestInit: RequestInit, timeoutMs: number): Promise<Response> {
+    const callerSignal = this.options.requestOptions?.signal;
+    if (this.requestTimeoutMs <= 0 || typeof AbortController === 'undefined') {
```
codecapitano, line 249: Without `AbortController`, this falls back to a plain `fetch()` with no timeout. The constructor already logs a warning about this. But a request that truly hangs can still hold a concurrency slot indefinitely on affected browsers.

Suggestion: add a short doc note about this limit. If it matters for the target browsers, consider a fallback timeout that does not depend on `AbortController`.

_Posted by an AI reviewer._

packages/web-sdk/src/transports/fetch-v2/transport.ts
```diff
@@ -0,0 +1,339 @@
...
+    const value = response.headers.get('Retry-After')?.trim();
+    if (!value) {
+      return undefined;
+    }
+    if (/^\d+$/.test(value)) {
+      const delay = Number(value) * 1000;
+      return Number.isFinite(delay) ? delay : undefined;
+    }
+    const retryAt = Date.parse(value);
+    return Number.isNaN(retryAt) ? undefined : Math.max(0, retryAt - this.getNow());
```
codecapitano: Two edge cases here do not match this PR's stated goal ('drop a batch when the delay is too long'):
- `Date.parse` accepts many non-standard date formats, so a bad value can still produce a large delay and drop the batch by mistake.
- A very large delta-seconds value overflows to `Infinity`, becomes `undefined`, and falls back to normal backoff instead of being dropped.

Suggestion: treat the overflow case as 'too long' explicitly, and check the date format more strictly.

_Posted by an AI reviewer._

packages/web-sdk/src/transports/fetch-v2/deliveryQueue.ts
```diff
@@ -0,0 +1,228 @@
...
+        if (waitingIndex >= 0) {
+          this.waiting.splice(waitingIndex, 1);
+        }
+        this.nextSendAt = this.options.getNow() + 1;
+        next.resolve(false);
+        this.scheduleNext();
+      },
+      Math.max(0, releaseAt - this.options.getNow())
+    );
+  }
```
codecapitano, line 272: `nextSendAt` spacing applies when a waiting item is released, not when its `perform()` call actually starts. If `attemptQueue` is already backed up, several released items can still fire together once a concurrency slot frees up. This can create a burst of requests against a rate-limited collector, which is the case this spacing is meant to prevent.

Suggestion: apply the `nextSendAt` check when an attempt actually starts, not only when it is released from the wait list.

_Posted by an AI reviewer._

packages/web-sdk/src/transports/fetch-v2/transport.ts
```diff
@@ -0,0 +1,339 @@
...
+    } catch (error) {
+      if (
+        reservation.keepalive &&
+        this.isFetchNetworkError(error) &&
+        !(error instanceof RequestTimeoutError) &&
+        !this.options.requestOptions?.signal?.aborted
+      ) {
+        this.logDebug('Retrying failed keepalive request with keepalive disabled.');
+        const remainingTimeoutMs = this.requestTimeoutMs - (this.getNow() - startedAt);
+        return this.fetchWithTimeout({ ...requestInit, keepalive: false }, remainingTimeoutMs);
```
codecapitano: Open question, not a clear bug: each attempt (up to 3) can trigger 2 raw fetches through this keepalive fallback. So one network or timeout failure can lead to up to 6 real requests. The stable `Idempotency-Key` (line 153) should prevent duplicate data on the server side, but it does not reduce load on the collector.

Question: is 6 requests per batch acceptable for this experimental flag? Should the fallback count against the attempt limit instead?

_Posted by an AI reviewer._

codecapitano (commented):
This is an independent review by two AI models (Claude and Codex/gpt-5.6-sol) of the retry and delivery-queue logic. Two issues should be fixed before this leaves experimental status: `Promise.finally` is not supported on part of the target browsers, and the `pagehide` handler does not account for the back/forward cache. Three smaller issues and one open question are in the comments below.

This review was posted by an AI assistant. Please check the findings before acting on them.

packages/web-sdk/src/transports/fetch-v2/deliveryQueue.ts
Blinkuu: Fixed in [e74c91d](https://github.com/grafana/faro-web-sdk/pull/2247/commits/e74c91dbe95a9a178364e33ebedd36e2d5bbd037).

I have refactored to use `then()` syntax.

packages/web-sdk/src/transports/fetch-v2/transport.ts
Blinkuu: Fixed in [e74c91d](https://github.com/grafana/faro-web-sdk/pull/2247/commits/e74c91dbe95a9a178364e33ebedd36e2d5bbd037).

We now set two event listeners and respective callbacks:

```ts
      window.addEventListener('pagehide', () => this.deliveryQueue.flush());  
      window.addEventListener('pageshow', (event) => {
        if (event.persisted) {
          this.deliveryQueue.resume();
        }
      });
```

packages/web-sdk/src/transports/fetch-v2/deliveryQueue.ts
Blinkuu: Addressed above.

packages/web-sdk/src/transports/fetch-v2/transport.ts
Blinkuu: Let's add a comment. Where would you like me to add it?

packages/web-sdk/src/transports/fetch-v2/transport.ts
Blinkuu: I have implemented a small date parser that adheres to [RFC 9110](https://httpwg.org/specs/rfc9110.html#field.retry-after).

[02329fb](https://github.com/grafana/faro-web-sdk/pull/2247/commits/02329fb1a284aa915f6a369ddf2df01d305113bf)

packages/web-sdk/src/transports/fetch-v2/deliveryQueue.ts
Blinkuu: Fixed in [e74c91d](https://github.com/grafana/faro-web-sdk/pull/2247/commits/e74c91dbe95a9a178364e33ebedd36e2d5bbd037).

packages/web-sdk/src/transports/fetch-v2/transport.ts
Blinkuu: The `keepalive` fallback now counts simply against `maxAttempts`. There's no way we can issue more than `maxAttempts` requests.

packages/web-sdk/src/transports/fetch-v2/deliveryQueue.ts
```diff
@@ -0,0 +1,271 @@
...
+    const jitteredDelay = Math.min(delayMs * (1 + this.options.getRandom() * 0.2), this.options.retry.maxBackoffMs);
+    return new Promise<boolean>((resolve) => {
+      this.waiting.push({ sequence, readyAt: this.options.getNow() + jitteredDelay, resolve });
+      this.waiting.sort((left, right) => left.sequence - right.sequence);
+      if (this.waitingTimer != null) {
+        clearTimeout(this.waitingTimer);
+        this.waitingTimer = undefined;
+      }
+      this.scheduleNext();
+    });
```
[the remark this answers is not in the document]
Blinkuu: Addressed in [2464db0](https://github.com/grafana/faro-web-sdk/pull/2247/commits/2464db02d945bf1a2a1c35f31812dc7e950b3617)

packages/web-sdk/src/transports/fetch-v2/transport.test.ts
```diff
@@ -0,0 +1,508 @@
...
+    );
+  });
+
+  it('flushes a waiting batch once on page hide without scheduling another retry', async () => {
+    fetchMock.mockResolvedValueOnce(response(503)).mockResolvedValueOnce(response(503));
+    const { transport, internalLogger } = createTransport();
+    const sending = transport.send([item]);
+    await jest.advanceTimersByTimeAsync(0);
+
+    window.dispatchEvent(new PageTransitionEvent('pagehide'));
```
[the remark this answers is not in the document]
Blinkuu: Fixed in [fdd4335](https://github.com/grafana/faro-web-sdk/pull/2247/commits/fdd43352c84bafdb7dcd9f4b7c7b333be676cc58)

packages/web-sdk/src/transports/fetch-v2/transport.ts
codecapitano: I would recommend adding it to our docs in Grafana cloud. But tbh, I need to double check if we already have documentation for the fetch transport config and related important information.

codecapitano (approved):
Huge improvement. Thanks for tackling this.

PR #2248, opened 2026-08-20
fix(core): use findIndex when removing instrumentations

### LINKED ISSUE #2235: Instrumentations.remove() fails to remove instrumentations unless they are last in the array
## Description

`Instrumentations.remove()` uses `Array.prototype.reduce` to find the index of the instrumentation to remove, but the reducer resets the accumulator back to `null` on every iteration that doesn't match — even after a match was already found. As a result, the found index is only preserved if the matching instrumentation happens to be the *last* element in the `instrumentations` array. For any instrumentation removed that is not last in the array, the lookup incorrectly returns `null`, so removal is silently skipped (a `"... is not added"` warning is logged and `destroy()` is never called).

## Steps to reproduce

1. Add Faro to the project with multiple instrumentations, e.g. `instrumentations: [instrumentationA, instrumentationB, instrumentationC]`.
2. Call `faro.instrumentations.remove(instrumentationA)` (an instrumentation that is not last in the internal array).
3. Observe that `instrumentationA` is not removed and a warning `Instrumentation "instrumentationA" is not added` is logged, even though it was added.

## Expected behavior

Calling `remove()` with any previously added instrumentation should remove it, regardless of its position in the `instrumentations` array or the order passed to `remove()`.

## Actual behavior

`remove()` only successfully removes an instrumentation when it is the last item in the internal `instrumentations` array (or when `instrumentationsToRemove` is passed in the reverse order of `instrumentations`). Otherwise the internal index lookup resets to `null` and the removal is silently skipped.

## Environment

- **SDK version:** 2.9.0

BetterAndBetterII:
`Instrumentations.remove()` used `reduce` and returned `null` on every non-match, wiping an index that had already been found. Only the last array slot could be removed, so `destroy()` never ran for earlier items.

Use `findIndex` by name, matching how transports already look up items.

Fixes #2235

### DIFF
--- a/packages/core/src/instrumentations/initialize.test.ts
+++ b/packages/core/src/instrumentations/initialize.test.ts
@@ -1,0 +1,95 @@
+import type { API } from '../api';
+import type { InternalLogger } from '../internalLogger';
+import type { Metas } from '../metas';
+import { mockConfig, mockInternalLogger } from '../testUtils';
+import type { Transports } from '../transports';
+
+import { BaseInstrumentation } from './base';
+import { initializeInstrumentations } from './initialize';
+import type { Instrumentations } from './types';
+
+class MockInstrumentation extends BaseInstrumentation {
+  readonly version = '1.0.0';
+
+  initialize = jest.fn();
+  destroy = jest.fn();
+
+  constructor(readonly name: string) {
+    super();
+  }
+}
+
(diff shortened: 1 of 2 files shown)


domasx2 (approved):
LGTM

martinkuba:
@BetterAndBetterII Before we can merge this, we require that all commits are signed. Can you please setup commit signing and rebase? After that, I think we can merge this.

PR #2249, opened 2026-08-21
fix: remove instrumentations regardless of position

### LINKED ISSUE #2235: Instrumentations.remove() fails to remove instrumentations unless they are last in the array

dhruvxvaishnav:
## Why

`Instrumentations.remove()` uses a reducer that resets a previously found index when a later instrumentation does not match. Because of that, removing an instrumentation only works when it is the last item in the array, or when multiple removals are supplied in reverse order.

## What

- Replace the reducer-based lookup with `findIndex`, so a matching instrumentation can be removed from any position.
- Add regression coverage for removing the first, middle, and last instrumentation.
- Cover multiple instrumentations removed in forward order and verify that `destroy()` runs exactly once for every removed instrumentation.

## Links

Closes #2235

## Testing

The regression test was first run against the unchanged implementation: 3 cases failed and only the last-position removal passed. After the fix:

- Focused regression suite: 4 tests passed
- Full `@grafana/faro-core` suite: 188 tests passed
- Core package build passed
- Core TypeScript check passed
- ESLint passed for both changed files
- Prettier check passed for both changed files

## Checklist

- [x] Tests added
- [ ] Changelog updated — release-please generates the changelog from commit messages
- [ ] Documentation updated — no documentation changes are needed for this internal bug fix

### DIFF
--- a/packages/core/src/instrumentations/initialize.test.ts
+++ b/packages/core/src/instrumentations/initialize.test.ts
@@ -1,0 +1,69 @@
+import type { API } from '../api';
+import type { Metas } from '../metas';
+import { mockConfig } from '../testUtils/mockConfig';
+import { mockInternalLogger } from '../testUtils/mockInternalLogger';
+import type { Transports } from '../transports';
+
+import { BaseInstrumentation } from './base';
+import { initializeInstrumentations } from './initialize';
+
+class MockInstrumentation extends BaseInstrumentation {
+  readonly version = '1.0.0';
+
+  initialize = jest.fn();
+  destroy = jest.fn();
+
+  constructor(readonly name: string) {
+    super();
+  }
+}
+
+function createInstrumentations() {
+  return initializeInstrumentations(
+    console,
+    mockInternalLogger,
+    mockConfig(),
+    {} as Metas,
+    {} as Transports,
+    {} as API
+  );
+}
+
+function createMockInstrumentations() {
+  return ['A', 'B', 'C'].map((name) => new MockInstrumentation(name));
+}
+
+describe('instrumentations', () => {
+  describe('remove', () => {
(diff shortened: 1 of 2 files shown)


domasx2:
Thank you @dhruvxvaishnav ! This looks good, but https://github.com/grafana/faro-web-sdk/pull/2248 arrived a bit earlier with same exact fix:) closing this as duplicate, but thanks again.

PR #2251, opened 2026-08-26
feat(web-tracing): omit trace context for unsampled sessions

### LINKED ISSUE #2250: Add an opt-in to omit trace context for unsampled sessions
## Description

When a session isn't sampled, Faro drops every signal in the browser, but the fetch and XHR instrumentations still inject the `traceparent` header with the sampled flag set to `00`.

Backend OTel SDKs default to `parentbased_always_on` sampler, which use the parent's sampling decision. That makes head sampling unusable for anyone relying on backend tail sampling.

Reproduced against a Node service on the default sampler:

| Session | trace context sent | Server span |
|---|---|---|
| sampled | `…-01` | recorded |
| unsampled | `…-00` | dropped |
| unsampled, header omitted | none | recorded, as a new root trace |

## Proposed solution

An opt-in flag on `TracingInstrumentationOptions`:

```ts
new TracingInstrumentation({
  omitTraceContextForUnsampledSessions: true,
})
```

When this flag is present, the instrumentation would not include the `traceparent` header at all. This would in turn allow backend services make their own sampling decision.

The trade-off is generating traces that will not have the frontend span as a root span. So, there will be two sets of traces with different root spans. But this might be preferable than losing the traces completely.

I would suggest the flag becomes the default in the next major version.

martinkuba:
Faro drops every signal for an unsampled session, but the fetch and XHR instrumentations still injected a `traceparent` carrying the unsampled flag `00`. Backend OTel SDKs default to `parentbased_always_on`, which honours that flag and drops the server span, so lowering `sessionTracking.samplingRate` silently removed backend traces too and blocked any backend tail-sampling strategy.

Add an opt-in `omitTraceContextForUnsampledSessions` on `TracingInstrumentationOptions`. When enabled, no trace context is written onto requests from an unsampled session, so each backend service applies its own sampling rules. Defaults to false to preserve current behaviour.

`FaroPausablePropagator` already implemented the whole mechanism -- suppress `inject` while a predicate holds, always delegate `extract` and `fields` -- with nothing pause-specific in it, so both reasons Faro withholds trace context now share that one wrapper rather than nesting a second near-identical propagator. Its constructor parameter is renamed accordingly and the class TSDoc names both reasons.

`isSessionSampled` is lifted out of `getSamplingDecision` so the propagator and the sampler read the same comparison and cannot drift. It stays internal to the package; the option is the whole public interface, so the export surface is unchanged.

Closes #2250

### DIFF
--- a/docs/sources/tutorials/quick-start-browser.md
+++ b/docs/sources/tutorials/quick-start-browser.md
@@ -235,6 +235,48 @@ const faro = initializeFaro({
 });
 ```
 
+### With session sampling below 100%
+
+Session sampling is all or nothing per session. When a session is not part of the sample, Faro
+discards every signal it produces in the browser. The fetch and XMLHttpRequest instrumentations
+still add a `traceparent` header to outgoing requests, but that header carries the unsampled flag
+`00`.
+
+Backend OpenTelemetry SDKs use the `parentbased_always_on` sampler by default, which honors an
+unsampled parent and drops the server span. Lowering `sessionTracking.samplingRate` therefore also
+removes traces from your backend services, not only from the browser.
+
+Set `omitTraceContextForUnsampledSessions` to leave the header off entirely for unsampled sessions,
+so each backend service applies its own sampling rules to those requests.
+
+```ts
+import { getWebInstrumentations, initializeFaro } from '@grafana/faro-web-sdk';
+import { TracingInstrumentation } from '@grafana/faro-web-tracing';
+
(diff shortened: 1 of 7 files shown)


codecapitano:
@martinkuba did you also open a PR to update the cloud docs?

codecapitano (approved):
Great update, thanks. 🦖 
Please don't forget to update the cloud docs as well.

Issue #2255, opened 2026-08-27
Collector-invalidated sessions force unthrottled session re-creation (one per in-flight batch)


Blinkuu:
## Parent

- Session replay delivery reliability epic — `grafana/faro-web-sdk#2241`

## What happens

When the collector responds `202` with `X-Faro-Session-Status: invalid`, both Fetch transports call the session updater with `forceSessionExtend: true`, which unconditionally creates a new session and calls `faro.api.setSession` (`packages/web-sdk/src/transports/fetch/transport.ts`, `packages/web-sdk/src/transports/fetch-v2/transport.ts`, `packages/web-sdk/src/instrumentations/session/sessionManager/sessionManagerUtils.ts`).

The forced re-creation is neither throttled nor checked against the current session. Up to the transport concurrency (default 5) in-flight batches can carry the same stale session id; each receives its own `202`-invalid response, and each response force-creates another session. One stale session fans out into a burst of back-to-back rotations. Every rotation notifies metas listeners twice (core `setSession` removes and re-adds the session meta), and for session replay each rotation stops and restarts rrweb — a full DOM snapshot per rotation and churned recording identity, several times within milliseconds.

## What to fix

Coalesce forced session re-creation: after one `202`-invalid response has rotated the session, further responses for batches that carried the same stale session id must not create additional sessions. Comparing the invalidated batch's session id against the current session before forcing is one option; both transports have the batch's session id in scope.

## Acceptance criteria

- [ ] N concurrent in-flight batches carrying one stale session id produce exactly one new session.
- [ ] A `202`-invalid response for the current session still rotates it exactly once.
- [ ] The fetch and fetch-v2 transports share the guarded behavior.
- [ ] Regression tests cover the concurrent-stale-batch burst for both transports.

## Blocked by

None — can start immediately.

## Further notes

- Found while specifying `grafana/faro-web-sdk#2242`, which requires the replay instrumentation to tolerate these bursts (at most one recording per observed session id); this issue removes the burst at its source.

Issue #2254, opened 2026-08-27
BatchExecutor flush silently discards signals pushed during the flush


Blinkuu:
## Parent

- Session replay delivery reliability epic — `grafana/faro-web-sdk#2241`

## What happens

`BatchExecutor.flush()` sends the buffered signal groups synchronously (`itemGroups.forEach(this.sendFn)`) and only afterwards resets the buffer with `this.signalBuffer = []` (`packages/core/src/transports/batchExecutor.ts`). Any signal pushed through `addItem` while `sendFn` is still executing lands in the buffer between those two steps and is silently discarded — no log, no error, no retry.

Re-entrant pushes during a flush are not exotic. The session instrumentation registers a transport `beforeSend` hook that runs `updateSession()` per item during the send; a session rotation there calls `faro.api.setSession`, metas listeners fire synchronously, and any listener that emits telemetry in response pushes items mid-flush. The replay instrumentation is a concrete victim: rotation stops and restarts recording, and the fresh Meta, FullSnapshot, and `faro.session_recording.started` events it emits are wiped by the buffer reset. The post-rotation recording then starts with orphan incremental events and stays unplayable until the next 5-minute checkout.

The itemLimit-triggered flush (inside `addItem`) and the `visibilitychange: hidden` flush have the same shape, so the loss window exists on every flush path.

## What to fix

`flush()` must not discard items added during the send. Snapshot-and-swap the buffer before iterating (take the current buffer, assign a fresh empty one, then group and send the snapshot), so re-entrant pushes accumulate for the next flush instead of vanishing.

## Acceptance criteria

- [ ] A signal pushed while `flush()` is sending is delivered in a later flush, never discarded.
- [ ] Items sent in one flush keep their order; re-entrant items do not join the in-flight batch.
- [ ] A regression test drives a re-entrant `addItem` from inside a mocked send function and asserts the item is delivered.

## Blocked by

None — can start immediately.

## Further notes

- Found while specifying `grafana/faro-web-sdk#2242`. That issue defers the replay rotation restart out of the flush call stack as a local workaround; this issue is the general fix benefiting all signal types.

Blinkuu:
This is fixed now as part of https://github.com/grafana/faro-web-sdk/pull/2264 PR chain.

PR #2256, opened 2026-08-27
feat(replay): preserve recording identity across tab handoffs

### LINKED ISSUE #2242: Add stable delivery identity to session replay events
## Parent

## What to build

Give every session recording event stable delivery identity that survives request retries. The identity should distinguish the recording, checkpoint generation, and event sequence without changing ordinary Faro events or requiring replay payload inspection by generic transport code.

This establishes the browser-side protocol needed for idempotent delivery and durable retry mechanisms.

## Acceptance criteria

- [ ] Every new session recording event carries stable recording, checkpoint-generation, and sequence identity.
- [ ] Identity is assigned before transport batching and remains unchanged when a request body is retried.
- [ ] Sequence identity remains monotonic within a recording while allowing a fresh checkpoint generation after recovery.
- [ ] Meta, FullSnapshot, and incremental events all participate in the same documented identity model.
- [ ] Existing replay privacy transformations cannot remove or rewrite delivery identity accidentally.
- [ ] Ordinary non-replay Faro events and transport payload construction remain unchanged.
- [ ] Compatibility behavior for older producers and consumers is documented.
- [ ] Tests cover initial recording, periodic checkpoints, inactivity resume, retry reuse, and recovery generation changes.

## Blocked by

None — can start immediately.

Blinkuu:
**Stack 4/5:** #2259 → #2261 → #2260 → #2256 → #2264.

Replay events lacked the identity needed to distinguish concurrent recordings or detect missing events. This adds recording identity and preserves it across safe navigation and same-document handoffs, while recovering with a new identity when ownership is uncertain.

## Changes

- Add `recording_id`, checkpoint generation `gen`, and recording-wide `seq` to replay events; lifecycle markers carry `recording_id`. Reserve counters after Replay's `beforeSend` accepts an event and before serialization, so serialization failures and later global filtering leave detectable gaps.
- Store shared checkpoints in `localStorage` and a per-tab pointer in `sessionStorage`, scoped by global key and app name/namespace/environment. Deployment version changes retain continuity.
- Continue only a clean checkpoint for the same session or an explicit same-document handoff. Seal on `pagehide`, reclaim on BFCache restore, and create a new recording after session rotation or unsafe ownership.
- On a competing storage claim, abandon the identity before restarting. Abandonment survives instrumentation replacement and failed pointer writes; successful handoffs repair stale pointers without overwriting another document's checkpoint.
- Prune malformed and older-than-24-hour checkpoints at claim time, then enforce a 16-checkpoint cap by evicting clean entries before active ones, oldest first. Exempt the claimed recording. Storage failure falls back to document-local continuity; events do not write storage individually.
- Add an experimental changelog entry. Recording-identity documentation is still missing from the README.

## Compatibility and limits

Attributes are additive; the serialized rrweb event is unchanged. Older producers lack identity, so consumers must handle legacy events. Instances sharing the same global key and app identity share a tab pointer, including same-origin frames.

Simultaneous claims can briefly overlap until cross-process `storage` notifications arrive. Pruning can sacrifice continuity for long-lived tabs; recovery starts a new recording. Receiver deduplication is outside this PR.

## Validation

CI passes at `b695fc93` (checked 2026-09-11), including Node tests and browser smoke tests. Coverage includes sequence gaps and reentrant serialization, cloned tabs, ownership loss, failed pointer writes, replacement, pruning, namespace isolation, navigation/BFCache, and real-session/Fetch integration across expiry and delayed invalidation.

Closes #2242. Refs #2241, #2254, #2255.

### DIFF
--- a/experimental/instrumentation-replay/README.md
+++ b/experimental/instrumentation-replay/README.md
@@ -119,12 +119,6 @@ initializeFaro({
 | ------------ | -------------------------------------------------------------- | ----------- | --------------------------------------------------------------------------------------------- |
 | `beforeSend` | `(event: eventWithTime) => eventWithTime \| null \| undefined` | `undefined` | Transform or filter events before they are sent. Return `null` or `undefined` to skip sending |
 
-Replay reconciles session state before starting, pausing, resuming, or accepting recorder events.
-Synchronous capture keeps that decision consistent through event submission. If `beforeSend` or
-rrweb callbacks such as `maskInputFn` change the session, revoke sampling, or destroy Replay,
-the invalidated attempt's pending events are discarded. Startup publishes nothing until rrweb
-returns a stop function and the attempt remains eligible.
-
 ## Privacy and Security
 
 This instrumentation records user interactions on your website. Make sure to:
--- a/experimental/instrumentation-replay/src/instrumentation.test.ts
+++ b/experimental/instrumentation-replay/src/instrumentation.test.ts
@@ -11,12 +11,14 @@ import { EventType } from '@grafana/rrweb-types';
 
 import { defaultMaskInputFn } from './const';
 import { ReplayInstrumentation } from './instrumentation';
+import { replayRecordingCheckpointKeyPrefix } from './recordingState';
 import { MaskInputFn, ReplayInstrumentationOptions } from './types';
 
 // Mock rrweb
-jest.mock('@grafana/rrweb', () => ({
-  record: jest.fn(),
-}));
+jest.mock('@grafana/rrweb', () => {
+  const record = Object.assign(jest.fn(), { takeFullSnapshot: jest.fn() });
+  return { record };
+});
 
 class BatchedBodyTransport extends BaseTransport {
   readonly name = '@grafana/transport-batched-body-mock';
@@ -44,14 +46,16 @@ function createSeededRandom(seed: number): () => number {
 
 describe('ReplayInstrumentation', () => {
   let instrumentation: ReplayInstrumentation;
-  let mockRecord: jest.Mock;
+  let mockRecord: jest.Mock & { takeFullSnapshot: jest.Mock };
   let mockGetSession: jest.Mock;
   let mockAddListener: jest.Mock;
   let mockPushEvent: jest.Mock;
-  let mockCapture: jest.Mock;
+  const mockCapture = (callback?: () => void) => callback?.();
 
   beforeEach(() => {
     jest.clearAllMocks();
+    window.sessionStorage.clear();
+    window.localStorage.clear();
     mockRecord = require('@grafana/rrweb').record;
     mockRecord.mockReturnValue(jest.fn());
 
@@ -59,7 +63,6 @@ describe('ReplayInstrumentation', () => {
     mockGetSession = jest.fn();
     mockAddListener = jest.fn();
     mockPushEvent = jest.fn();
-    mockCapture = jest.fn((callback?: () => void) => callback?.());
   });
 
   afterEach(() => {
@@ -68,6 +71,17 @@ describe('ReplayInstrumentation', () => {
       instrumentation.destroy();
     }
   });
+  function initSampled(
+    options: ReplayInstrumentationOptions = {},
+    sessionId: string = 'test-session'
+  ): ReplayInstrumentation {
+    const inst = new ReplayInstrumentation(options);
+    mockGetSession.mockReturnValue({ id: sessionId, attributes: { isSampled: 'true' } });
+    inst['api'] = { getSession: mockGetSession, pushEvent: mockPushEvent } as any;
+    inst['metas'] = { addListener: mockAddListener, capture: mockCapture } as any;
+    inst.initialize();
+    return inst;
+  }
 
   describe('constructor', () => {
     it('should have correct name and version', () => {
@@ -351,7 +365,9 @@ describe('ReplayInstrumentation', () => {
 
       instrumentation.initialize();
 
-      expect(mockPushEvent).toHaveBeenCalledWith('faro.session_recording.started', {});
+      expect(mockPushEvent).toHaveBeenCalledWith('faro.session_recording.started', {
+        recording_id: expect.any(String),
+      });
     });
 
     it('should not push a faro.session_recording.started event when session is not sampled', () => {
@@ -419,6 +435,9 @@ describe('ReplayInstrumentation', () => {
 
       expect(mockPushEvent).toHaveBeenCalledWith('faro.session_recording.event', {
         event: JSON.stringify(testEvent),
+        recording_id: expect.any(String),
+        gen: '0',
+        seq: '0',
       });
     });
 
@@ -443,6 +462,9 @@ describe('ReplayInstrumentation', () => {
       expect(beforeSend).toHaveBeenCalledWith(testEvent);
       expect(mockPushEvent).toHaveBeenCalledWith('faro.session_recording.event', {
         event: JSON.stringify({ ...testEvent, modified: true }),
+        recording_id: expect.any(String),
+        gen: '0',
+        seq: '0',
       });
     });
 
@@ -759,8 +781,7 @@ describe('ReplayInstrumentation', () => {
         );
 
         api.setSession({ id: 'test-session', attributes: { isSampled: 'true' } });
-        // Session-triggered (re)starts are deferred out of the metas-listener call
-        // stack; flush that microtask before relying on the recorder being active.
+        // Listener-triggered starts are deferred by a microtask.
         await Promise.resolve();
         jest.advanceTimersByTime(1);
         transport.sentBodies = [];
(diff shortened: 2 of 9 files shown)


experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -45,45 +57,96 @@ export class ReplayInstrumentation extends BaseInstrumentation {
   }
 
   initialize(): void {
```
codecapitano, line 116: **Automated agentic review**

`destroy()` sets `this.destroyed = true` (line 494) but nothing resets it back to `false`. If a consumer calls `instrumentations.remove(replayInstance)` and later `instrumentations.add(replayInstance)`, `initialize()` runs again and starts recording once, but every later session-rotation notification goes through `scheduleStartRecording()`, whose deferred check (lines 141-143) returns early because `destroyed` is still `true`. The instance then permanently stops reconciling with session changes, while still looking "started." `destroy()` also never calls `metas.removeListener()`, so repeated destroy/re-add cycles leak listeners too.

Suggested fix: reset `this.destroyed = false` at the top of `initialize()`, and remove the metas listener in `destroy()` (keep a reference to the listener function to pass to `removeListener`).

Severity: medium, not blocking — the instrumentation's normal lifecycle (added once at Faro init, never destroyed) is unaffected.

Blinkuu: Good catch, fixed in [2f40239](https://github.com/grafana/faro-web-sdk/pull/2256/commits/2f40239b6347daad2f497cda94c163d69758bf12)

codecapitano (dismissed):

d0ugal (dismissed):

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -311,15 +552,48 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+        recording_id: this.recordingId,
+        // Clamped for the never-observed case of an event emitted before the first Meta.
+        gen: String(Math.max(this.gen, 0)),
+        // seq is recording-wide and never resets on a generation change, so any hole in
+        // stored sequence numbers is proof of loss.
+        seq: String(this.seq),
+      };
+      this.seq++;
+
+      this.api.pushEvent(faroSessionReplayEventName, attributes);
```
codecapitano, line 574: **High** — A session rotation triggered synchronously from inside `beforeSend` only schedules a deferred reconciliation (`scheduleStartRecording`). But `api.pushEvent` reads the current session from a live getter at push time (`packages/core/src/api/events/initialize.ts:39`, `packages/core/src/metas/initialize.ts:56`). If the reentrant rotation has already updated the session before this push runs, the event goes out with the old `recording_id` under the new session's `meta.session`. Worth revalidating the target session right before the push, not only at claim time.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-sonnet-5 + OpenAI/gpt-5.6-sol (codex-cli, independent second reviewer); version: 2.1.250._

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -155,30 +292,124 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+    const bufferedEvents: eventWithTime[] = [];
+    const attempt: { phase: 'buffering' | 'active' | 'discarded' } = { phase: 'buffering' };
+    const wasRecording = this.isRecording;
+    const wasPaused = this.isPaused;
+    const discardAttempt = (): void => {
+      attempt.phase = 'discarded';
+      bufferedEvents.length = 0;
+    };
+
+    this.isStarting = true;
```
codecapitano, line 305: **High** — Same underlying issue as the `beforeSend` comment on `handleEvent`. If a session rotation fires synchronously from inside `record()` (for example from a user-supplied `maskInputFn`), `isStarting` only defers reconciliation. The `started` marker and any buffered snapshot events still get pushed with the old `recording_id` before that deferred check runs, and by then `meta.session` (read live at push time) can already reflect the new session.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-sonnet-5 + OpenAI/gpt-5.6-sol (codex-cli, independent second reviewer); version: 2.1.250._

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -45,43 +114,116 @@ export class ReplayInstrumentation extends BaseInstrumentation {
   }
 
   initialize(): void {
-    // Check if current session is sampled before starting recording
-    this.checkAndUpdateRecording();
+    this.destroyed = false;
+    this.pageHidden = false;
+    this.recordingStorageKey = `${replayRecordingStorageKey}:${this.config.globalObjectKey}`;
```
codecapitano, line 119: **Medium** — This key is namespaced only by `config.globalObjectKey`, whose default (`'faro'`) is shared by any Faro instance that does not override it (`packages/core/src/config/const.ts:1`, and core itself already warns about two instances sharing this key: `packages/core/src/sdk/faroGlobalObject.ts:11`). Two default-configured Faro instances on the same page would read and write the same `sessionStorage` entry and the same in-memory handoff, and could reset or take over each other's recording identity.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-sonnet-5 + OpenAI/gpt-5.6-sol (codex-cli, independent second reviewer); version: 2.1.250._

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -311,15 +552,48 @@ export class ReplayInstrumentation extends BaseInstrumentation {
         this.sanitizeMetaHref(processedEvent);
       }
 
-      this.api.pushEvent(faroSessionReplayEventName, {
+      // gen advances when the emitted event is a Meta — the event that opens a new
+      // independently playable chain. rrweb's isCheckout flag is not usable as the
+      // trigger: it marks both the Meta and the FullSnapshot of a scheduled checkout
+      // and neither event of an initial or post-resume snapshot.
+      if (processedEvent.type === EventType.Meta) {
```
codecapitano, line 559: **Medium/High** — `gen` advances unconditionally here, but `seq` only advances after the `attributes` object below is built successfully, and its first key (`JSON.stringify(processedEvent)`) can throw on a circular reference or a BigInt value. If it throws, the surrounding catch swallows the error: `gen` has already moved but no `seq` was consumed, so the next successfully-pushed event's `seq` looks contiguous even though the generation-opening Meta event was lost. That defeats the "a seq gap proves loss" guarantee this PR is built around. Consider reserving/advancing `seq` before the step that can throw.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-sonnet-5 + OpenAI/gpt-5.6-sol (codex-cli, independent second reviewer); version: 2.1.250._

codecapitano (commented):
Automated cross-model review: an independent second-opinion pass from a different model, then each finding verified by tracing the code. Four findings inline below.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-sonnet-5 + OpenAI/gpt-5.6-sol (codex-cli, independent second reviewer); version: 2.1.250._

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -62,15 +121,43 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
     this.destroyed = false;
+    this.pageHidden = false;
+    // Owner identity excludes deployment versions so navigation can cross releases.
+    this.recordingStorageKey = `${replayRecordingStorageKey}:${JSON.stringify([
+      this.config.globalObjectKey,
+      this.config.app?.name ?? '',
+      this.config.app?.namespace ?? '',
+      this.config.app?.environment ?? '',
+    ])}`;
+    this.recordingStateStore = new ReplayRecordingStateStore(
```
codecapitano: Handoff namespaces are derived from `config.globalObjectKey` plus app name, namespace, and environment. Two independent Faro instances on the same page with the same app identity and the default `globalObjectKey` compute the same key and share one `sessionStorage` entry, since `claim()` in `recordingState.ts` only checks `sessionId` and `handoff`, not which instance is asking.

When the two instances track different sessions, the second instance's claim overwrites the first instance's persisted state, so the first instance loses recording continuity across its next navigation. This is called out as a caveat in the PR description, but nothing enforces it and no test covers two instances running at the same time.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5; version: 2.1.263._

codecapitano (commented):
This review is part of a pass over the full session-ownership / Replay stack, reviewed against this PR's own stacked base (#2260) rather than `main`. It also includes an independent second-model pass (OpenAI GPT-6 via codex-cli, high effort) over this same diff.

Three findings worth addressing before this layer is restacked onto `main`: a cross-instance storage key collision, a same-origin-tab storage copy that can merge two tabs' recordings into one recording ID, and a generation/sequence ordering edge case during reentrant event serialization. See the inline comments for details.

experimental/instrumentation-replay/src/recordingState.ts
```diff
@@ -0,0 +1,123 @@
...
+    private readonly documentId: string,
+    private readonly generateRecordingId: () => string
+  ) {}
+
+  claim(sessionId: string, activeHandoffRecordingId?: string): ReplayRecordingState {
+    const persisted = this.read();
+    if (
+      persisted?.sessionId === sessionId &&
+      (persisted.handoff === 'clean' ||
+        (persisted.handoff === 'active' && persisted.recordingId === activeHandoffRecordingId))
```
codecapitano: `claim()` treats a `sessionStorage` entry with a matching session ID and `handoff: 'clean'` as proof that this tab is continuing its own earlier recording. The HTML Standard states that a new same-origin tab created from an existing one copies the opener's `sessionStorage`. This happens with `window.open`, a `target="_blank"` link without `rel="noopener"`, and even a plain "duplicate tab" browser action.

If the new tab shares the same Faro session ID as the original, for example under persistent session tracking, it reads the same clean checkpoint and continues the same `recording_id`. Both tabs then advance `gen` and `seq` independently, which produces two colliding replay streams under one recording ID. This path is not covered by `instrumentation.namespace.test.ts`, which only exercises one instance at a time.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5 + OpenAI/gpt-6-astra (codex-cli, independent second reviewer); version: 2.1.263._

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -543,12 +637,21 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+        const seq = this.seq++;
         const serializedEvent = JSON.stringify(processedEvent);
         if (!isCurrentAttempt()) {
           return;
         }
         this.api.pushEvent(faroSessionReplayEventName, {
           event: serializedEvent,
+          recording_id: this.recordingId,
+          gen: String(Math.max(this.gen, 0)),
+          seq: String(seq),
```
codecapitano: `seq` is captured into a local constant before the `JSON.stringify` call that can throw or re-enter. `gen` is read directly from `this.gen` inside the `pushEvent` call below, after `JSON.stringify` has already returned.

If serializing the current event re-enters event handling for a later Meta event, for example through a `beforeSend`-attached value with a custom `toJSON`, that nested call increments `this.gen` before the outer call reads it. The outer, earlier event then reports a generation that a later event actually opened, breaking the ordering the `gen`/`seq` scheme is meant to guarantee. Capturing `gen` into a local constant at the same point as `seq` would close this gap.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5 + OpenAI/gpt-6-astra (codex-cli, independent second reviewer); version: 2.1.263._

experimental/instrumentation-replay/README.md
```diff
@@ -109,6 +109,32 @@ rrweb callbacks such as `maskInputFn` change the session, revoke sampling, or de
...
+
+Each rrweb event carries:
+
+- `recording_id`: one top-level tab's recording within a Faro session.
+- `gen`: checkpoint generation, advancing on each accepted Meta event.
+- `seq`: recording-wide sequence, reserved after `beforeSend` acceptance and before serialization.
+
+Lifecycle markers (`started`, `paused`, `resumed`) carry `recording_id` only. A serialization
+failure leaves a sequence gap; a failed Meta still opens the generation for its following
+FullSnapshot. Intentional filtering consumes no sequence. A gap proves loss, but not every loss
```
codecapitano: "Intentional filtering consumes no sequence" does not say which filter it means. Replay's own `beforeSend` option is checked before `seq` is reserved, in `instrumentation.ts`'s `handleEvent`, so it is correctly gap-free. The separate, global `config.beforeSend` hook runs later, inside the transport layer, after `seq` has already been assigned, so it can leave the same visible gap as a real delivery failure.

Consider naming both filters here, so a reader relying on the "a gap proves loss" guarantee knows that a `config.beforeSend` drop looks identical to a corruption case, not a silent one.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5 + OpenAI/gpt-6-astra (codex-cli, independent second reviewer); version: 2.1.263._

experimental/instrumentation-replay/src/session.integration.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    jest.useFakeTimers();
+    window.sessionStorage.clear();
+    window.localStorage.clear();
+    requests = [];
+    globalThis.fetch = jest.fn(async (_url, init) => {
+      requests.push({
+        sessionId: (init!.headers as Record<string, string>)['x-faro-session-id']!,
+        body: JSON.parse(init!.body as string),
+      });
+      return { status: 202, headers: { get: () => null }, text: async () => '' } as unknown as Response;
```
codecapitano, line 30: Both scenarios in this file use a fetch mock that always returns `headers: { get: () => null }`, so `extendFaroSession()`'s stale/cross-tab guard from #2259 is never exercised together with a real session rotation and a live Replay recording.

Consider adding a scenario where a request captured under an about-to-expire session is delayed past the point the session rotates, then receives a `202` with `X-Faro-Session-Status: invalid`. Verify the SDK ignores it and that Replay's recording under the new session is unaffected. This is the one integration point where all three layers (#2259, #2261, #2260/#2256) interact, and it is currently verified only in each layer's own isolated unit tests.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5 + OpenAI/gpt-6-astra (codex-cli, independent second reviewer); version: 2.1.263._

codecapitano (commented):
Follow-up pass, prompted by cross-checking an internal design-doc explainer for this stack against the actual shipped code and docs. Two findings survived that check and apply to this PR's own documentation and test coverage, independent of that explainer.

experimental/instrumentation-replay/src/recordingState.ts
```diff
@@ -0,0 +1,123 @@
...
+    private readonly storageKey: string,
+    private readonly documentId: string,
+    private readonly generateRecordingId: () => string
+  ) {}
+
+  claim(sessionId: string, activeHandoffRecordingId?: string): ReplayRecordingState {
+    const persisted = this.read();
+    if (
+      persisted?.sessionId === sessionId &&
+      (persisted.handoff === 'clean' ||
```
codecapitano: Agentic review finding (Medium): any clean state is claimable without an ownership token. Independently cloned sessionStorage areas can claim the same recording ID during a page transition; require an unambiguous claim owner.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: This is documented in the README:

```text
Ownership is scoped by `globalObjectKey` and app `name`, `namespace`, and
`environment`. Independent owners with identical app identity must use distinct
global keys. Version and release are excluded so navigation across deployments
can continue. Older unscoped state is not adopted.
```

I don't have a good solution for solving this automatically.

Blinkuu: I have updated the README to make it more explicit: [cd0fd54](https://github.com/grafana/faro-web-sdk/pull/2256/commits/cd0fd5419f86693ebd3445b5541e5873f129c74d)

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -45,21 +57,45 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
     });
+
+    this.checkAndUpdateRecording(false);
   }
 
-  private checkAndUpdateRecording(): void {
+  private checkAndUpdateRecording(deferStart: boolean): void {
+    // A notification can arrive synchronously from a transport hook while a start is
+    // pushing its lifecycle marker or rrweb's initial events. Reconcile once the
+    // current call stack has finished and the recorder has a stop function.
```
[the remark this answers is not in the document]
Blinkuu: Fixed — the comment now says this specifically covers synchronous notifications while `rrweb.record()` is running, before its stop function is installed.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -45,21 +57,45 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+    const sessionId = session.id ?? null;
+    const isSampled = session.attributes?.['isSampled'] === 'true';
+
+    // Globally sampled — apply replay sub-sampling using a deterministic hash of the
+    // session ID so the decision is stable across page reloads within the same session.
+    const replaySampled = sessionId !== null && isSampled && this.shouldReplaySample(sessionId);
+
+    if (!replaySampled || sessionId === null) {
       if (this.isRecording) {
         this.logDebug('Session is not sampled, stopping recording');
```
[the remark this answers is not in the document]
Blinkuu: Fixed — the log now covers both unsampled sessions and sessions without an ID.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: Fixed in `2f40239`: `initialize()` resets the destroyed state and `destroy()` removes the metadata listener. There’s also a remove/re-add regression test covering a later session rotation.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -1,14 +1,41 @@
...
 
 import { defaultMaskInputFn, defaultReplayInstrumentationOptions } from './const';
+import { ReplayRecordingStateStore } from './recordingState';
 import type { ReplayInstrumentationOptions } from './types';
 
 const faroSessionReplayEventName = 'faro.session_recording.event';
 const faroSessionReplayStartedEventName = 'faro.session_recording.started';
 const faroSessionReplayPausedEventName = 'faro.session_recording.paused';
 const faroSessionReplayResumedEventName = 'faro.session_recording.resumed';
+const replayRecordingStorageKey = 'com.grafana.faro.replay';
```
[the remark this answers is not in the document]
Blinkuu: This became obsolete after the storage redesign. Replay now uses separate namespaced tab-pointer and recording-checkpoint keys, and the old unreleased key is removal-only.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -45,43 +114,117 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+    // Globally sampled — apply replay sub-sampling using a deterministic hash of the
+    // session ID so the decision is stable across page reloads within the same session.
+    const replaySampled = sessionId !== null && isSampled && this.shouldReplaySample(sessionId);
+
+    if (!replaySampled || sessionId === null) {
       if (this.isRecording) {
-        this.logDebug('Session is not sampled, stopping recording');
+        this.logDebug('Session is not eligible for replay (unsampled or missing id), stopping recording');
         this.stopRecording();
       } else {
```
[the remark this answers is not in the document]
Blinkuu: Aligned the contract and tests: sampling off and back on within the same Faro session intentionally preserves the recording ID and counters. Session rotation still creates a new recording.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -45,43 +114,116 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
-    this.checkAndUpdateRecording();
+    this.destroyed = false;
+    this.pageHidden = false;
+    this.recordingStorageKey = `${replayRecordingStorageKey}:${this.config.globalObjectKey}`;
+    this.recordingStateStore ??= new ReplayRecordingStateStore(
+      this.getSessionStorage(),
+      this.recordingStorageKey,
+      genShortID(),
+      genShortID
+    );
```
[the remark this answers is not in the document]
Blinkuu: Fixed — the owner namespace and `ReplayRecordingStateStore` are rebuilt on every `initialize()`, so re-adding the instance with different config cannot reuse the previous store.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: Fixed — event submission now captures metadata and revalidates the recorder attempt after Replay’s `beforeSend` and again after serialization, immediately before `pushEvent`. The rotation, unsampling, and destroy cases have regression coverage.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: Fixed — synchronous rrweb emissions are buffered until `record()` returns. If the attempt became stale during startup, it is stopped and discarded before the `started` marker or buffered events are published.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: Addressed by expanding the owner namespace to `globalObjectKey` plus app name, namespace, and environment. Identical owner identities still intentionally share state; the README now calls that out and requires distinct stable keys for independent instances.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: Fixed — `gen` advances and `seq` is reserved before `JSON.stringify`. A serialization failure therefore leaves a visible sequence gap, including when the failed event is Meta.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: I kept identical owner identities as an explicit shared-owner contract rather than trying to infer instance identity. The README now makes this more explicit: independent instances, including same-origin iframes, need distinct stable `globalObjectKey`/app identities.

experimental/instrumentation-replay/src/recordingState.ts
Blinkuu: Fixed with the recording-state redesign. `sessionStorage` now holds only a pointer; ownership and counters live in a shared `localStorage` checkpoint with a document owner. The first claimant continues and later claimants recover under a new recording ID. Clone-first and original-first cases are covered.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: Fixed — both `seq` and `gen` are captured before `JSON.stringify`. The regression test re-enters with a Meta event from `toJSON` and verifies the earlier outer event keeps its original generation.

experimental/instrumentation-replay/README.md
Blinkuu: Clarified — Replay’s own `beforeSend` runs before reservation and leaves no gap, while global `config.beforeSend` runs afterward and can leave the same gap as delivery loss.

experimental/instrumentation-replay/src/session.integration.test.ts
Blinkuu: Added. The integration test delays an invalid response for the old session until after rotation, then verifies the new session and Replay recording remain active and correctly attributed. It runs in both persistence modes.

experimental/instrumentation-replay/src/recordingState.ts
Blinkuu: Addressed with document ownership and ownership-loss recovery. A foreign claim makes the current document abandon the disputed recording and recover under a new ID. The remaining cross-process read/write overlap before the storage event arrives is documented as an accepted browser limitation.

experimental/instrumentation-replay/src/instrumentation.test.ts
```diff
@@ -816,6 +826,909 @@
...
+      const recordingId = replayAttributes()[0]!['recording_id'];
+      const key = `${replayRecordingCheckpointKeyPrefix}${ownerNamespace}:${recordingId}`;
+
+      // Another Document read the same clean checkpoint inside the replication window and
+      // wrote its own claim; the browser delivers that write here as a storage event.
+      const foreignClaim = { ...JSON.parse(window.localStorage.getItem(key)!), documentId: 'other-document' };
+      window.localStorage.setItem(key, JSON.stringify(foreignClaim));
+      mockPushEvent.mockClear();
+      window.dispatchEvent(
+        new StorageEvent('storage', { key, newValue: JSON.stringify(foreignClaim), storageArea: window.localStorage })
```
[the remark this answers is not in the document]
Blinkuu: This looks like a CodeQL false positive. `StorageEvent` accepts `(type, eventInitDict)`, and this call passes exactly those two arguments. The latest CodeQL run is green.

experimental/instrumentation-replay/src/instrumentation.test.ts
```diff
@@ -816,6 +826,909 @@
...
+    ])('should keep recording on a storage event for %s', (_scenario, makeEvent) => {
+      window.sessionStorage.clear();
+      instrumentation = initSampled({}, 'session-a');
+      const ownerNamespace = instrumentation['recordingOwnerNamespace'];
+      emitCallback(metaEvent());
+      const recordingId = replayAttributes()[0]!['recording_id'];
+      const key = `${replayRecordingCheckpointKeyPrefix}${ownerNamespace}:${recordingId}`;
+      mockPushEvent.mockClear();
+
+      window.dispatchEvent(new StorageEvent('storage', { ...makeEvent(key), storageArea: window.localStorage }));
```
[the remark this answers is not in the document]
Blinkuu: Same false positive here: the constructor receives the event type and one `StorageEventInit` object, matching the DOM signature. The latest CodeQL run is green.

experimental/instrumentation-replay/src/instrumentation.test.ts
```diff
@@ -816,6 +837,891 @@
...
+            if (this === window.sessionStorage) {
+              throw new DOMException('The quota has been exceeded.', 'QuotaExceededError');
+            }
+            originalSetItem.call(this, key, value);
+          });
+        }
+        mockPushEvent.mockClear();
+
+        window.dispatchEvent(
+          new StorageEvent('storage', { key, newValue: JSON.stringify(foreignClaim), storageArea: window.localStorage })
```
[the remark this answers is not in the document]
Blinkuu: Same result after the test consolidation: there are exactly two valid constructor arguments, and the latest CodeQL run passes.

codecapitano (approved):
Approved after re-review. Replay startup recovery remains accepted as follow-up #2270.

codecapitano (dismissed):

legander (dismissed):
No blocking findings in recording identity and handoff handling 👍

legander (approved):
👍 LGTM, no blocking findings in recording identity and handoff handling

PR #2257, opened 2026-09-01
fix(core): do not drop signals added during BatchExecutor flush

### LINKED ISSUE #2254: BatchExecutor flush silently discards signals pushed during the flush

RealBhupesh:
## Summary

`BatchExecutor.flush()` grouped and sent `this.signalBuffer`, then assigned `this.signalBuffer = []`. `sendFn` is synchronous, so `beforeSend` / `setSession` can `addItem` during send. Those items were wiped.

## Fix

Snapshot-and-swap before iterating: capture the buffer, reset it, then group/send the snapshot. Re-entrant items land on the new buffer and go out on the next flush.

## Tests

Regression in `batchExecutor.test.ts`: mock `sendFn` that calls `addItem`; the extra item is delivered on the following flush.

Fixes #2254


packages/core/src/transports/batchExecutor.ts
```diff
@@ -81,8 +81,15 @@ export class BatchExecutor {
...
-    const itemGroups = this.groupItems(this.signalBuffer);
-    itemGroups.forEach(this.sendFn);
+    const snapshot = this.signalBuffer;
     this.signalBuffer = [];
+
+    try {
+      const itemGroups = this.groupItems(snapshot);
+      itemGroups.forEach(this.sendFn);
+    } catch (error) {
+      this.signalBuffer = [...snapshot, ...this.signalBuffer];
```
kpelelis, line 91: This effectively overrides the limit of the buffer opening up for a memory overflow if the send function is constantly failing.

The optimal solution here would be to have a retry/exponential backoff mechanism but in order not to blow up the scope of this PR, trimming the buffer to the limit could work as an incomplete alternative.

packages/core/src/transports/batchExecutor.test.ts
```diff
@@ -189,5 +189,86 @@ describe('BatchExecutor', () => {
...
+      expect(mockSendFunction).toHaveBeenCalledTimes(1);
+      expect(mockSendFunction.mock.calls[0]![0]).toEqual([item1]);
+
+      jest.advanceTimersByTime(1000);
+
+      expect(mockSendFunction).toHaveBeenCalledTimes(2);
+      expect(mockSendFunction.mock.calls[1]![0]).toEqual([item2]);
+    });
+
+    it('retries buffered signals after synchronous dispatch fails', () => {
```
kpelelis, line 222: I would add a corner case where the buffer is almost full or full to its limit and after an error happens, there are signals in the buffer again that would overflow the limit

kpelelis (changes requested):
Hey @RealBhupesh this is a very nice find and I am glad you are tackling it. In the lack of a real retry/exponential back off mechanism this is a great alternative and I would be glad to merge it. There is an edge case when running close to the limits of the buffer that could possibly result in undefined behaviour. See my inline comments for the actual issue

Other than that LGTM

packages/core/src/transports/batchExecutor.ts
RealBhupesh: Thanks for the review @kpelelis — agreed on the edge case.

After a failed flush we now call `trimBufferToLimit()` so re-entrant `addItem` calls cannot push the buffer past `itemLimit`. It keeps the oldest items (`slice(0, itemLimit)`) so the failed batch is still retried first.

Added a regression test for a full buffer (`itemLimit: 2`) where the send throws after adding a re-entrant item, and verified the buffer stays at length 2 across repeated timer flushes.

Pushed in 5a470526.

packages/core/src/transports/batchExecutor.test.ts
RealBhupesh: Added `trims the buffer to the item limit after a failed flush when the buffer is already full` in `batchExecutor.test.ts` — covers the full-buffer + failed send + re-entrant `addItem` path and asserts the buffer stays at `itemLimit` across repeated timer flushes.

kpelelis:
@RealBhupesh, LGTM but you need to fix cursoragent so that the CLA is signed

RealBhupesh:
Done. Fixed it!

kpelelis (approved):

martinkuba:
@RealBhupesh The only thing pending is that we require commits to be signed. Can you please setup commit signing and rebase? Thanks!

PR #2258, opened 2026-09-03
fix(web-sdk): coalesce session rotation on collector invalidation

### LINKED ISSUE #2255: Collector-invalidated sessions force unthrottled session re-creation (one per in-flight batch)

RealBhupesh:
## Summary
- Coalesce forced session rotation when the collector returns `202` with `X-Faro-Session-Status: invalid`
- Skip rotation if the invalidated batch session id no longer matches the current in-memory session (concurrent stale batches)
- Share guarded logic between fetch and fetch-v2 transports

Fixes #2255

## Test plan
- [x] Transport tests for fetch and fetch-v2
- [x] Concurrent stale-batch burst produces exactly one `forceSessionExtend` call
- [x] Single invalid response for the current session still rotates once


packages/web-sdk/src/transports/extendSessionOnCollectorInvalidation.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function extendSessionOnCollectorInvalidation(
+  config: Config,
+  invalidatedSessionId: string | undefined,
+  getCurrentSessionId: () => string | undefined,
+  logDebug: BaseExtension['logDebug']
+): void {
+  const currentSessionId = getCurrentSessionId();
+
+  if (invalidatedSessionId != null && currentSessionId !== invalidatedSessionId) {
```
kpelelis: This check effectively skips the new anti-duplicate-rotation guard whenever invalidatedSessionId is undefined (before any session is captured), so concurrent invalid responses in that state still cause the multi-rotation burst the PR is meant to fix.

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -232,12 +232,17 @@ export class FetchTransport extends BaseTransport {
...
     if (response.status === ACCEPTED) {
       const sessionExpired = response.headers.get('X-Faro-Session-Status') === 'invalid';
 
       if (sessionExpired) {
-        this.extendFaroSession(this.config, this.logDebug);
+        extendSessionOnCollectorInvalidation(
+          this.config,
+          invalidatedSessionId,
+          () => this.metas.value.session?.id,
+          this.logDebug
```
kpelelis: This needs to be bound to the current instance as done in fetch-v2 transport

packages/web-sdk/src/transports/fetch-v2/transport.ts
```diff
@@ -149,6 +152,7 @@ export class FetchTransport extends BaseTransport {
     const sessionId = this.metas.value.session?.id;
```
kpelelis: This reads the session id after async header resolution and compression. With async headers or compression enabled, a session can rotate mid-flight, causing the batch to see the wrong id and re-rotate a session that's only milliseconds old

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -213,7 +213,7 @@ export class FetchTransport extends BaseTransport {
         keepalive: keepaliveReservation.keepalive,
       });
 
-      return this.handleResponse(response);
+      return this.handleResponse(response, invalidatedSessionId);
```
kpelelis: returns this.handleResponse(...) without await inside a try block, so a synchronous throw there bypasses the local catch and gets mislogged by send() as a failed delivery for a request the collector actually accepted.

packages/web-sdk/src/transports/fetch-v2/transport.ts
```diff
@@ -185,7 +189,7 @@ export class FetchTransport extends BaseTransport {
         attemptsRemaining,
         () => attemptsMade++
       );
-      this.handleResponse(response);
+      this.handleResponse(response, prepared.invalidatedSessionId);
```
kpelelis: misclassifies a JSON.parse SyntaxError from corrupted session storage as a terminal delivery failure, even though the collector returned 202.

kpelelis (changes requested):
This is a tricky problem and I am glad you are tackling it. I left some inline comments and I have a couple of architectural concerns on the scope of it

1. sessionManagerUtils reads the global faro singleton instead of per-instance state. With isolate: true and two Faro instances, this can update the wrong instance's session.
2. The actual root cause (unthrottled force-rotation in updateSession) is untouched. The fix is patched into both transports instead, so any future third transport or new retry path can reintroduce the same bug.

packages/web-sdk/src/transports/extendSessionOnCollectorInvalidation.ts
RealBhupesh: Addressed in `3a81f457` — coalescing and the stale-session guard now live in `getUserSessionUpdater` rather than the transport helper.

When `invalidatedSessionId` is `undefined`, concurrent invalid responses are coalesced via an `isForceExtending` flag on the shared updater (cached per `metas` in a `WeakMap`). If a session already exists in memory but the batch carried no id (post-rotation), the updater returns early instead of rotating again.

RealBhupesh:
Thanks for the thorough review @kpelelis — pushed `3a81f457` addressing the inline comments and the architectural concerns:

1. **Per-instance state instead of global `faro`** — `getUserSessionUpdater` now accepts a `UserSessionUpdaterContext` (in-memory session id, config, attributes/overrides, `setSession`, `onSessionChange`) bound from the transport's `metas`/`config`/`api`. Transports receive `api` via a small core change (`apiHolder` pattern) so `isolate: true` instances update the correct session.

2. **Root cause fixed at the session layer** — `forceSessionExtend` coalescing (`isForceExtending`) and stale-session guards (mismatched `invalidatedSessionId`, or undefined id with an already-rotated session) are in `getUserSessionUpdater`. `extendSessionOnCollectorInvalidation` only resolves the session manager, builds context, and delegates to a per-`metas` cached updater — any transport calling it gets the same behaviour without duplicating guards.

Happy to adjust further if anything looks off.

packages/web-sdk/src/transports/fetch/transport.ts
RealBhupesh: Fixed — `fetch` now passes `this.logDebug.bind(this)` to `extendSessionOnCollectorInvalidation`, matching `fetch-v2`.

packages/web-sdk/src/transports/fetch/transport.ts
RealBhupesh: Fixed — both code paths in `fetchWithKeepaliveRetry` now `await this.handleResponse(...)`, so synchronous throws from session extension are handled inside the retry path instead of being mislogged as delivery failures.

packages/web-sdk/src/transports/fetch-v2/transport.ts
RealBhupesh: Fixed — `prepareRequest` now captures `invalidatedSessionId` from `this.metas.value.session?.id` before async header resolution and compression, and reuses that value for both the `x-faro-session-id` header and the invalidation handler.

packages/web-sdk/src/transports/fetch-v2/transport.ts
RealBhupesh: Fixed — `performAttempt` wraps `handleResponse` in a try/catch. On 2xx, session/storage errors (e.g. corrupted storage `JSON.parse`) are logged but the attempt still returns success, so they are not treated as terminal delivery failures.

kpelelis (changes requested):
Thanks for addressing the previous comments. I pulled the latest head and there are still a few blockers:

- Both fetch transports take the session header/invalidation id from the current metas, not from the queued payload. A batch captured under session A can be sent with session B in the header and then rotate the wrong session.
- The per-instance context is still incomplete. `isSampled()` and `createUserSessionObject()` read the global `faro`, so isolated instances can use another instance's sampling config and `generateSessionId`.
- For persistent sessions the stale check only compares in-memory state. If another tab has already replaced A with B in local storage, a delayed invalid response for A will overwrite B with another new session.

I reproduced all three cases with focused tests. The existing targeted tests and typechecks pass, but they don't cover these scenarios.

This branch also has rebase conflicts with current `main`, including the session manager and both transport implementations; `fetch-v2` has since been removed. Please rebase, resolve the conflicts, and re-evaluate the changes against the session/delivery work merged in #2259/#2261/#2264. Since current `main` already contains the request-correlated memory/storage guard, closing this as superseded may be cleaner.

PR #2259, opened 2026-09-07
fix(fetch-v2): bind session invalidation to request identity

Blinkuu:
**Stack 1/5:** #2259 → #2261 → #2260 → #2256 → #2264.

Merge in this order into `main`, retargeting each child after its parent merges and retaining parent branches until restacked. Release the complete stack together, after collector CORS support for `Idempotency-Key` is deployed.

With Fetch v2, a batch captured under session A could be sent with session B's header after async preparation. A delayed collector invalidation for A could then rotate B again and restart replay.

## Changes

- Capture the payload's session ID before serialization, async headers, and compression; reuse it for the request header and response handling across retries.
- Rotate on `202` + `X-Faro-Session-Status: invalid` only when the request session still matches both memory and storage. Ignore stale responses and requests without a session ID.
- Document the request/session contract. #2264 brings this behavior to the default Fetch transport.

The storage check and rotation are still separate operations; the remaining cross-tab race is tracked in #2267.

## Validation

CI passes at `72806ef8` (checked 2026-09-11), including Node tests and browser smoke tests. Tests cover async-header and serialization callbacks, queued snapshots, retries across rotation, concurrent invalidations, and a session already replaced by another tab.

Refs #2241, #2255, #2242, #2267.

PR #2260, opened 2026-09-07
fix(replay): discard invalidated recorder attempts

Blinkuu:
**Stack 3/5:** #2259 → #2261 → #2260 → #2256 → #2264.

rrweb can emit events and invoke application callbacks before `record()` returns. Those events could escape a failed startup or be published after a callback changed the session, revoked sampling, or destroyed Replay.

## Changes

- Buffer startup events until rrweb returns a stop function and the attempt is still eligible. Publish `started`/`resumed` before buffered events and discard invalidated attempts.
- Reconcile start, pause, resume, and event submission through metadata capture; recheck attempt validity after user callbacks and serialization.
- Ignore transient missing-session notifications and eligible same-session updates. Stop on rotation and coalesce the replacement start into one microtask.
- Scope deferred starts to their initialize/destroy lifecycle. Reentrant capture cannot install a duplicate recorder or stop a fresh recorder installed during resume.
- Catch and log stop failures. Inactivity pause still stops rrweb when metadata capture or lifecycle-event publication fails; later interaction can retry resume.
- Use seeded randomness in the sampling distribution test.

## Compatibility

Session-triggered restarts now happen one microtask later. This layer uses #2261's capture API; #2264 switches Replay to the compatibility helper. Recording identity fields are added in #2256.

## Validation

CI passes at `a96126d2` (checked 2026-09-11), including Node tests and browser smoke tests. Real-core/rrweb regressions cover startup failure, callback invalidation, lifecycle-marker ordering, stale callbacks, pause failures, expired-session resume, and destroy/reinitialize during capture.

Refs #2241, #2242, #2254. A release-note entry for these safety changes is still needed; #2256's changelog entry covers recording identity.

### DIFF
--- a/experimental/instrumentation-replay/src/instrumentation.ts
+++ b/experimental/instrumentation-replay/src/instrumentation.ts
@@ -42,6 +42,8 @@ export class ReplayInstrumentation extends BaseInstrumentation {
   // Coalesce re-entrant session changes into one deferred start.
   private pendingStart: boolean = false;
   private destroyed: boolean = false;
+  // Deferred starts belong to the lifecycle that scheduled them.
+  private lifecycle = 0;
   // Prevent an invalidated attempt from reviving if the same session returns.
   private attemptRevision = 0;
 
@@ -62,6 +64,9 @@ export class ReplayInstrumentation extends BaseInstrumentation {
 
   initialize(): void {
     this.destroyed = false;
+    this.lifecycle++;
+    // A listener already being notified can still queue work after destroy().
+    this.pendingStart = false;
 
     // Listen for session changes. Starts triggered from the listener are deferred out
     // of the call stack (see scheduleStartRecording).
@@ -134,8 +139,12 @@ export class ReplayInstrumentation extends BaseInstrumentation {
       return;
     }
     this.pendingStart = true;
+    const lifecycle = this.lifecycle;
 
     void Promise.resolve().then(() => {
+      if (lifecycle !== this.lifecycle) {
+        return;
+      }
       this.pendingStart = false;
       if (this.destroyed) {
         return;
@@ -289,9 +298,12 @@ export class ReplayInstrumentation extends BaseInstrumentation {
       discardAttempt();
       try {
         stop();
+      } catch (err) {
+        this.logWarn('Failed to stop session replay', err);
       } finally {
         this.stopRecording();
       }
+      this.logDebug('Recorder start attempt invalidated');
       return false;
     }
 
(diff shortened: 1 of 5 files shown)


experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -194,15 +393,21 @@ export class ReplayInstrumentation extends BaseInstrumentation {
       return;
     }
 
-    if (this.inactivityTimer !== null) {
-      clearTimeout(this.inactivityTimer);
-      this.inactivityTimer = null;
-    }
+    this.metas.capture(() => {
```
codecapitano: Agentic review finding (Medium): capture can rethrow a listener error, and the inactivity timer calls this path without a catch. A throwing listener escapes the timer and prevents replay pause; handle errors consistently with resumeRecording.

Blinkuu: fixed in [aa2aa28c](https://github.com/grafana/faro-web-sdk/commit/aa2aa28c).

The inactivity pause now clears the timer and stops rrweb before metadata capture, so a throwing listener cannot leave recording active. Reconciliation and paused-event publication are caught and logged, with session eligibility rechecked before emitting the event; regression coverage verifies failure containment, recovery, and session changes.

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -158,33 +231,159 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+        if (!this.isRecording) {
+          return;
+        }
 
-      this.setupInactivityTracking();
+        this.logDebug('Session replay started');
+
+        this.setupInactivityTracking();
+      });
     } catch (err) {
```
codecapitano, line 399: Agentic review finding (Medium): A capture-listener exception during initial or rotated-session startup clears ownership but schedules no retry. Interaction listeners are installed only after startup, so Replay can remain disabled until another metadata notification. Add startup recovery and regression coverage.

codecapitano (commented):
Agentic review: one Medium finding.

experimental/instrumentation-replay/src/instrumentation.ts
Blinkuu: Created a follow-up issue: https://github.com/grafana/faro-web-sdk/issues/2270

codecapitano (dismissed):
Approved after re-review. Replay startup recovery remains accepted as follow-up #2270.

codecapitano (approved):

legander (approved):
No blocking findings 👍

PR #2261, opened 2026-09-07
feat(session): reconcile ownership before telemetry capture

Blinkuu:
**Stack 2/5:** #2259 → #2261 → #2260 → #2256 → #2264.

Queued telemetry could change session and sampling ownership when flushed. Delivery-time expiry checks also rotated idle sessions inside the batch flush and let old batches refresh a replacement session.

## Changes

- Reconcile session expiry and cross-tab adoption when metadata is captured. Queued items retain their captured session and sampling decision; delivery only records activity for the same, still-valid session.
- Add `Metas.capture(callback?)` and capture listeners. Nested captures drain pending listeners without rerunning them; failed captures recover cleanly. `metas.value` remains a passive read.
- Capture events, logs, errors, and measurements after deduplication, with a second dedupe check for nested submissions. Traces capture at export.
- Separate session reconciliation from activity refresh, bound checks by expiry deadlines, and handle backward clock changes. Returning to a visible tab still refreshes activity.
- Remove the instrumentation's capture, metadata, and before-send hooks on destruction so re-adding it does not duplicate them.

Capture can rotate a session before `config.beforeSend` rejects a signal. Unsampled telemetry still counts as activity; retries do not. The general batch-flush re-entry issue (#2254) and span-start ownership (#2269) remain separate.

## Compatibility

This layer adds required `Metas` methods and makes public session-manager `updateSession()` check expiry without refreshing activity. #2264 makes the methods optional and restores activity refresh as the public default before release.

## Validation

CI passes at `badb3583` (checked 2026-09-11), including Node tests and browser smoke tests. Coverage includes nested capture and failure recovery, dedupe, hook ordering, expiry and delayed delivery, sampling, clock changes, retries, and instrumentation remove/re-add.

Refs #2241, #2242, #2254, #2269.


codecapitano (commented):
This review is part of a pass over the full session-ownership / Replay stack (#2259 → #2261 → #2260 → #2256), reviewed against this PR's own stacked base (#2259) rather than `main`.

One finding worth addressing: the `Metas` interface change looks like a public API addition, not a patch-level fix. See the inline comment for details.

packages/core/src/metas/types.ts
```diff
@@ -9,6 +9,10 @@ export interface Metas {
   remove: (...getters: MetaItem[]) => void;
   addListener: (listener: MetasListener) => void;
   removeListener: (listener: MetasListener) => void;
+  /** Reconcile before capture. An optional synchronous callback shares this capture with nested telemetry. */
+  capture: (callback?: () => void) => Meta;
+  addCaptureListener: (listener: () => void) => void;
+  removeCaptureListener: (listener: () => void) => void;
```
codecapitano, line 15: `Metas` gains three new required methods (`capture`, `addCaptureListener`, `removeCaptureListener`) and is exported publicly from `@grafana/faro-core`. Any external code that implements this interface will fail to compile until it adds these methods. This PR itself had to update `packages/core/src/api/apiTestHelpers.ts` for the same reason.

The PR title uses `fix`, which this repository's release process maps to a patch version. Adding required members to a public interface usually needs `feat` (minor) or a breaking-change marker (major), not a patch. Please confirm the intended version bump, and consider making the two listener methods optional or internal if external implementers do not need them.

_AI-assisted — Tool: Claude Code; model: Anthropic/claude-opus-5; version: 2.1.263._

Blinkuu: I've updated the title :+1:

packages/core/src/metas/initialize.ts
```diff
@@ -51,11 +53,35 @@ export function initializeMetas(
...
   };
 
+  const capture: Metas['capture'] = (callback) => {
+    // Nested telemetry shares the same synchronous capture, including callbacks
+    // that assign recording identity before submitting their events.
+    const nested = capturing;
+    capturing = true;
+    try {
+      if (!nested) {
+        captureListeners.forEach((listener) => listener());
```
codecapitano: Agentic review finding (Medium): an earlier capture listener can emit telemetry before later listeners run, while nested capture skips those remaining listeners. Complete reconciliation before nested submissions are captured.

packages/core/src/api/traces/initialize.ts
```diff
@@ -43,7 +43,7 @@ export function initializeTracesAPI(
       const item: TransportItem<TraceEvent> = {
         type: TransportItemType.TRACE,
         payload,
-        meta: metas.value,
+        meta: metas.capture(),
```
codecapitano, line 46: Agentic review finding (High): a span is sampled and stamped with session A at start, but export captures current session B here. The session filter can drop the span using B or send an A-tagged span under B; preserve span-start ownership through export.

packages/web-sdk/src/instrumentations/session/instrumentation.ts
```diff
@@ -206,4 +197,11 @@ export class SessionInstrumentation extends BaseInstrumentation {
 
     this.metas.addListener(this.sendSessionStartEvent.bind(this));
   }
+
+  destroy(): void {
+    if (this.captureListener) {
+      this.metas.removeCaptureListener(this.captureListener);
```
codecapitano, line 207: Agentic review finding (High): destroy removes only the capture listener; the before-send hook and bound metas listener remain registered. Re-adding the instrumentation installs a second filter that can drop every item; retain and remove all registrations.

packages/core/src/api/traces/initialize.ts
Blinkuu: The same ownership problem exists on the stacked base (#2259), so it is not introduced by this PR.

Opened #2269 as a subissue of #2241.

packages/core/src/metas/initialize.ts
Blinkuu: Fixed in https://github.com/grafana/faro-web-sdk/commit/badb3583657633c3b20c4edb72f9b548ddf9f1a6. Nested captures now drain pending reconciliation listeners before snapshotting, without rerunning active/completed listeners. Added regression coverage.

packages/web-sdk/src/instrumentations/session/instrumentation.ts
Blinkuu: Fixed in https://github.com/grafana/faro-web-sdk/commit/badb3583657633c3b20c4edb72f9b548ddf9f1a6. destroy() now removes the retained before-send hook and exact bound metadata listener; remove/re-add regressions pass. This covers instrumentation-owned registrations only; session-manager teardown remains out of scope.

codecapitano (dismissed):
Approved after re-review. No blocking findings.

codecapitano (approved):

PR #2264, opened 2026-09-08
feat(web-sdk): make reliable Fetch transport the default

### LINKED ISSUE #2252: Graduate Fetch v2 and remove the legacy transport
## Parent

## What to build

Graduate the reliable Fetch v2 transport from its experimental opt-in to the single stable Fetch transport. Make it the default implementation, preserve the public `FetchTransport` entry point, and remove the legacy transport and experimental selection path.

## Acceptance criteria

- [ ] The reliable transport is used by default when Faro creates a Fetch transport from `url` configuration.
- [ ] The public `FetchTransport` export resolves to the reliable implementation and uses stable, non-v2 runtime naming.
- [ ] `experimental.fetchTransportV2` is removed from public configuration, normalized configuration, selection logic, documentation, and tests.
- [ ] The legacy Fetch transport implementation, tests, types, and obsolete rate-limit behavior are removed.
- [ ] Existing public Fetch transport options that remain applicable stay source-compatible; retired legacy-only options have migration guidance and a semver-appropriate release note.
- [ ] The stable default remains bounded and in-memory; durable replay delivery remains opt-in.
- [ ] Explicit user-provided transports remain unaffected.
- [ ] Internal directories, imports, diagnostics, and tests no longer use the `fetch-v2` or `transport-fetch-v2` names.
- [ ] Tests cover default construction, the public transport export, retry behavior, explicit custom transports, and the removal of the experimental flag.

## Blocked by

- Retry transient Faro delivery failures — #2243

Blinkuu:
**Stack 5/5:** #2259 → #2261 → #2260 → #2256 → #2264.

Make reliable Fetch the default so capture-time session ownership also holds through delivery. Previously, the default transport could send session A's payload with B's header and rotate B on a delayed invalidation for A.

## Changes

- Replace the legacy `FetchTransport` with the reliable implementation: bounded queueing, three attempts by default, a 10-second attempt timeout, jittered backoff, and pagehide/BFCache handling. Preserve body, session header, and one `Idempotency-Key` across retries; apply #2259's request-correlated invalidation guard.
- Always select this transport for URL-based initialization; preserve explicitly supplied transports. Remove `experimental.fetchTransportV2` and the `fetch-v2` module paths.
- Preserve existing constructor options and custom `promiseBuffer` scheduling. Add retry/timeout controls; direct buffer tasks run once. `defaultRateLimitBackoffMs` becomes a deprecated alias for `retry.initialBackoffMs`.
- Make capture methods optional on `Metas` and export `captureMetas` with a legacy fallback. Reconcile fresh custom `transports.execute()` submissions while preserving captured or explicitly different session ownership.
- Restore activity refresh as the public session-manager `updateSession()` default; capture checks opt out. Fetch retries do not refresh activity.

## Migration and delivery behavior

- **Breaking:** remove `experimental.fetchTransportV2` from configuration and replace `fetch-v2` deep imports with package-root exports from `@grafana/faro-web-sdk`.
- **Release prerequisite:** collector/proxy CORS policies must allow `Idempotency-Key`, which is sent without an opt-out. Retries can duplicate already-ingested signals; the header enables receiver deduplication but does not implement it.
- A 429 now retries the affected batch instead of dropping intervening batches during a global cooldown. `Retry-After` on 429/503 is honored within `retry.maxBackoffMs` (30 seconds by default); longer waits terminate that batch. Permanent failures and exhausted retries are logged without payloads.

## Validation

CI passes at `f535aa11` (checked 2026-09-11), including Node tests and browser smoke tests. Coverage includes constructor/buffer compatibility, custom metadata, session ownership, compression, retry limits and cancellation, plus a Playwright test asserting identical body, session header, and idempotency key across a 503 retry. Migration and capture compatibility are documented in the architecture docs.

Closes #2252, #2255. Refs #2241, #2247.


packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -138,121 +202,215 @@ export class FetchTransport extends BaseTransport {
...
+        failure,
+        attemptsMade,
+        retryAfterMs: WAIT_INTERVAL_STATUS_CODES.has(response.status) ? this.getRetryAfterDelayMs(response) : undefined,
+      };
+    } catch (error) {
+      const failure: DeliveryFailure = { error };
+      if (callerSignal?.aborted || !this.isFetchNetworkError(error) || unloading) {
+        return { kind: 'terminal', failure, attemptsMade };
       }
+      return { kind: 'retry', failure, attemptsMade };
```
codecapitano, line 293: Agentic review finding (Medium): a timeout after receiver ingestion reaches this retry path, while current receivers do not guarantee idempotency-key deduplication. Retries can duplicate logs, traces, and billing volume; require receiver deduplication or explicitly accept at-least-once delivery.

packages/web-sdk/src/transports/fetch/transport.ts
```diff
@@ -138,121 +202,215 @@ export class FetchTransport extends BaseTransport {
...
+      requestInit: {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          ...compressionHeaders,
+          ...resolvedHeaders,
+          ...(this.options.apiKey ? { 'x-api-key': this.options.apiKey } : {}),
+          ...(sessionId ? { 'x-faro-session-id': sessionId } : {}),
+          // Unlike dynamic headers, this key is generated once and must remain stable across attempts.
+          'Idempotency-Key': genShortID(20),
```
codecapitano, line 244: Agentic review finding (High): this header is unconditional. Strict collector or proxy CORS allowlists that lack Idempotency-Key will reject preflight and drop all telemetry; gate release on deployed support or provide an opt-out.

Blinkuu: This is already addressed in both Faro Collector and Alloy:

https://github.com/grafana/app-o11y-kwl-endpoint/pull/1793
https://github.com/grafana/alloy/pull/7054

packages/web-sdk/src/transports/fetch/transport.ts
Blinkuu: We have an issue for this in our backlog, I'll address this separately as a follow-up. That said, it's low risk and can technically happen even today.

codecapitano (approved):

legander (approved):
LGTM!
Non blocking point: The new window listeners (pagehide, pageshow) don’t seem to be cleaned up when a transport is removed. I wouldn't expect this to have a noticeable impact in typical usage though 👍

PR #2278, opened 2026-09-15
feat: migrated packages to ESM

jurerotar:
## Why

Updating to a more modern module system.

## What

Updated packages and root files to use ESM instead of CJS.

## Checklist

- [ ] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


domasx2 (changes requested):
there seems to be some mess up with the dependencies, like `npm pkg fix` was run by accident :O it doesn't even install cleanly

jurerotar:
Sorry about that! Re-did it, I converted the project and packages to ESM, then went through resolutions and removed the ones that don't change the lock file, then went through them again and removed the ones that don't throw any audit warnings. There's a few more things I dare not touch (faro-web-tracing seems unused in react package, but I don't have enough context to know if that's for a reason, so I'm not touching it.) 😄

Issue #2268, opened 2026-09-09
Session replay logs page-thrown errors at error level, reading as replay failures


legander:
### Summary

The session replay instrumentation logs every error rrweb's `errorHandler` sees at `error` level. Most of these are not replay failures — they are exceptions thrown by userland APIs that rrweb only happens to be observing. Customers read them as "session replay is broken" and file support tickets.

### Concrete case

A customer using styled-components sees this on every page load right after enabling session replay:

```
Faro
@grafana/faro-instrumentation-replay
Error occurred during session replay SyntaxError: Failed to execute 'insertRule' on 'CSSStyleSheet': Failed to parse the rule '.dpnMSS:-ms-input-placeholder{color:var(--flo-sem-color-text-muted);}'.
    at observer.ts:625:38
    at Object.apply (error-handler.ts:25:24)
    at t.insertRule (Tag.js:37:18)
    at t.insertRules (GroupedTag.js:61:20)
    at e.generateAndInjectStyles (ComponentStyle.js:103:22)
    ...
```

Chain:

1. stylis (styled-components' prefixer) expands `::placeholder` into four selectors, one of which is the legacy single-colon IE form — `Middleware.js` does `replace(value, /:(plac\w+)/, '-ms-' + 'input-$1')`, producing `.dpnMSS:-ms-input-placeholder`.
2. No current browser recognises that selector, so `CSSStyleSheet.insertRule()` throws `SyntaxError`. This happens with or without Faro — it is pre-existing invalid CSS in the app.
3. rrweb proxies `CSSStyleSheet.prototype.insertRule` (`record/observer.ts`) and the proxy is wrapped in `callbackWrapper`, which catches the throw, passes it to the configured `errorHandler`, then rethrows (`record/error-handler.ts`).
4. We register that handler as `errorHandler: (err) => this.logError('Error occurred during session replay', err)` (`experimental/instrumentation-replay/src/instrumentation.ts:152`).
5. The rethrown error lands back in styled-components' own try/catch, which returns `false` and moves on.

Nothing is broken. The rule was invalid and would have been dropped anyway, recording continues, and because `logError` goes through `internalLogger`/`unpatchedConsole` the message is never captured as a Faro error either. It is purely a console message — but it is a red console message with our package name on it, repeated on every render that injects styles, which is why it reads as a product failure.

### Why the current behaviour is wrong

[...]

PR #2284, opened 2026-09-16
fix(replay): stop logging page errors as session replay errors

legander:
## Why

A customer saw errors in their console saying session replay had failed. Nothing was broken. Recording kept working the whole time.

The errors came from their own app. They use styled-components, which builds a CSS rule that no browser accepts. Their app calls `insertRule` with it and it throws. rrweb watches `insertRule`, so it catches the throw, hands it to us, then throws it again so the app handles it as normal.

We logged everything rrweb handed us as an error, with our package name on it, every time the app added styles. So it read like our product was failing when it was not.

## What

- Log these at warn instead of error. Warn is off at the default log level, so the console stays quiet.
- Log the same message once per session instead of every time it happens.
- Reworded it so it no longer says session replay failed.

We still do not swallow the error. If we did, the app would think its call worked and would end up with wrong styles.

Real failures to start recording are untouched and still logged.

Checked both builds in Chrome with 20 bad rules in a row. Main logs 20 errors. This branch logs none at the default level, or one if you turn warn on. The app gets all 20 throws either way.

## Links

- https://github.com/grafana/faro-web-sdk/issues/2268

**In main**
Screenshot 2026-09-16 at 08 59 03

**This PR**
Screenshot 2026-09-16 at 08 58 31

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [ ] Documentation updated


w1kman (dismissed):

experimental/instrumentation-replay/src/instrumentation.ts
```diff
@@ -373,11 +383,53 @@ export class ReplayInstrumentation extends BaseInstrumentation {
...
+      return;
+    }
+
+    if (this.observedErrorSignatures.size >= MAX_OBSERVED_ERROR_SIGNATURES) {
+      return;
+    }
+
+    this.observedErrorSignatures.add(signature);
+
+    this.logWarn('Session replay observed an error thrown by the page. Recording is unaffected.', err);
```
d0ugal: Could `Recording is unaffected` be misleading? I think rrweb also sends recorder failures through this handler. For example, a custom `maskInputFn` which throws a `RangeError`. The replay then misses that input update but still logs this warning.

d0ugal (dismissed):

experimental/instrumentation-replay/src/instrumentation.ts
legander: Good catch, you're right. A maskInputFn that throws kills rrweb's input handler before it emits, so that input never gets recorded. And it's on by default.
I reworded the warn log to "Recording continues, but this event may not have been captured". Also turns out insertRule isn't free either, rrweb records the rule before the real call throws so I added a test with real rrweb so the rethrow is covered in CI.

d0ugal (approved):

Issue #1806, opened 2025-12-05
Unable to initialize Faro with empty transports or in paused state — instrument data is lost when transports are added later


aexei:
## Description
It is currently not possible to initialize the `faro-web-sdk` with an **empty `transports` array**.  
Attempts to use `paused: true` as a workaround also fail because:

1. The SDK starts with `paused: true` or with all transports individually paused.  
2. Instrument data (logs, events, traces) is produced correctly during this time.  
3. When a transport is later added or unpaused, the previously collected instrument data **is never sent**.

This prevents several real-world initialization scenarios, such as:
- dynamic creation of transports after authentication  
- transports that rely on tokens or runtime configuration  
- environments where endpoints cannot be known at startup  

## Expected Behavior
- Initialization with **no transports** (e.g., `transports: []`) should be supported.  
- The SDK should queue instrument data while no active transport exists.  
- Once a transport becomes available, all queued data should be flushed automatically.

## Actual Behavior
- Initializing with an empty `transports` array is currently not allowed.  
- Initializing with `paused: true` causes all instrument data collected during the paused period to be discarded.  
- Adding or unpausing transports later does **not** trigger any flush of previously captured data.

## Steps to Reproduce
```ts
import { initializeFaro } from '@grafana/faro-web-sdk';

// Currently not possible:
const faro = initializeFaro({
  transports: [],   // empty array not allowed
  paused: true,     // also results in data loss
});

// Later at runtime (e.g., after fetching token)
faro.api.addTransport(activeTransport);
faro.unpause();

// Expected: previously collected instrument data should be flushed.
// Actual: no data is sent; everything collected before is lost.
```

## Proposed Solution
1. Permit initialization with zero transports.  
2. Introduce (or fix) an internal queue for all instrument data while no active transports exist.  
3. Automatically flush queued data when a transport is added  
4. Optionally: provide a public `flush()` API.

## Impact
Applications with dynamic authentication or late-bound configuration currently **lose telemetry data**, making reliable instrumentation impossible when transports cannot be initialized upfront.

PR #2285, opened 2026-09-16
fix(tracing): refresh exclusions when transports change

### LINKED ISSUE #1787: Tracing ignores are not updated when transports change, causing Faro to trace its own telemetry endpoint and create an infinite loop
## Description

Tracing ignores are only applied at initialization time.  

If Faro is initialized with an empty `transports` array and the transport is set (or changed) later, the tracing instrumentation does not update its `ignoreUrls`. As a result, Faro starts tracing requests to its own telemetry endpoint (the URL used by the newly configured transport), which causes traces to be sent about those traces and leads to an infinite loop.

The problematic behavior seems to originate here in the web-tracing instrumentation:

https://github.com/grafana/faro-web-sdk/blob/8c4280ce9ed9e53bd9afb3aace2c2d83327a6383/packages/web-tracing/src/instrumentation.ts#L140

At this point, the URLs to be ignored are derived only once from the current config/transport, and are not updated when transports change later.

## Steps to reproduce

1. Initialize Faro with tracing enabled and **no transport**:
   - `transports: []`
   - `TracingInstrumentation` included in `instrumentations`.
2. After initialization (e.g. once an auth token is available), create a `FetchTransport` that points to the telemetry endpoint (e.g. `/collect`) and set/add it to Faro.
3. Trigger any action that produces traces (e.g. normal app usage that generates HTTP spans).
4. Observe that the request to the telemetry endpoint itself is traced.
5. Each telemetry request triggers another traced request to the same endpoint, resulting in a loop.

The issue occurs **every time** this pattern is used (always reproducible).

## Expected behavior

When transports are added or changed after initialization:

- Tracing should automatically ignore the URLs of all active transports.
- Any request going to a URL used by a configured transport (e.g. telemetry/trace endpoint) should **never** be traced.
- Updating the transport (e.g. after a token refresh or when switching endpoints) should not create new traces for the telemetry calls themselves, and therefore no infinite loop.

In short: tracing should always ignore the URLs of the current transports, not only the ones that existed at `initializeFaro` time.

## Actual behavior

- URL ignore logic (`ignoreUrls` or equivalent) is effectively applied only during `initializeFaro()`.
- When the transport is set later and points to the telemetry endpoint, that URL is not ignored by tracing.
- Tracing creates spans for the telemetry requests.
- Each telemetry POST to the trace/telemetry endpoint produces additional traced requests to the same endpoint.
- This results in an effective infinite loop of telemetry calls to the trace endpoint (visible as a continuous stream of POSTs in the browser Network tab).

## Demo

- Browser dev tools → Network tab:
  - Repeated `POST` requests to the telemetry endpoint (e.g. `/collect`) even when there is no new user interaction.
  - Each telemetry request itself carries trace headers and is treated as a traced request, causing the loop.

## Context

- Faro is integrated into an Angular application.
- Faro is initialized without a transport because an Authorization token must be obtained first (via an OAuth2/OIDC AuthService).
- After the token is available, a `FetchTransport` is created with `Authorization: Bearer <token>` and attached to `faro.transports`.
- Tracing instrumentation is active from the start.
- Because tracing ignores are not updated when the transport changes, the telemetry endpoint used by the new transport is not excluded from tracing.

kpelelis:
## Problem and change

Adding or replacing a transport after tracing initializes leaves OTel's ignore list pointing at the old transport configuration. With a custom collector path such as `/telemetry`, Faro traces its own exports and can create a feedback loop. The existing `/collect` smoke test masks this because the default collector regex already excludes that path.

Notify subscribers after transport additions/removals and refresh the default fetch/XHR instrumentations' ignore lists. Resolve relative patterns against the document base URL, preserve configured exclusions and instrumentation options, and unsubscribe when the tracing instrumentation is removed. Explicitly supplied OTel instrumentations remain caller-managed.

Fixes #1787.

## Validation

- New real-browser regression failed before the production change with exported collector spans.
- Three self-tracing browser tests pass: existing relative `/collect`, late-added custom endpoint, and replacement with an absolute endpoint (removed endpoint becomes traceable).
- 14 core transport tests and all 48 web-tracing tests pass, including subscription cleanup and caller-managed instrumentation coverage.
- Full build, repository type checks and lint pass.

This change does not buffer telemetry emitted before a transport is available (#1806).

Implementation and validation were performed by Codex (an LLM) on behalf of Kostas Pelelis; human review is requested.


packages/core/src/transports/types.ts
```diff
@@ -38,6 +38,8 @@ export interface TransportBody {
 }
 
 export interface Transports {
+  /** Observe transport additions/removals. Returns a function that removes the listener. */
+  onChange: (listener: () => void) => () => void;
```
codecapitano, line 42: `Transports` is publicly re-exported by core, web-sdk, and React. Adding a required member breaks existing structurally typed implementations and test doubles at compile time on a non-major update. Please make `onChange` optional (`onChange?: ...`) or move the observer behind an internal interface; the tracing caller already handles its absence with optional chaining.

PR #2161, opened 2026-06-27
fix: serialize all OTLP measurement values

Saithej2k:
Summary

- Add one OTLP log attribute for every measurement payload value.
- Keep the existing measurement.name and measurement.value attributes for the first value.
- Add regression coverage for multi-value measurements.

Testing

- git diff --check
- yarn eslint experimental/transport-otlp-http/src/payload/transform/transform.ts experimental/transport-otlp-http/src/payload/transform/toMeasurementLogRecord.test.ts
- yarn prettier --check experimental/transport-otlp-http/src/payload/transform/transform.ts experimental/transport-otlp-http/src/payload/transform/toMeasurementLogRecord.test.ts
- Focused Jest run for experimental/transport-otlp-http/src/payload/transform/toMeasurementLogRecord.test.ts

Closes #2105

PR #2288, opened 2026-09-16
fix(transport-otlp-http): serialize all measurement values as OTLP attributes

### LINKED ISSUE #2105: OtlpHttpTransport drops all but first value from measurement payload.values (web vitals attribution data lost)

niukanen1:
## What does this PR do?

Fixes #2105

The OTLP transport only extracted the first key-value pair of a measurement's `values`, silently dropping web vitals attribution data (delta, input_delay, processing_duration, presentation_delay)

Now `toMeasurementLogRecord` also emits a `measurement.values` kvlist attribute carrying every entry. The existing `measurement.name` / `measurement.value` attributes are kept for the first value for backwards compatibility — if you'd rather drop the legacy pair, happy to do that

Also fixes the README example which relied on the same broken destructuring

Tests cover multi-value serialization and assert no duplicate attribute keys (per review feedback on #2161)


domasx2 (approved):
nice!

PR #2289, opened 2026-09-17
refactor: replaced direct window access with globalThis to allow faro initialization in workers

jurerotar:
## Why

Faro currently cannot be initialized in web workers due to trying to access `window` object.

## What

Replaced `window` and `global` with `globalThis`.

## Links

[](https://github.com/grafana/faro-web-sdk/issues/1803)

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [x] Documentation updated


packages/core/src/initialize.ts
```diff
@@ -13,11 +13,11 @@ export function initializeFaro(config: Config): Faro {
...
 
   if (isInternalFaroOnGlobalObject() && !config.isolate) {
-    internalLogger.error(
-      'Faro is already registered. Either add instrumentations, transports etc. to the global faro instance or use the "isolate" property'
+    internalLogger.warn(
+      'Faro is already registered. Reusing the existing global Faro instance. Either add instrumentations, transports etc. to the global faro instance or use the "isolate" property'
     );
 
-    return undefined!;
+    return getInternalFaroFromGlobalObject()!;
```
domasx2: This is a public contract change, and it is not in the PR description. Before, a second non-isolated `initializeFaro` returned `undefined` and logged an error. Now it returns the first instance and logs a warning. The second call's config is dropped.

I confirmed this at runtime: a second call with a different `app.name` returned the first instance, and every payload, including a log pushed through the second reference, carried the first app name and version.

Callers that check for `undefined` to detect a double init will now get an instance built from a different config, with no loud signal. If the reuse is intended, please call it out in the description and add a changelog entry. If not, keep the `undefined` return.

docs/sources/developer/architecture/initialization.md
```diff
@@ -29,8 +29,8 @@ by marking them as isolated using the `isolate` config option. This is useful in
 multiple sub-applications which should be instrumented independently.
 
 The process of isolation check is pretty easy: when a singleton Faro instance is initialized, we define it on the
-global object of the environment (either `window` in browsers or `global` in Node.js). If we detect that there is an
-instance already initialized, we stop the process here.
+global object of the environment (`globalThis` when available, then `global` or `self`). If we detect that there is an
+instance already initialized, we reuse it and stop the process here.
```
domasx2: The doc now says Faro reuses the existing instance, but the changelog checkbox is unchecked. Users who upgrade will not learn that double-init semantics changed. Please add a changelog entry.

domasx2 (commented):
Thanks for the PR. I built the `@grafana/faro-web-sdk` IIFE bundle from this head and drove it in a browser page and in a dedicated Web Worker against a local Faro receiver. The page path works as before, and the k6 meta still populates. The worker path does not work yet:

- **Default instrumentations:** `initializeFaro` throws `ReferenceError: window is not defined` from `PerformanceInstrumentation.initialize`. The support check is `'PerformanceObserver' in window` in `packages/web-sdk/src/instrumentations/performance/performanceUtils.ts:58`.
- **`instrumentations: []`:** init returns an instance, but every `pushLog` fails with the same error. `browserMeta` is always on and reads `window.innerWidth` / `window.innerHeight` at `packages/web-sdk/src/metas/browser/meta.ts:23-24`. No payload reached the receiver.

Compared with `main`, this PR moves the worker failure later. `main` throws in `makeCoreConfig` on the old `window.k6` check. So the two k6 lookups are a needed step, but the browser meta and the default instrumentations also need a guard before Faro can start in a worker.

A test that runs `initializeFaro` under a worker-like global would catch this. Right now only the reuse return value is covered.

domasx2 (changes requested):
does not seem to work in workers yet

jurerotar:
Closing in favor of https://github.com/grafana/faro-web-sdk/pull/2294, which is much further along 😄

PR #2291, opened 2026-09-17
fix(web-sdk): wait for prerender activation before starting sessions

### LINKED ISSUE #2290: Chrome prerendering creates two sessions for one visit
## Description

Opening a Faro app in Chrome can create two sessions for one visit. The first gets the initial web vitals. The second gets later activity and session replay. The measurements show `navigation_type=prerender`.

Faro creates a session while Chrome prerenders the page. Chrome replaces the temporary `sessionStorage` when the page becomes active, so Faro loses that session ID and creates another.

_Chrome dev docs specifically mention this having impacting on Analytics and RUM instrumentation https://developer.chrome.com/docs/web-platform/prerender-pages#impact_

## Steps to reproduce

1. Initialize Faro with default session tracking and session replay.
2. Have Chrome prerender the page using [speculation rules](https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API) _(typing in the address to my Faro instrumented app is literally the only thing I did to trigger this)_
3. Navigate to that page and check the session IDs sent to the collector.

https://github.com/user-attachments/assets/e3341682-5149-4761-ab32-c26c9918535a

## Expected behavior

One session for the visit, created or resumed when the page becomes active.

## Environment

Faro 2.11.0, Chrome, macOS.

legander:
## Why

Opening an app instrumented with Faro in Chrome can create two sessions for one visit. The first gets the initial web vitals. The second gets later activity and session replay. The measurements show `navigation_type=prerender`.

Chrome can load a page and run its JavaScript before the user navigates to it. Sites can request this through the [Speculation Rules API](https://developer.chrome.com/docs/web-platform/prerender-pages#the_speculation_rules_api), and Chrome can also trigger it from address bar predictions. The page stays hidden until the user opens it, which activates the prerendered page.

Faro starts during that hidden phase and writes a session ID to temporary `sessionStorage`. Chrome replaces that storage on activation, so Faro loses the ID and starts a second session. One visit is then split across two session IDs.

Chrome's [analytics and RUM guidance](https://developer.chrome.com/docs/web-platform/prerender-pages#impact) recommends waiting for activation before starting analytics on prerendered pages.

## What

- Check `document.prerendering` and wait for `prerenderingchange` before creating or resuming a session.
- Read session storage and make the sampling decision after activation. Prerenders that are never opened send no telemetry.
- Track explicit `setSession()` and `setView()` writes, including overrides that match configuration. Honor `resetSession()` by creating a fresh session at activation. Preserve pending writes when the session instrumentation is removed and readded.
- Preserve API and stored overrides across resets and session replacements unless explicitly replaced or cleared.
- Defer replay initialization until activation, even for explicitly sampled sessions, so recording includes its opening snapshot.
- Collect buffered navigation and resource timings after activation, and report the current view once without intermediate prerender views.
- Drop speculative signals before deduplication and buffering, so identical signals after activation are still captured.
- Establish the session before web vitals emitted during activation are captured, so those measurements and replay use the same session.

`initializeFaro()` still returns synchronously. Apps do not need to add their own prerender guard.

Added session, performance, view and replay regression tests. Core, web SDK and replay tests, typechecks and changed-code lint pass. Verified in Chromium with speculation rules: one session, linked navigation and resource events, one current view event, complete opening replay snapshots, preserved metadata after instrumentation reinitialization, and no telemetry from abandoned prerenders.

## Links

Fixes #2290

**Screen capture of the issue**

https://github.com/user-attachments/assets/1fc4f5b8-ce34-45b4-9c18-9d9b7f01ea6a

**Screen capture with this PR fix**

https://github.com/user-attachments/assets/e952b23d-3ca0-44d4-8576-462c78913d11

## Checklist

- [x] Tests added
- [x] Changelog updated
- [x] Documentation updated


packages/web-sdk/src/instrumentations/session/prerenderSession.ts
```diff
@@ -0,0 +1,82 @@
...
+    document.addEventListener('prerenderingchange', this.activate);
+    // Earlier activation listeners can emit web vitals before our DOM listener runs.
+    this.metas.addCaptureListener?.(this.activate);
+  }
+
+  destroy(): void {
+    this.waiting = false;
+    document.removeEventListener('prerenderingchange', this.activate);
+    this.metas.removeSessionUpdateListener?.(this.recordUpdate);
+    this.metas.removeCaptureListener?.(this.activate);
```
kpelelis, line 81: The capture filter is removed correctly, but by then performance instrumentation has already read the navigation entry and tried to send faro.performance.navigation. We drop that event during prerendering, and Chrome doesn’t emit another navigation callback on activation.

I reproduced this in Chromium: the session starts and resource events are sent, but their navigation ID points to a navigation event we never sent. Can we defer navigation timing collection until activation?

packages/core/src/api/events/initialize.ts
```diff
@@ -37,6 +37,9 @@ export function initializeEventsAPI({
     { skipDedupe, spanContext, timestampOverwriteMs, customPayloadTransformer = (payload: EventEvent) => payload } = {}
   ) => {
     try {
+      if (metas.shouldCapture?.() === false) {
```
kpelelis: If the app calls `setView()`, we drop the event while ViewInstrumentation still marks the view as notified. On activation, it thinks we already reported it.
I think we should report the current view once the page activates, without replaying any intermediate views from prerendering.

packages/web-sdk/src/instrumentations/session/instrumentation.ts
```diff
@@ -165,16 +202,34 @@ export class SessionInstrumentation extends BaseInstrumentation {
...
 
-    const sessionTrackingConfig = this.config.sessionTracking;
+    const prerendering = (document as Document & { prerendering?: boolean }).prerendering;
+    if (this.config.sessionTracking?.enabled && prerendering) {
+      // Chromium discards speculative sessionStorage writes on activation.
+      this.prerenderSession = new PrerenderSession(this.metas, this.transports, (changes) => {
+        this.prerenderSession = undefined;
+        this.initializeSession(changes);
+      });
+      this.prerenderSession.initialize();
```
kpelelis, line 212: This is an edge case, but we already test removing and readding the instrumentation. If we call setSession({ id: 'explicit-session' }) before removing it, readding creates a fresh PrerenderSession without that pending write. Activation then overwrites the explicit session ID.
Can we extend that test to include a pending session write and preserve it across reinitialization?

kpelelis (changes requested):
Amazing find @legander, thanks for reporting and providing a fix for it. I took a look and overall it looks good. There a couple of cases where the SDK might drop signals that we want, but other than that LGTM

packages/web-sdk/src/instrumentations/session/prerenderSession.ts
legander: 👍👍 Performance collection now waits for activation and reads the buffered entries. Verified in Chrome that the navigation event is exported and resource events reference its ID.

packages/core/src/api/events/initialize.ts
legander: Fixed 👍 View reporting now waits for activation and emits the current view once. Verified that view changes during prerender don’t produce intermediate events, and the final view is reported after activation.

packages/web-sdk/src/instrumentations/session/instrumentation.ts
legander: Fixed it. Pending session changes now survive removing and readding the instrumentation. Verified that the explicit session ID, custom attributes and service override are preserved after activation.

legander:
Good findings, and thank's so much for the review! 🙏 

As an aside: I also tested against the published 2.11.0. Some checks failed there too, but for different reasons. For example, it exports intermediate views during prerender, while the earlier PR build drops the view events entirely. Compared both PR revisions in Chrome and verified that the three reported cases fail before these fixes and pass after.

kpelelis (dismissed):

Blinkuu (dismissed):

legander:
Let's keep this PR open a little longer, @Blinkuu had a good point about expansion on the metas interface, so I'm looking into a solution where we keep these coordination hooks an internal interface instead.

legander:
Alright so I did some changes to move the prerender coordination hooks into an internal coordinator instead of extending the public Meta interface. 
Also fixed an activation-ordering edge case in tracing. Session metadata is now established before sampling, so spans started by an earlier prerenderingchange listener aren’t accidentally dropped.

packages/web-sdk/src/instrumentations/session/sessionManager/sessionManagerUtils.ts
```diff
@@ -14,15 +14,13 @@
...
 };
 
-export function createUserSessionObject({
-  sessionId,
-  started,
-  lastActivity,
-  isSampled = true,
-}: CreateUserSessionObjectParams = {}): FaroUserSession {
+export function createUserSessionObject(
+  { sessionId, started, lastActivity, isSampled = true }: CreateUserSessionObjectParams = {},
```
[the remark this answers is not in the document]
legander: Complaining about the use of Math.random for the sampling decision in https://github.com/grafana/faro-web-sdk/blob/3e151dc048e1d0453fa9034bac0e7abf9dacaa3f/packages/web-sdk/src/instrumentations/session/sessionManager/sampling.ts#L10. I'll open a separate PR in Main for improving this rather than blocking this PR which fixes another unrelated issue 👍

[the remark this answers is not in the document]
legander: Draft PR over here https://github.com/grafana/faro-web-sdk/pull/2299, dismissing the alert from the PR

codecapitano (dismissed):

packages/web-sdk/tsconfig.typecheck.json
```diff
@@ -6,7 +6,8 @@
     "isolatedDeclarations": true,
     "noEmit": true,
     "paths": {
-      "@grafana/faro-core": ["../core/src"]
+      "@grafana/faro-core": ["../core/src"],
+      "@grafana/faro-core/internal": ["../core/src/internal"]
```
Blinkuu, line 10: **AI-Generated**

**[P2] Complete the internal-entrypoint mapping in dependent workspaces**

The new `@grafana/faro-core/internal` mapping is also needed in `packages/react/tsconfig.typecheck.json` and `packages/web-tracing/tsconfig.typecheck.json`. Both configs map `@grafana/faro-web-sdk` to its source, so they encounter the new import in `prerenderSession.ts`, but neither maps the internal core subpath to source.

Before core has been built, both typechecks fail with `TS2307: Cannot find module '@grafana/faro-core/internal'`. When build output exists, they instead resolve core's built declarations, which may be stale. This affects pre-build local typechecks; it does **not** break CI, which builds first.

Please add `"@grafana/faro-core/internal": ["../core/src/internal"]` to both sibling typecheck configs. Verified with TypeScript's compiler API: hiding core build output reproduces the error in both packages, and adding the mapping in memory clears both errors.

packages/web-tracing/src/instrumentation.ts
```diff
@@ -109,8 +109,14 @@ export class TracingInstrumentation extends BaseInstrumentation {
       resource,
       sampler: {
         shouldSample: () => {
+          // Drop speculative spans before OpenTelemetry can queue them for export after activation.
+          if ((document as Document & { prerendering?: boolean }).prerendering) {
+            return { decision: SamplingDecision.NOT_RECORD };
+          }
```
Blinkuu, line 118: **AI-Generated**

**[P2] Honor the session-tracking opt-out when dropping prerender spans**

This returns `NOT_RECORD` whenever `document.prerendering` is true, even when `sessionTracking.enabled` is false. That changes behavior for applications managing their own sampled session:

```ts
// Configuration:
sessionTracking: { enabled: false }

// Application-managed session:
faro.api.setSession({
  id: 'application-managed',
  attributes: { isSampled: 'true' },
});
```

Spans started during prerender are now silently discarded despite the documented opt-out from the session activation guard. Please make this guard respect disabled session tracking.

Verified with a temporary Jest/jsdom reproduction: both normal-page and prerender cases pass with `main`'s tracing implementation; on this branch, the normal-page case passes but the prerender span is non-recording. Without explicitly sampled session metadata, spans were already non-recording on `main`; the regression specifically affects application-managed sampled sessions.

Blinkuu (dismissed):

eskirk:
hey, I found one remaining edge case in `packages/web-sdk/src/instrumentations/session/prerenderSession.ts`.

When `setSession()` receives both payload and option overrides, `recordUpdate()` currently stores only `update.overrides`, so payload overrides are lost before activation:

```ts
faro.api.setSession(
  { overrides: { geoLocationTrackingEnabled: true } },
  { overrides: { serviceName: 'checkout' } }
);
```

I suggest changing the override handling to merge the option overrides into the explicit session overrides, while keeping explicit `{ overrides: {} }` as a clear:

```ts
const hasSessionOverrides = session?.overrides !== undefined;
const overrides = hasSessionOverrides
  ? { ...session.overrides, ...(update.overrides ?? {}) }
  : update.overrides
    ? { ...previous?.session.overrides, ...update.overrides }
    : previous?.session.overrides;

this.pending = {
  reset: reset || (!session?.id && (previous?.reset ?? false)),
  inheritOverrides: hasSessionOverrides ? false : (previous?.inheritOverrides ?? true),
  session: {
    id: session?.id,
    attributes: session?.attributes && { ...session.attributes },
    overrides: overrides && { ...overrides },
  },
};
```

I’d also add a regression test covering both override sources in the same `setSession()` call.

legander:
Appreciate the review @eskirk 🙏  Verified that this issue exists in main and predates this PR,even for normal navigation. Something we should fix for sure but would suggest we raise a separate PR against main 👍

eskirk (approved):

eskirk:
@legander cool, lets create a follow up issue for that. in that case, we are good to merge.

d0ugal (approved):

legander:
Opened a new issue for your find @eskirk https://github.com/grafana/faro-web-sdk/issues/2303 👍

PR #2294, opened 2026-09-18
feat(web-sdk): support shared and dedicated web workers

bereziukadam:
## Why

Faro defaults access `window`, `document`, and web storage, preventing initialization in SharedWorkers. This PR adds shared and dedicated worker support, based on custom instrumentations I have used in some products for few months.

## What

- Select worker-safe errors, console, metadata, and in-memory sessions automatically.
- Reuse existing error handlers with worker-safe globals; retain fixes for falsy rejections and instrumentation removal.
- Initialize per-instance session sampling before signals and skip XHR tracing when XMLHttpRequest is unavailable.
- Update support documentation and add worker regression tests.

Worker sessions last for the Faro instance's lifetime, without storage persistence or inactivity rotation. Service workers have **not been tested in a browser** and remain unsupported. Unit tests only simulate worker globals and missing APIs; they do not validate service-worker lifecycle or delivery.

## Validation

- Build, type checks, lint, circular-dependency checks, unit suites, and all 18 Chromium smoke tests pass, including worker telemetry, HTTP tracing, and multi-tab session continuity.
- WebKit limitation: shared workers stop responding after the creating tab closes; reproduced without Faro.
- Firefox validation is incomplete because the browser failed to launch locally.

## Links

Related issues: #1803, #2235, #1643.

Builds on #2122 (worker-safe batching).

## Checklist

- [x] Tests added
- [ ] Changelog updated
- [x] Documentation updated


domasx2:
Service worker is instrumented silently despite being documented as unsupported

Registering sw.js that calls initializeFaro with getWebInstrumentations() returned scope '[object ServiceWorkerGlobalScope]', initialized=true, session id zPcNPE3UFM and the three worker instrumentations, with no warning at internalLoggerLevel WARN; session_start and sw-event were ingested by Pinot under worker-serviceworker-2294. Users reading the support table will not learn they are on an untested path.

Service worker is instrumented silently despite being documented as unsupported

domasx2:
console.error(Error) inside a worker is dropped with 'document is not defined'

In every worker run (shared, dedicated, persistent), calling console.error(new Error('worker-console-error-object')) produced the worker console line 'Faro @grafana/faro-web-sdk:instrumentation-console ReferenceError: document is not defined at wn (faro-web-sdk.iife.js:18:35811)' and no exception for it appeared in any /collect body, while console.error with a string and all uncaught errors in the same batch were delivered. Worker code that reports caught errors via console.error loses them entirely.

domasx2:
`buildStackFrame` uses `document.location.href` when no filename is given. `getDetailsFromErrorArgs` calls it eagerly with `source`, which is undefined for `console.error(err)` and empty for some `onerror` events, so it throws a ReferenceError in a worker.

In a worker, `console.error(new Error(...))` is swallowed by ConsoleInstrumentation's catch and never reported. An `onerror` event with an empty source makes the Faro handler throw, and that error is lost.

domasx2 (changes requested):
Hi, thanks! End to end testing came  up with some issues, also has conflicts

